PR #2, opened 2025-06-30
Swarm rework dev merged

lmoresi:
Here we go again !

This is reworking the PIC swarm to use the PETSc BASIC swarm type. There is a fair bit of fixing up to make this happen, though the results should simplify much of our workflow in the long run.

The BASIC swarm bypasses all of the PETSc-provided migration tools and does not need us to identify the relationships between cells and particles. To provide migration tools of our own, a k-d Tree based navigation toolkit is provided. We can quickly identify which domain a given particle belongs to. We can be precise about this or we can estimate / iterate. 

Why ?

  - We don't really use PIC at this point. We instead develop field-based proxy variables and pass those to the point wise function framework. 
  - The overhead with PIC is substantial, particularly with simplices. There are some good ways to map particles to nearby points / cells (yes, we can still do that !) that work with k-d Trees to make things efficient. 
  - We can send particles directly to new domains without risk of losing them. This means we can use particle swarms to quickly pass information globally around a mesh. e.g. passing data between two meshes that have different decompositions
  - We can rely on global sampling for things like the semi-Lagrange advection schemes. We quickly determine global sample points and send them - / send them back - to the location where we need the information.


lmoresi:
Not yet ready for merging. Several tests failing.

Issue #3, opened 2025-07-01
error in meshvariable to meshvariable projection


gthyagi:
Hi @lmoresi and @julesghub,

I’m using the mesh-to-mesh variable projection from [this section of the code](https://github.com/underworldcode/underworld3/blob/faed46cc4535fcdcd1876206a822d20b38f8e036/src/underworld3/adaptivity.py#L370).

It works well in general, except at the origin (0, 0, 0).

I’m attaching a simple Python script to reproduce the issue. The projected values between meshes can be found in the variable `v_soln_2_m1tom2`.

[demo_meshvar2meshvar_err-Copy1.py.zip](https://github.com/user-attachments/files/20999207/demo_meshvar2meshvar_err-Copy1.py.zip)

Thanks for taking a look!

gthyagi:
When I change the origin from (0, 0, 0) to (10, 10, 10), the issue with anomalous values at the cell containing (0, 0, 0) disappears. This suggests it’s related to floating point precision.

Note: I haven’t checked yet whether this issue originates from the Gmsh level or the DMPlex level.

gthyagi:
@knepley I believe this issue is occurring at the DMPlex and DMSwarm level. Do you have any comments on this?

It seems that if the model’s bounding box coordinates include zero, there’s a floating point precision error involving the zero value.

Note: Mesh1 and Mesh2 have the same bounding box coordinates but different cell sizes.

![Image](https://github.com/user-attachments/assets/385bae79-893f-4e31-bb8b-0e6acbdcbc13)

knepley:
@gthyagi What routine has the problem?

jcgraciosa:
@gthyagi, I had a look at the attached code. The main cause of the issue is the behaviour of ```swarm.dm.addNPoints()``` inside  ```mesh2mesh_swarm```:

- If the swarm is still empty, to add N points, you need to to do ```swarm.dm.addNPoints(N + 1)``` and this is what you see in lines 105 - 108.
- If the swarm already has particles inside, then to add M points, you need to do ```swarm.dm.addNPoints(M)```

With this, even if ```n_found1``` is equal to zero, ```adds``` will be equal to 1. Consequently, ```addNPoints``` in line 161 adds a particle, but is not assigned a proper coordinate so it retains its default coordinate of (0., 0., 0.). This results to the issue you see.  

[...]

PR #4, opened 2025-07-08
Added fix to mesh2mesh_meshVariable()

jcgraciosa:
Proposed fix to issue #3. 
Also added quick pytest to monitor changes in the behaviour of ```swarm.dm.addNPoints()```.

### DIFF
--- a/src/underworld3/adaptivity.py
+++ b/src/underworld3/adaptivity.py
@@ -268,27 +268,30 @@ def mesh2mesh_swarm(mesh0, mesh1, swarm0, swarmVarList, proxy=True, verbose=Fals
     ## some of these. So let's add them now
 
     cell = mesh1.get_closest_local_cells(global_unallocated_coords)
-    found1 = np.where(cell >= 0)[0]
-    not_found1 = np.where(cell == -1)[0]
+    # found1 = np.where(cell >= 0)[0]
+    # not_found1 = np.where(cell == -1)[0]s
+    cell_arr = np.atleast_1d(cell)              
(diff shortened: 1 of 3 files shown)


src/underworld3/adaptivity.py
```diff
@@ -268,27 +268,30 @@ def mesh2mesh_swarm(mesh0, mesh1, swarm0, swarmVarList, proxy=True, verbose=Fals
...
+    cell_arr = np.atleast_1d(cell)              
+    found1 = np.where(cell_arr >= 0)[0]
+    not_found1 = np.where(cell_arr == -1)[0]
 
     n_found1 = found1.shape[0]
     n_not_found1 = not_found1.shape[0]
 
-    psize = swarm1.dm.getLocalSize()
-    adds = n_found1 + 1
+    if n_found1 > 0:
```
julesghub, line 280: Should this be the length of n_found1?

julesghub (changes requested):

src/underworld3/adaptivity.py
gthyagi: `n_found1` gets an integer, so that should be fine. @jcgraciosa can you please confirm? thnks

jcgraciosa: Yes, that is correct @gthyagi.

gthyagi: @julesghub, please merge this when you can. I’ll rerun my fault models with this fix. Thanks! Cheers.

PR #21, opened 2025-08-19
Development

lmoresi:
This brings the README updates (these include the JOSS publication information).

The nasty bug that I introduced into the kdtree algorithm - a fix for this, I mean !

That's all.


lmoresi:
dev to main for the purposes of bringing the new code in to development for wider testing.

Issue #25, opened 2025-09-22
[BUG] -AttributeError: 'NaturalBC' object has no attribute 'fn_F' when using add_natural_bc


Belay-129:
Hi uw3 support team,
 I encountered an error when applying a natural boundary condition to the Poisson solver in Underworld3, which prevents the solve from running successfully.

poisson.add_dirichlet_bc([17.0], "Top")
q_bottom = 0.065 # W/m2
poisson.add_natural_bc([q_bottom], "Bottom")

poisson.solve()

When applying a natural boundary condition to a Poisson solver, the following error occurs:

AttributeError                            Traceback (most recent call last)
Cell In[18], line 1
----> 1 poisson.solve()

File src/underworld3/cython/petsc_generic_snes_solvers.pyx:872, in underworld3.cython.generic_solvers.SNES_Scalar.solve()

File src/underworld3/cython/petsc_generic_snes_solvers.pyx:194, in underworld3.cython.generic_solvers.SolverBaseClass._build()

File src/underworld3/cython/petsc_generic_snes_solvers.pyx:744, in underworld3.cython.generic_solvers.SNES_Scalar._setup_pointwise_functions()

AttributeError: 'NaturalBC' object has no attribute 'fn_F'

Thanks for your time and help

gthyagi:
Hi @Belay-129,

Please attach the example script you are running so we can debug. It also looks like you’re using an older version of Underworld, please update to the latest release, as some of these issues are already fixed.

Cheers,
Thyagi

Belay-129:
Hi Thyagi,
Thank you for your response. Could you please confirm the current version of Underworld? I checked on GitHub and found the version listed as:
__version__ = "0.99.0b"

This matches the version I already have installed.
Cheers,
Belay



________________________________

Belay-129:
'''
mesh = uw.meshing.UnstructuredSimplexBox(
    minCoords=(423755, -10000),
    maxCoords=(496255, 0),
    cellSize=1000,
    regular=False,
)
'''

'''
k_scalar = uw.discretisation.MeshVariable("k", mesh, 1)
'''

'''
T_soln = uw.discretisation.MeshVariable("T", mesh, 1)
'''

'''
**Set up the Poisson system**
poisson = uw.systems.Poisson(mesh, u_Field=T_soln)

poisson.constitutive_model = uw.constitutive_models.DiffusionModel

poisson.constitutive_model.Parameters.diffusivity = k_scalar.sym[0]

[...]

PR #26, opened 2025-09-23
Fix for bug #25

jcgraciosa:
Changes: 

- Fix for bug #25 
- Made changes in setting natural BCs in ```SNES_Scalar``` solver for natural BCs to work. 
- Added pytest for Poisson (scalar) natural BC


lmoresi (approved):
We should certainly keep the commented out Jacobians in place (don't delete them) - they will be needed for surface problems in general.

PR #32, opened 2025-11-18
UW3 release candidate

lmoresi:
The full release of UW3 should be at 3.0 ... when we do this, I would like to introduce some new coding patterns that are more conducive to the having and AI-bot (AI-Bat ?) give advice. 

So that means there is a lot to review in here.


lmoresi:
Remarkably, a long-standing bug in the interpolation code showed up and has been killed.

lmoresi:
All test now passing - let's merge and finalise ...

PR #44, opened 2026-01-27
Fix PETSc build by auto-detecting active Pixi environment

gthyagi:
This PR replaces hard-coded Pixi environment detection in `build-petsc.sh` with a robust, automatic detection of the active Pixi environment. The script now reliably uses the environment invoked via `pixi run -e <env>`, eliminating mismatches between MPI/HDF5 paths and preventing PETSc/PTScotch build failures.


lmoresi (approved):
Tested - this is a legit bug fix !

PR #49, opened 2026-02-18
Fix UWexpression parameter display and tensor assignment

lmoresi:
## Summary

- **ExpressionDescriptor.__set__**: Store UWexpression as symbolic reference instead of extracting inner value. Fixes parameter display showing numeric values (e.g. `1e21 Pa.s`) instead of symbolic names (e.g. `eta_0`), and restores lazy evaluation.
- **_unwrap_atom**: Recursively resolve UWQuantity inside UWexpression chains during JIT compilation, preventing `SympifyError`.
- **_unscaled_matrix_to_rank4**: Add `__getitem__` wrapping guard for NDimArray assignment -- the single choke point for all constitutive models after `simplify()` strips per-model wrapping.
- **_unscaled_matrix_to_rank2**: Fix pre-existing bug where loop body never assigned values (always returned zeros).
- **get_units() in unit_conversion.py**: Restore delegating wrapper for compiled Cython modules (ckdtree.pyx) that import from this location.
- **Surface coordinate space handling**: Add dimensional/ND gateways to Surface class. Fix `_interpolate_to_proxy()` to use internal model-space coordinates for KDTree, avoiding mismatch when units system is active.

## Test plan

- [x] ViscousFlowModel with UWexpression(UWQuantity) -- solve succeeds
- [x] TransverseIsotropicFlowModel with UWexpression(UWQuantity) -- solve succeeds
- [x] Plain numeric and plain UWexpression parameters -- no regression
- [x] Stokes solver tests (test_1010) -- 11/11 passing
- [x] Poisson solver tests (test_1000, test_1001) -- 9/9 passing
- [x] Rank-2 tensor roundtrip (mandel_to_rank2 fix) -- 2D and 3D verified
- [x] Surface _interpolate_to_proxy with units/scaling active -- KDTree coordinate match verified

### DIFF
--- a/src/underworld3/function/expressions.py
+++ b/src/underworld3/function/expressions.py
@@ -106,7 +106,12 @@ def _unwrap_atom(atom, mode='nondimensional'):
                 except Exception:
                     pass
             # Recursively unwrap to get inner expression
-            return atom.sym
+            inner = atom.sym
+            # If inner is a UWQuantity (not SymPy-compatible), resolve it
+            # to a plain value so subs() doesn't fail with SympifyError.
+            if isinstance(inner, UWQuantity) and not isinstance(inner, UWexpression):
+                return _unwrap_atom(inner, mode)
+            return inner
         else:  # symbolic
             return atom.sym
 
--- a/src/underworld3/function/unit_conversion.py
+++ b/src/underworld3/function/unit_conversion.py
@@ -123,11 +123,18 @@ def has_units(obj):
         return True
 
     return False
-# DEPRECATED: Old get_units() from function.unit_conversion has been removed.
-# Use uw.get_units() from the units module instead, which provides the unified,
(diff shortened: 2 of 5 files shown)


lmoresi:
Open review period ended, ready to merge.

lmoresi:
These are important bug fixes to the code rather than feature upgrades. This will be merged once the tests are fixed.

lmoresi:
CI failure is pre-existing and unrelated to this PR: 7 tests marked `xfail` now pass (`xpassed`), which pytest treats as failure in strict mode. The `xfail` markers on those tests are stale — the bugs they expected have been fixed. These markers should be removed in a follow-up so CI goes green.

The Fatal Python abort is also pre-existing (PyVista/VTK cleanup in headless CI).

All 280 actual tests pass. Merging.

Issue #54, opened 2026-02-24
No variables saved when using mesh.petsc_checkpoint


bknight1:
After upgrading to development and petsc-3.24.4 (from 3.21.5, joss branch) I no longer get any variables in the h5 and xdmf files. xdmf files in paraview do not contain the variables for visualisation, and the h5 files have all default values (0) when opening in a h5 viewer.

Maybe a duplicate of https://github.com/underworldcode/underworld3-archive/issues/259

lmoresi:
We need to do a full shake-out of the checkpointing. We have two parallel systems going right now and that's not going to be sustainable.

lmoresi:
Fixed on `development` in commit 2c5f295.

**Root cause**: After PETSc solves, solution data was copied only to `_lvec` (local vector) but `_gvec` (global vector) was never updated. Since all write/checkpoint methods write from `_gvec`, every saved field was zeros.

**Fix**: Added `_sync_lvec_to_gvec()` method that uses the correct subdm scatter pattern, called in:
- All 3 base solver finalizations (SNES_Scalar, SNES_Vector, SNES_Stokes_SaddlePt) — covers all 10 solver subclasses
- All write/checkpoint methods (belt-and-suspenders)

The operation is idempotent so the double coverage is safe.

A second fix (commit 5a6d63c) also addresses `read_timestep` failing when a `Model` with length scaling is active — the KDTree query coordinates had pint units that didn't match the plain floats stored in HDF5.

PR #63, opened 2026-03-02
Fix for xdmf output

bknight1:
Further fixes for https://github.com/underworldcode/underworld3/issues/54
- Add EnhancedMeshVariable._sync_lvec_to_gvec
- update generateXdmf and generate_uw_Xdmf to use the new collector for consistent field discovery.


lmoresi (approved):
Thanks, Ben

PR #76, opened 2026-03-10
perf(mpi): cap thread pools for MPI and document policy

gthyagi:
Cherry-picks only these commits from gthyagi:feature/boundary-integrals:\n- 014e97c perf(mpi): cap thread pools for MPI and document policy\n- d42f6b2 revert: restore uw wrapper from development\n\nThis keeps boundary-integral development/testing isolated in the original branch/PR.


gthyagi:
With this fix and OpenMPI on macOS, I am no longer seeing any CPU spikes (>100%).

Issue #82, opened 2026-03-18
[BUG] - Navier-stokes solution unstable after first timestep


bknight1:
**Describe the bug**
The numerical solution for the Navier-Stokes solver is unstable after the first solve. 

To debug, I created a Stokes solver which has been extended to include the NS terms. After debugging, I believe the error is due to the SemiLagrangian history manager for DuDt, as when switching to the Eulerian history manager the solution is stable. I believe the flux term may also need to be switched to Eulerian, although this hasn't been tested as in my current implementation I re-construct the flux term from the velocity history. Interestingly, the bug only appears at lower cell sizes (<0.1). 

There is also an error from the Eulerian history manager during the pre-solve update, but doesn't seem to be altering the numerical solution:
`Callback error in <function _BaseMeshVariable._create_canonical_data_array.<locals>.canonical_data_callback at 0x32a94eac0>: cannot reshape array of size 1969 into shape (2)`
**Screenshots**
SL after first timestep
![Image](https://github.com/user-attachments/assets/3ea78744-c166-4d69-84e9-6d7c686b8a26)
Eulerian after first timestep
![Image](https://github.com/user-attachments/assets/2578b6ca-3270-4ddb-bc4d-34eda22ea612)

lmoresi:
@bknight1 — Thanks for the detailed diagnosis. The SemiLagrangian vs Eulerian comparison is useful, though we suspect the Eulerian DDt may be masking the problem through numerical diffusion rather than actually fixing the underlying issue (the Eulerian advection term is only a cursory implementation and effectively lowers the Reynolds number).

To dig into this properly, could you share:

1. **A reproducer script** (or point us to the test/notebook you're running) — the solver setup, mesh, BCs, and timestepping parameters
2. **Which branch/commit** you're testing against — there are DDt changes in progress from the viscoelastic work that may touch the same code paths
3. **The timestep size and Reynolds number** — to check whether the CFL condition is being respected for the SemiLagrangian transport

The resolution dependence (only cellSize < 0.1) is interesting — finer meshes mean more swarm particles and more migration events, which could point to a swarm data synchronisation issue in the SemiLagrangian history manager. The reshape error in the Eulerian path (`cannot reshape array of size 1969 into shape (2)`) also looks like a data cache bug worth tracking separately.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

bknight1:
[...]

PR #83, opened 2026-03-18
Fix for NS solver

bknight1:
Fix for the solver #https://github.com/underworldcode/underworld3/issues/82
- Change from SL to Eulerian history manager for flux & unknown in NS solver.
- Does NOT fix the SL issue. Further investigation required.
- Does NOT fix the callback error in the Eulerian history manager.

### DIFF
--- a/src/underworld3/systems/solvers.py
+++ b/src/underworld3/systems/solvers.py
@@ -2737,29 +2737,19 @@ class SNES_NavierStokes(SNES_Stokes_SaddlePt):
         ### sets up DuDt and DFDt
         ## ._setup_history_terms()
 
-        # If DuDt is not provided, then we can build a SLCN version
+        # If DuDt is not provided, build it here. We prefer the Eulerian_DDt
+        # formulation over the SemiLagrangian alternative to keep the solver
+        # initialisation simpler and easier to maintain.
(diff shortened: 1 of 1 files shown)


lmoresi:
@bknight1 — Thanks for the reproducer script and diagnosis. We've identified and fixed the root cause on `development` (commit 4d522d8), so this workaround PR is no longer needed.

**Root cause**: The SemiLagrangian DDt's `update_pre_solve` computed upstream departure points without clamping them to the domain boundary. Near corners (especially the lid-driven cavity singularities), departure points landed outside the mesh. `global_evaluate` then extrapolated using the P2 basis, producing values ~1000x the actual field maximum. This injected a massive spurious acceleration that blew up the solver on step 2.

**Fix**: Two lines — clamp `mid_pt_coords` and `end_pt_coords` using `mesh.return_coords_to_bounds()` in `SemiLagrangian.update_pre_solve()`. The Lagrangian DDt already did this for swarm advection; the SemiLagrangian (nodal) path was missing it.

Closing this PR in favour of the direct fix on development. We'd appreciate it if you could validate the fix and do further benchmarking — including testing the second-order DuDt term. See issue #82 for details and benchmark results.

PR #84, opened 2026-03-19
Fix Stokes velocity FE dual-space setup for high-order simplex solves

gthyagi:
## Summary

This fixes a Stokes FE setup inconsistency that breaks higher-order triangular velocity solves.

In `SNES_Stokes_SaddlePt._setup_discretisation()`, the private velocity `PetscFE` was created with the requested polynomial degree, but without the matching PETSc dual-space options that UW3 already uses for mesh variables and the other solver setup paths.

This PR adds the missing velocity FE options before `PETSc.FE().createDefault(...)`:

- `private_<prefix>_u_petscdualspace_lagrange_continuity = self.u.continuous`
- `private_<prefix>_u_petscdualspace_lagrange_node_endpoints = False`

## Reproducer

A minimal reproducer script is included in this PR:

- `examples/stokes_box_mms_simplex.py`

Run it directly from the repo root, for example:

```bash
python examples/stokes_box_mms_simplex.py
python examples/stokes_box_mms_simplex.py --vdegree 2 --pdegree 1
python examples/stokes_box_mms_simplex.py --vdegree 3 --pdegree 2
```

The script uses a simple manufactured Stokes solution on the unit square with triangular elements:

- stream function: `psi = x^2 y^2`
- exact velocity: `u = (2 x^2 y, -2 x y^2)`
- exact pressure: `p = x^2 - 1/3`
- exact Dirichlet velocity is imposed on all four sides

This is a clean test because the exact velocity is cubic and the exact pressure is quadratic, so `P3/P2` should represent the solution essentially exactly on a straight box.

## Error norms before and after

Using the same box MMS setup at `h = 0.125`:

Before this fix:

- `P2/P1`: velocity `1.5068686519735996e-04`, pressure `3.9062499996873525e-03`
- `P3/P2`: velocity `1.2277197041312973e-02`, pressure `2.0668703128596483e-10`

After this fix:

- `P2/P1`: velocity `1.5068686519735996e-04`, pressure `3.9062499996873525e-03`
- `P3/P2`: velocity `3.6635897127143294e-12`, pressure `1.796728310502028e-10`

So the fix leaves `P2/P1` unchanged and restores `P3/P2` to the expected near-machine-precision result on this exact polynomial test.

## Notes

This PR is intentionally minimal:

- one solver-side FE setup fix
- one small reproducer script

It does not include the separate imported-mesh / XDMF work that came up during debugging.

### DIFF
--- a/examples/stokes_box_mms_simplex.py
+++ b/examples/stokes_box_mms_simplex.py
@@ -1,7 +1,7 @@
 """
-Simple manufactured-solution Stokes example on a triangular box.
+Notebook-style manufactured-solution Stokes example on a triangular box.
 
-This script is useful for checking higher-order simplex velocity / pressure
+This example is useful for checking higher-order simplex velocity / pressure
 pairs on a straight-edged domain without curved-geometry effects.
 
 Default setup:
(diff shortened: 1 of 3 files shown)


lmoresi (approved):
Reviewed the fix and audited all `PETSc.FE().createDefault()` call sites across the codebase.

**Stokes velocity fix (Tyagi's original commit):** Correct and minimal. The Stokes velocity FE was the only solver field missing the dual-space options — Scalar solver, Vector solver, and the Stokes pressure FE all had them already.

**Added in follow-up commit:**
- Comment explaining *why* these options matter (node placement on simplices, P2 works by accident, P3+ breaks)
- Same fix applied to the mesh coordinate projection FE in `discretisation_mesh.py` — same omission, currently harmless for P1 coordinate meshes but would break higher-order (curved) coordinate meshes on simplices

All other FE creation sites (MeshVariable, gradient evaluation, field projection) already set the dual-space options correctly.

PR #85, opened 2026-03-19
Fix SemiLagrangian DDt blowup: clamp departure points to domain

### LINKED ISSUE #82: [BUG] - Navier-stokes solution unstable after first timestep

lmoresi:
## Summary

- Fix SemiLagrangian DDt upstream departure point extrapolation bug (fixes #82)
- Rewrite NS pipe flow example as standalone notebook

## Root cause

The SemiLagrangian `update_pre_solve` computed upstream departure points (`mid_pt_coords` and `end_pt_coords`) without clamping them to the domain boundary. Near corners -- especially velocity discontinuities in the lid-driven cavity -- departure points landed outside the mesh. `global_evaluate` then extrapolated using the P2 basis, producing values ~1000x the actual field maximum. This injected a spurious acceleration that blew up the solver on the second timestep.

## Fix

Clamp both `mid_pt_coords` and `end_pt_coords` using `mesh.return_coords_to_bounds()` in `SemiLagrangian.update_pre_solve()`. The Lagrangian DDt already did this for swarm advection; the SemiLagrangian (nodal) path was missing it. Two lines of code.

## Validation

**Stability**: All combinations stable (Re = 10, 100, 400, 1000; cellSize = 0.1, 0.05, 0.025).

**Ghia et al. (1982) benchmark** at Re=100 (cellSize=0.04, 200 steps, t=10): RMS error = 0.004 against published centreline velocity data.

Benchmark images (Re=100 and Re=400 streamlines) to follow in comments.

## Test plan

- [x] Stability across Re and cellSize combinations
- [x] Ghia et al. (1982) centreline velocity comparison (Re=100)
- [x] Re=400 streamline visualisation (corner vortices visible)
- [ ] @bknight1 to validate with original reproducer script
- [ ] Test with order=2 (second-order DuDt)
- [ ] Check that the Eulerian DDt callback reshape error is a separate issue

### DIFF
--- a/docs/examples/fluid_mechanics/intermediate/Ex_NS_PipeFlow.py
+++ b/docs/examples/fluid_mechanics/intermediate/Ex_NS_PipeFlow.py
@@ -1,351 +1,174 @@
 # %% [markdown]
 """
-# 🔬 NS PipeFlow
+# Navier-Stokes: Poiseuille (Pipe) Flow
 
-**PHYSICS:** fluid_mechanics  
-**DIFFICULTY:** intermediate  
-**MIGRATED:** From underworld3-documentation/Notebooks
+Transient development of parabolic flow between parallel plates driven by
+a uniform inlet velocity. The flow starts from rest and evolves toward
+the classical Poiseuille profile.
 
-## Description
-This example has been migrated from the original UW3 documentation.
-Additional documentation and parameter annotations will be added.
+This is a useful test case because:
 
-## Migration Notes
-- Original complexity preserved
-- Parameters to be extracted and annotated
-- Claude hints to be added in future update
-"""
+- The steady-state solution is known analytically
+- The transient spin-up exercises the time-derivative (DDt) machinery
+- The Reynolds number controls how quickly inertial effects appear
 
-# %% [markdown]
-"""
-## Original Code
-The following is the migrated code with minimal modifications.
+**Reference:** Cengel, Y. A. (2010). *Fluid Mechanics: Fundamentals
+and Applications*. Tata McGraw Hill.
+
+**Original example by** Juan Carlos Graciosa.
 """
 
 # %%
-# ---
-# jupyter:
-#   jupytext:
-#     cell_metadata_filter: -all
-#     custom_cell_magics: kql
-#     text_representation:
-#       extension: .py
-#       format_name: percent
-#       format_version: '1.3'
-#       jupytext_version: 1.11.2
-#   kernelspec:
-#     display_name: uw3-venv-run
-#     language: python
-#     name: python3
-# ---
-
-# %% [markdown]
-# # Navier Stokes benchmark: 2D pipe flow
-# By: Juan Carlos Graciosa 
-# &nbsp;
-# &nbsp;
-#
-# References:  
-# - https://www.fifty2.eu/innovation/planar-poiseuille-flow-2-d-in-preonlab/ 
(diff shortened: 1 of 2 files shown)


lmoresi:
Here are two quick "benchmark" runs at Re=100, Re=400 for the Ghia et al benchmark. Qualitatively, this looks promising with the counterflow vortices developing at approximately the correct size ratios. @bknight1 - needs more testing, but merge if you think the main bugs are killed. 

NS_Re100_streamlines
NS_Re400_streamlines

bknight1:
Looks good to me! Thanks @lmoresi. 

A quick suggestion would be allowing the user to pass a DFDt history manager, as we currently enforce the SL version. We currently do this for all solvers that require the flux history.

lmoresi:
## Order-1 and Order-2 validation: Lid-driven cavity Re=100

Tested both `order=1` and `order=2` SemiLagrangian on the lid-driven cavity benchmark (cellSize=0.04, 200 steps, dt=0.05, t_final=10) with the departure-point clamping fix applied.


### Results

| Order | RMS error vs Ghia et al. (1982) |
|-------|-------------------------------|
| 1     | 0.0038                        |
| 2     | 0.0415                        |

### Order 1
sl_cavity_order1

### Order 2
sl_cavity_order2

**Note on order-2**: This does not really look stable, so we will keep this open for now and keep digging

lmoresi:
**Update: Order-2 results with working BDF/AM coefficients**

  The earlier order-2 results were invalid. We discovered that BDF/AM coefficients were frozen at order 1 from the first solve — `effective_order` ramping had no effect on the compiled pointwise functions. All three order-2 BDF/AM combinations produced identical RMS (0.041469) because they were all running BDF1+AM1 with extra history noise.

  This has been fixed in PR #88 (BDF/AM coefficients routed through `PetscDSSetConstants()`). With the fix applied, order-2 (BDF2+AM2) produces a
  clean solution:

sl_cavity_order2_streamlines



  | Order | RMS vs Ghia et al. (1982) |
  |-------|--------------------------|
  | 1 (BDF1+AM1) | 0.0038 |
  | 2 (BDF2+AM2) | 0.0065 |

  Both orders are stable with the departure-point clamping fix from this PR. The order-2 instabilities seen previously were entirely the frozen-coefficients bug, not a problem with the clamping.

  - [x] Test with order=2 (second-order DuDt) — **validated, clean solution**

lmoresi:
I think we can merge this fix. Test passing, comments from @bknight1 approving 1st order variant of the PR

PR #89, opened 2026-03-22
Explicit stress history for viscoelastic Stokes solver

lmoresi:
## Summary

Explicit stress history architecture for VE_Stokes: store actual computed stress
in psi_star (rather than flux), advect stored values along characteristics, and
use BDF-1/2/3 discretisation of the Maxwell constitutive law.

### Key changes

- **Explicit stress history**: after each solve, the actual deviatoric stress is
  projected into psi_star[0] and shifted down the history chain. This replaces
  the previous approach of storing/advecting the flux expression.
- **solver.tau property**: access the computed stress tensor (projected, not the
  symbolic formula)
- **BDF-3 constitutive formulae**: ve_effective_viscosity, E_eff, and stress()
  now have three-way branching for BDF-1/2/3 with correct coefficients
- **effective_order property**: ramps from 1 to requested order as DDt history
  accumulates, preventing use of higher-order coefficients before distinct
  history is available
- **dt_elastic is independent of timestep**: solve(timestep=) no longer
  overwrites the constitutive relaxation parameter
- **Analytical formula fix**: removed spurious De factor in Maxwell oscillatory
  solution (was invisible at De=1, caused apparent errors at other De)
- **Benchmark documentation**: oscillatory shear validation with convergence
  tables and resolution study

### Convergence (constant shear, Maxwell analytical)

| dt/t_r | BDF-1 | BDF-2 | BDF-3 |
|--------|-------|-------|-------|
| 0.200 | 3.0e-2 | 4.0e-3 | 6.4e-3 |
| 0.100 | 1.5e-2 | 9.3e-4 | 1.7e-3 |
| 0.050 | 7.8e-3 | 2.3e-4 | 4.3e-4 |
| 0.020 | 3.1e-3 | 3.7e-5 | -- |

BDF-2 is the recommended default (second-order convergence, 85x better than
BDF-1 at small dt).

### Oscillatory validation

A shear layer with a harmonic (in time) velocity boundary condition. For a Maxwell material, the analytic solution can be derived - it shows a phase lag in stress compared to the driving velocity, and the amplitude evolves with time. 

The Maxwell constitutive law for simple shear:

  $$\dot{\sigma}{xy} + \frac{\sigma{xy}}{t_r} = \mu \dot{\gamma}_0 \sin(\omega t)$$

  where $t_r = \eta/\mu$ is the relaxation time and $\text{De} = \omega t_r$ is the Deborah number.

  Full solution with $\sigma(0) = 0$:

  $$\sigma_{xy}(t) = \frac{\eta \dot{\gamma}_0}{1 + \text{De}^2}\left[\sin(\omega t) - \text{De} \cdot \cos(\omega t) + \text{De},e^{-t/t_r}\right]$$

  Steady-state amplitude: $A = \eta \dot{\gamma}_0 / \sqrt{1 + \text{De}^2}$

  Phase lag: $\delta = \arctan(\text{De})$

ve_oscillatory_De5_resolution

### Notes on dt_elastic separation

A running-average approach for history accumulation when dt << dt_elastic was
investigated but found to be extremely diffusive for semi-Lagrangian transport
and is not implemented. To prevent unstable behaviour when timesteps become
small, we advise limiting the minimum effective viscosity in line with the
physics of the problem.

## Test plan

- [x] Constant shear benchmark (order 1, 2, 3) against Maxwell analytical
- [x] Oscillatory shear validation (De=1, 1.5, 5) with phase lag and amplitude
- [x] Resolution study (De=5, three timestep sizes)
- [x] VP and VEP shear box tests
- [ ] CI pipeline


lmoresi:
We seem to be getting pretty close now. Here is a "benchmark" run with approximately a square wave (fourier series) boundary condition and variable timestepping. The black line is the series solution (aka analytic), the red dots are the variable timestep solution. There blue dots are not easily visible, they are the small timestep solution.

ve_square_wave_10h

We seem to be capturing the elastic behaviour quite well. Well enough to move on to the next set of physics, I would think.

PR #93, opened 2026-03-25
Refactor JIT cache layer with JITCallbackSet

lmoresi:
## Summary

Builds on PR #92 (@gthyagi) which fixed the Integral/BdIntegral JIT cache collision bug. This PR refactors the fix into a reusable architectural pattern that simplifies the JIT interface for all callers.

## What PR #92 identified

When `Integral(fn=1.0)` and `BdIntegral(fn=1.0)` ran on the same mesh, the JIT cache key was a flat tuple of all expressions — losing which callback role each expression belonged to. The second call reused the first's cached module, wiring the wrong PETSc callback slot and returning `0.0`.

## What this PR does

Introduces `JITCallbackSet`, a frozen dataclass that groups the five PETSc callback lists (residual, bcs, jacobian, bd_residual, bd_jacobian) into a single structured container.

**Before (PR #92 fix):** The cache key was made structured by manually building parallel tuples for each role inside `getext()`. Callers still passed 5 positional lists:

```python
# Integral — which positional arg is which?
getext(mesh, [self.fn,], [], [], [], [], mesh.vars.values())
# BdIntegral
getext(mesh, [], [], [], [self.fn,], [], mesh.vars.values())
```

**After:** Callers declare intent by name, with a clean 3-argument interface:

```python
# Integral — self-documenting
getext(mesh, JITCallbackSet(residual=(self.fn,)), mesh.vars.values())
# BdIntegral
getext(mesh, JITCallbackSet(bd_residual=(self.fn,)), mesh.vars.values())
```

The structured cache key is now a data structure invariant — you can't construct a callback set without declaring roles, so the collision bug cannot recur.

### Changes

- **`JITCallbackSet`** dataclass with `flat()`, `signature()`, `map()`, `counts`
- **`prepare_for_cache_key()`** — extracted as a module-level function (was an inline closure). Two-phase expression preparation: constants → placeholders, then unwrap to pure SymPy. Reusable by any code needing structural expansion for hashing.
- **`getext(mesh, callbacks, primary_field_list, ...)`** — clean 3-positional-arg signature
- **`_createext(name, mesh, callbacks, primary_field_list, ...)`** — takes `JITCallbackSet` directly, uses `callbacks.counts` instead of `len(fns_residual)` etc.
- All 7 call sites updated: 3 solvers, 3 integrals, 1 test file

### Files changed

| File | Change |
|------|--------|
| `src/underworld3/utilities/_jitextension.py` | Add `JITCallbackSet`, `prepare_for_cache_key()`; refactor `getext()` and `_createext()` |
| `src/underworld3/cython/petsc_maths.pyx` | Integral/BdIntegral use `JITCallbackSet` |
| `src/underworld3/cython/petsc_generic_snes_solvers.pyx` | Scalar/Vector/Stokes solvers use `JITCallbackSet` |
| `tests/test_0004_pointwise_fns.py` | Update direct `getext()` calls |
| `tests/test_0502_boundary_integrals.py` | PR #92 regression tests (spherical shell cache collision) |

## Test plan

- [x] 21/21 boundary integral tests pass (including PR #92 spherical shell regression)
- [x] 18/18 core solver tests pass (Poisson, Stokes, VE-Stokes)
- [x] 374/374 level_1 tests pass (0 failures, 7 skipped, 1 xfail)

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -9,7 +9,7 @@ from petsc4py import PETSc
 
 import underworld3
 import underworld3 as uw
-from   underworld3.utilities._jitextension import getext
+from   underworld3.utilities._jitextension import getext, JITCallbackSet
 import underworld3.timing as timing
 
 from underworld3.utilities._api_tools import uw_object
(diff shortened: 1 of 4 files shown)


lmoresi:
This is just some tidying prompted by @gthyagi 's bug fix. Hardening the JIT infrastructure is always and improvement.

lmoresi:
Copilot fixes adopted. I think we can merge this as this passes all CI tests and it is really quite a simple task.

PR #99, opened 2026-03-30
Fix MPI BdIntegral on deformed spherical shells

gthyagi:
## Summary
- treat empty local boundary strata the same as missing boundary labels in `UW_DMPlexComputeBdIntegral`
- add an MPI regression test covering `BdIntegral` on a deformed spherical shell after `_deform_mesh()`
- keep existing parallel annulus boundary-integral coverage passing

## Verification
- `mpirun -np 2 /Users/tgol0006/.pixi/bin/pixi run -e amr-dev pytest tests/parallel/test_0765_internal_boundary_integral_mpi.py -q --with-mpi`
- `/Users/tgol0006/.pixi/bin/pixi run -e amr-dev pytest tests/test_0502_boundary_integrals.py -q -k spherical`

### DIFF
--- a/src/underworld3/cython/petsc_compat.h
+++ b/src/underworld3/cython/petsc_compat.h
@@ -173,15 +173,26 @@ PetscErrorCode UW_DMPlexComputeBdIntegral(DM dm, Vec X,
 {
     PetscSection  section;
     PetscInt      Nf;
+    PetscInt      localCount = 0;
 
     PetscFunctionBeginUser;
 
     PetscCall(DMGetLocalSection(dm, &section));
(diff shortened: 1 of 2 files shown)


lmoresi (approved):
Fix is correct — checks local stratum size before entering the Allreduce path, handling the general case of empty local boundary strata under MPI partitioning. Test covers the deformed-shell scenario that surfaced it. Approved.

lmoresi (approved):
This fixes an issue that was surfaced when dealing with deforming meshes but which is also quite general and is related to the surface integral code approach that we worked on recently. This should be accepted as is.

PR #103, opened 2026-03-31
Use PETSc CXX for C++ extension builds

gthyagi:
## Summary
- read `CXX` from PETSc `petscvariables` in `setup.py`
- export `CXX` alongside the existing `CC` override

## Why this is important
Pre-fix, `setup.py` only forced PETSc's `CC`. That meant C builds followed PETSc, but C++ builds could still fall back to the ambient environment compiler.

In practice, this can send `underworld3.ckdtree` through a different or broken C++ toolchain than the PETSc/MPI build, especially in Pixi/OpenMPI environments.

Using both PETSc `CC` and `CXX` keeps the Underworld extension build aligned with the PETSc toolchain.

## Validation
- `python3 -m py_compile setup.py`
- `OMPI_CC=/usr/bin/clang OMPI_CXX=/usr/bin/clang++ pixi run -e amr-dev pip install . --no-build-isolation`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -142,17 +142,22 @@ def configure():
 
     LIBRARIES += ["petsc"]
 
-    # set CC compiler to be PETSc's compiler.
-    # This ought include mpi's details, ie mpicc --showme,
-    # needed to compile UW cython extensions
-    compiler = ""
+    # Set C and C++ compilers to PETSc's toolchain so the UW extensions
+    # build with the same compiler family and MPI wrappers as PETSc.
+    cc = ""
+    cxx = ""
     with open(petscvars, "r") as f:
         for line in f:
             line = line.strip()
             if line.startswith("CC ="):
-                compiler = line.split("=", 1)[1].strip()
-    # print(f"***\n The c compiler is: {compiler}\n*****")
-    os.environ["CC"] = compiler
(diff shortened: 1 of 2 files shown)


gthyagi:
Added a follow-up fix for runtime JIT builds. The earlier  change made package builds use PETSc's , but  was still spawning temporary  builds without that compiler environment. This commit now passes PETSc-derived  into the JIT subprocess and sets  when the OpenMPI wrapper points to a missing compiler name. Verified by rerunning the three previously failing collection-time tests: , , and  ().

gthyagi:
Added a follow-up fix for runtime JIT builds.

The earlier `setup.py` change made package builds use PETSc `CC/CXX`, but `_jitextension.py` was still spawning temporary `setup.py build_ext --inplace` builds without that compiler environment. This commit now passes PETSc-derived `CC/CXX` into the JIT subprocess and sets `OMPI_CC/OMPI_CXX` when the OpenMPI wrapper points to a missing compiler name.

Verified by rerunning the three previously failing collection-time tests:
- `tests/test_0050_utils.py`
- `tests/test_1004_DarcyCartesian.py`
- `tests/test_1011_stokesSph.py`

Result: `6 passed`.

PR #124, opened 2026-04-20
Avoid PETSc DM churn: multi-component projector + in-place residual swap

lmoresi:
## Summary

Two complementary changes that reduce PETSc DM/SNES churn in solver rebuild paths — addressing UW3's long-standing object-lifetime / memory-leak issues from repeatedly destroying and recreating DMs on every parameter or component change.

- **Multi-component projector** — `SNES_MultiComponent` base solver and `SNES_MultiComponent_Projection` solve N scalar components in a single SNES on one DM, replacing the per-component cycling in `SNES_Tensor_Projection` for the VE Stokes tau projection (one DM build per outer solve instead of Nc per outer solve).
- **In-place residual / jacobian swap** — `_build()` now branches on three rebuild flags (`_needs_dm_rebuild`, `_needs_bc_reregister`, `_needs_function_rewire`). Constitutive-parameter changes, DuDt/DFDt swaps, and similar function-only invalidations re-run JIT and call `PetscDSSetResidual`/`SetJacobian` on the existing DS — no DM teardown, no SNES recreation, no coarse-grid hierarchy rebuild.
- **PETSc Jacobian layout fix** — corrects a latent `[fc, dg, df, gc]` vs `[fc, gc, df, dg]` bug in `SNES_Vector` that was hidden by F1-index symmetry in every in-repo consumer but would fire for any subclass with non-symmetric F1.
- **Developer documentation** — `docs/developer/subsystems/petsc-jacobian-layout.md` captures PETSc's pointwise-Jacobian index convention, the sympy-to-PETSc axis mapping, an audit of every existing solver, and a checklist for new solvers.

## Multi-component projector

- `SNES_MultiComponent` decouples the FE component count from `mesh.dim`; the unknown is a `(1, Nc)` MATRIX variable. Jacobians are built with explicit nested-loop construction directly into PETSc's `[fc*Nc + gc, df*dim + dg]` row-major 2D layout — no `permutedims`, no sympy-convention translation step.
- `SNES_VE_Stokes._setup_tau_projection` rewired to use the multi-component projector for the symmetric-tensor path. User-facing `_tau_var` remains a `SYM_TENSOR` so downstream `.array[:, i, j]` reads are unchanged; a flat `(1, Nc)` variable drives the actual solve and results fan out into the sym-tensor after each solve.
- `SNES_Vector._setup_pointwise_functions` (both main residual and natural-BC Jacobian paths) migrated to the same explicit-index construction pattern.
- `SNES_Scalar` (Nc=1) and `SNES_Stokes_SaddlePt` (permutation `(0, 2, 1, 3)`) already use correct layouts and are unchanged. `SNES_NavierStokes` inherits Stokes.

## In-place residual swap

- Three rebuild flags on `SolverBaseClass` replace the single `is_setup` bool. `is_setup` becomes a derived `@property` whose setter pessimistically raises all three flags, so every existing `self.is_setup = False` call site keeps its current full-rebuild semantics — backward-compatible by default, narrow opt-in for confirmed function-only sites.
- Opt-in sites: `Unknowns.DuDt.setter`, `Unknowns.DFDt.setter`, the `solve()` paths' `not constitutive_model._solver_is_setup` branch (in all four SNES subclasses), and `Constitutive_Model._reset()` (fires on every Parameter mutation).
- `_build()` branches: full DM teardown when DM/BC dirty, else fast path that re-runs `_setup_pointwise_functions` and `_setup_solver(..., _rewire_only=True)`. The rewire path keeps the existing `PetscDSSet*` calls (now actually replacing pointers in place), re-runs `_set_constants_on_ds`, and propagates the updated DS to coarse DMs via `copyDS` for multigrid.
- All four `_setup_solver` overrides (`SNES_Scalar`, `SNES_Vector`, `SNES_MultiComponent`, `SNES_Stokes_SaddlePt`) gated on `_rewire_only`. Stokes also skips `createFieldDecomposition` and `_attach_stokes_nullspace` on the rewire path — saddle-point block layout is unchanged.

## Why this matters

Wall-clock savings on the rewire path are modest (~25% per build when JIT cache hits, marginal when JIT misses — JIT recompilation is the dominant cost). The primary win is **structural**: every constitutive-parameter tweak, every solve-loop iteration, every nonlinear Picard step on the rewire path now reuses a single DM/SNES/coarse-grid hierarchy instead of allocating and destroying a new one. This directly addresses UW3's PETSc object-lifetime and memory-leak problems on long runs (time-stepping, parameter sweeps, component cycling).

## Test plan

- [x] 10 validation tests for `SNES_MultiComponent_Projection` in `tests/test_multicomponent_projection.py`:
  - Nc=1 agrees with `SNES_Projection` (1e-7 rel-L2)
  - Nc=3 sym-tensor agrees with `SNES_Tensor_Projection` (1e-7 rel-L2)
  - Nc=4 full-tensor agrees with `SNES_Tensor_Projection` (1e-7 rel-L2)
  - DM-rebuild count scales with outer solves, not Nc × outer solves
  - Smoothing-parametrised agreement with legacy solvers at 1e-4, 1e-2, 1.0
  - High-frequency damping with non-zero smoothing
- [x] 6 regression tests for `SNES_Vector` asymmetric-F1 layout in `tests/test_snes_vector_asymmetric_jacobian.py`:
  - Identical targets give identical components (would fail with the old layout)
  - Agreement with `SNES_MultiComponent_Projection` on different targets
  - Parametrised over smoothing ∈ {1e-4, 1e-2, 1e-1}
- [x] In-place rewire smoke tests:
  - Poisson: halving diffusivity gives steady-state T_max ratio 2.0000 (exact); no "Destroy solver DM" on the rewire path
  - Stokes: halving viscosity gives |v|_max ratio 2.0010 — two-field rewiring correct
  - Parallel `mpirun -np 2`: ratio 2.0000 — coarse-DS `copyDS` path intact
- [x] Full Stokes (`test_1010_stokesCart`) and VE Stokes (`test_1050_VEstokesCart`) suites pass — no regression on symmetric-F1 production paths.
- [x] tier_a level_1 suite: same pass/fail set as `development` baseline. The 16 `MultiComponent_Projection` AttributeError failures reproduce identically on baseline (test-collection contamination from a stale `.claude/worktrees/fault-system-workflow/` env, unrelated to this PR).
- [x] `docs-build` passes with the new page in the toctree.
- [ ] Reviewer run: verify VE benchmark (`docs/advanced/benchmarks/run_ve_square_wave.py`) wall-clock and accuracy are unchanged or improved vs `development`.

## Related planning items

Addresses "Avoid PETSc DM/SNES teardown in tensor projection loop (2026-03-25)" in the external planning file via **both** approaches proposed there: the multi-component projector for the tensor case, and `_swap_residual` (now `_needs_function_rewire`) as the general mechanism for any constitutive-only change.


lmoresi:
## VE_Stokes bypasses the multicomponent tau projection

The `SNES_Stokes.tau` property at `solvers.py:1294` overrides the base class `tau`:

```python
@property
def tau(self):
    if self.Unknowns.DFDt is not None:
        return self.DFDt.psi_star[0]   # VE path — bypasses base class
    return super().tau                  # non-VE path — uses base class
```

For VE solvers (`DFDt is not None`), the stress comes from `psi_star[0]` which is projected by the DDt's internal `_psi_star_projection_solver`. This is a separate `SNES_Tensor_Projection` that was NOT updated to use `SNES_MultiComponent_Projection`.

The profiling confirms: `SNES_Tensor_Projection.solve` = 100 calls, `SNES_Scalar` = 300 (3 components × 100). The multicomponent projector at line 1069 of the pyx is never reached.

The fix: the DDt stress projection (at `solvers.py:1269`) also needs to use `SNES_MultiComponent_Projection`, or the `SNES_Stokes.tau` override needs to route through the base class multicomponent path for VE too.

Reference: `solvers.py:1269`:
```python
self.DFDt._psi_star_projection_solver.uw_function = self.constitutive_model.flux
```

lmoresi:
Copilot review notes addressed:

1. **`cdim != dim` guard**: Agree this is too restrictive. Surface meshes (2D embedded in 3D) are a valid use case. Should be relaxed or removed — the solver can handle `cdim != dim` as long as the gradient operator is constructed correctly for the `dim` components.

2. **`_dm_build_count`**: Agree — should be a property. Minor cleanup.

Both are low-risk changes that can be done before merge.

PR #144, opened 2026-04-25
Make JIT source generation deterministic

gthyagi:
## Summary

This PR makes UW3 JIT source generation deterministic across MPI ranks by applying stable ordering when collecting symbols, constants, mesh variables, generated includes, and setup metadata.

## Motivation

Some parallel runs can generate rank-dependent C source ordering from unordered Python collections. That can trigger JIT hash mismatches even when the mathematical expression is identical on all ranks.

## Changes

- Add stable sorting helpers for JIT source generation inputs.
- Preserve PETSc field ordering by prioritising `field_id` when sorting mesh variables.
- Add regression tests for deterministic ordering.

## Validation

- `./uw python -m pytest tests/test_jit_deterministic_ordering.py tests/test_jit_cache.py -q`
- 9 passed
- Kramer spherical free-slip case was reproduced locally without requiring `PYTHONHASHSEED=0`.

## Scope

This should not change solver mathematics. It only makes generated JIT source ordering stable across ranks.

### DIFF
--- a/src/underworld3/utilities/_jitextension.py
+++ b/src/underworld3/utilities/_jitextension.py
@@ -11,6 +11,36 @@ from dataclasses import dataclass
 from pathlib import Path
 
 
+def _stable_sort_key(obj):
+    """Return a process-stable key for ordering symbolic/JIT objects.
+
+    Python hash randomisation means unordered containers such as ``set`` can
+    iterate differently on different MPI ranks. JIT source emission must not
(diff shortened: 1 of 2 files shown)


lmoresi:
Excellent discovery / fix !

lmoresi (approved):
Yep, good catch.  I thought that we'd fix things by not compiling across ranks, but this still leaves a problem if the cache cannot be located consistently

PR #145, opened 2026-04-25
Fix local macOS OpenMPI PETSc toolchain setup

gthyagi:
## Summary

This PR makes the local macOS OpenMPI PETSc build setup use a coherent compiler/toolchain configuration.

## Motivation

On macOS, local OpenMPI PETSc builds can mix pixi/conda compiler wrappers, Apple clang, Homebrew gfortran, and SDK/deployment-target settings. That can produce an inconsistent PETSc configure/build environment.

## Changes

- Applies only for `CLUSTER=local`, Darwin, and OpenMPI.
- Selects the external Fortran compiler consistently.
- Derives SDK/deployment target information from the selected compiler environment.
- Clears conflicting pixi/conda compiler variables for this local macOS path.
- Sets OpenMPI wrapper compiler variables explicitly before PETSc configure/build/test/petsc4py steps.

## Scope

This path is local macOS/OpenMPI-specific. It should not affect Gadi/HPC builds, Linux local builds, or MPICH builds.

## Local environment where this was needed

- macOS arm64
- Apple clang 21.0.0
- Homebrew gfortran 14.2.0_1
- pixi OpenMPI 5.0.10

### DIFF
--- a/petsc-custom/build-petsc.sh
+++ b/petsc-custom/build-petsc.sh
@@ -218,6 +218,70 @@ setup_gadi_build_env() {
     export PATH="/usr/bin:/usr/local/bin:${MPI_DIR}/bin:${PATH}"
 }
 
+# ── Local macOS/OpenMPI build environment setup ──────────────────────────────
+# The pixi osx-arm64 toolchain currently mixes a conda clang/ld stack, OpenMPI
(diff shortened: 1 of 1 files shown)


lmoresi (approved):
Seems clean - merging

Issue #138, opened 2026-04-23
Document and review spherical benchmark output/restart paths (write_timestep vs write_checkpoint)


gthyagi:
## Summary

This issue documents the current UW3 mesh/field output methods used in the spherical benchmark workflow and requests feedback on a restart-oriented follow-up design.

The immediate motivation is large spherical benchmark postprocessing, where:

- `write_timestep()` output works with `read_timestep(...)`
- but that reload path is expensive because it reconstructs fields from saved arrays and coordinates using a KDTree-based path
- exact restart-style reload needs PETSc restart metadata that is not written by `write_timestep()`

This issue is for documentation and design review only. No implementation is proposed until feedback is received.

@lmoresi could you please review the comparison below and comment on the proposed direction? I plan to wait for feedback before making implementation changes.

### What it writes

Typical output:

- one mesh HDF5 file
- one HDF5 file per mesh variable
- optional XDMF file
- ParaView-compatible groups such as `/vertex_fields`

Typical file set:

- `output.mesh.00000.h5`
- `output.mesh.Velocity.00000.h5`
- `output.mesh.Pressure.00000.h5`
- `output.mesh.00000.xdmf`

Typical variable-file structure:

- `/fields/<name>`
- `/fields/coordinates`
- `/vertex_fields/...`

### How reload works

`read_timestep(...)` does **not** perform an exact PETSc restart.

Instead it:

1. opens the variable HDF5 file with `h5py`
2. reads the saved field values from `/fields/<name>`
3. reads the saved coordinates from `/fields/coordinates`
4. builds a KDTree on the saved coordinates
5. reconstructs values on the current mesh variable by coordinate-based matching / interpolation

So this method works because the file contains field values and coordinates. It does not require PETSc section/order metadata.

### Advantages

- Produces the standard benchmark output files already used by the repo
- Generates XDMF for ParaView / visualisation
- Works with `read_timestep(...)`
- Flexible when reloading onto a different mesh or ordering because the reload path is coordinate-based

### Disadvantages

- Reload is expensive because `read_timestep(...)` uses `h5py` plus KDTree reconstruction
- At large MPI size this can become very memory-heavy
- Does not write the PETSc subDM section/order metadata needed for exact FE vector restart
- Direct PETSc vector load from the `/fields/...` datasets is not reliable for exact restart

### What it writes

Typical output:

- one mesh HDF5 file
- one restart/checkpoint HDF5 file

Typical file set:

- `restart.mesh.0.h5`
[...]

PR #146, opened 2026-04-25
Add PETSc DMPlex checkpoint reload for mesh variables

gthyagi:
## Summary

This PR adds a PETSc DMPlex checkpoint reload path for UW3 mesh variables and documents when to use checkpoint output versus timestep/visualisation output.

## Motivation

`write_timestep()` / `read_timestep()` is useful for visualisation and flexible coordinate remapping, but its KDTree/remap path is memory-heavy for large postprocessing jobs. Spherical benchmark postprocessing at `1/128` on Gadi used close to the full 4.5 TB allocation with the old reload path.

This PR adds a restart/postprocessing-oriented path that reloads variables through PETSc DMPlex section/vector metadata instead of KDTree remapping.

## Changes

- Add `MeshVariable.read_checkpoint(...)` for PETSc DMPlex section/vector reload.
- Preserve and compose the topology-load `PetscSF` needed by PETSc DMPlex checkpoint reload.
- Write checkpoint files with PETSc DMPlex HDF5 storage version `3.0.0`.
- Add `outputPath` support to `Mesh.write_checkpoint(...)`.
- Default `write_checkpoint()` to one HDF5 file per mesh variable.
- Keep `separate_variable_files=False` support for a combined variable checkpoint file.
- Add scalar, vector, and discontinuous variable roundtrip test coverage.
- Document the checkpoint reload design and compare output/reload methods.

## File layout

Default checkpoint output now uses:

```text
checkout.mesh.00000.h5
checkout.Velocity.00000.h5
checkout.Pressure.00000.h5
```

Combined variable output remains available with `separate_variable_files=False`:

```text
checkout.checkpoint.00000.h5
```

## Validation

Local tests:

```text
./uw python -m pytest tests/test_0003_save_load.py -q
5 passed

mpirun -np 2 ./uw python -m pytest tests/test_0003_save_load.py::test_meshvariable_checkpoint_roundtrip -q
checkpoint roundtrip passed under 2 MPI ranks
```

Gadi spherical benchmark evidence is documented in:

- `docs/developer/design/petsc-dmplex-checkpoint-reload-plan.md`
- `docs/developer/subsystems/checkpoint-output-and-reload-methods.md`

Key `1/128` result:

| Method | NCPUs | Walltime | Memory used |
| --- | ---: | ---: | ---: |
| `write_timestep/read_timestep` | 1152 | `00:13:55` | `3.92 TB` |
| `write_checkpoint/read_checkpoint` | 1152 | `00:03:57` | `1.83 TB` |

Velocity, pressure, and normal-velocity metrics matched to roundoff in the benchmark comparison. Boundary `sigma_rr` uses a benchmark-side post-reload stress-recovery path and is documented separately from the checkpoint reload mechanism.

## Notes

The full `tests/test_0003_save_load.py` module under MPI includes broader swarm save/load behavior and was not used as the MPI checkpoint validation target. The checkpoint-specific roundtrip test passes under 2 MPI ranks.

### DIFF
--- a/docs/developer/design/petsc-dmplex-checkpoint-reload-plan.md
+++ b/docs/developer/design/petsc-dmplex-checkpoint-reload-plan.md
@@ -1,0 +1,426 @@
+# PETSc DMPlex Checkpoint Reload Plan
+
+## Objective
+
+Implement and validate an exact PETSc DMPlex checkpoint reload path for UW3
+mesh variables.
+
+The target workflow is:
+
+```python
+mesh = uw.discretisation.Mesh("restart.mesh.0.h5", ...)
+v_soln = uw.discretisation.MeshVariable("Velocity", mesh, ...)
+p_soln = uw.discretisation.MeshVariable("Pressure", mesh, ...)
+
+v_soln.load_from_checkpoint(
+    "restart.checkpoint.00000.h5",
+    data_name="Velocity",
+)
+p_soln.load_from_checkpoint(
+    "restart.checkpoint.00000.h5",
+    data_name="Pressure",
+)
+```
+
+This reload path must not use KDTree remapping. It must restore the saved FE
+vectors using PETSc DMPlex topology, section, and vector metadata.
+
+## Commit And Test Workflow
+
+Use small, meaningful commits while implementing this work.
+
+Do `git add` and `git commit` after each coherent feature, fix, or debugging
+checkpoint so progress is easy to inspect and bisect. Do not wait until the end
+to commit everything as one large change, and do not commit after every tiny
+edit.
+
+Create or update unit tests whenever they are needed to prove the behavior or
+prevent regressions. Prefer adding a failing regression test before the fix
+when the failure mode is already known.
+
+Commit at useful milestones, such as:
+
+- adding the first failing regression test
+- adding storage-version `3.0.0` write support
+- exposing or preserving the checkpoint topology SF
+- adding the first working `load_from_checkpoint(...)` helper
+- adding MPI same-rank test coverage
+- adding different-rank reload coverage or documenting its limitation
+- integrating the helper into benchmark-facing workflow tests
+
+Each commit message should describe the specific feature, fix, or debugging
+result included in that commit.
+
+## Problem Statement
+
+UW3 currently has two related but different output paths:
+
+- `mesh.write_timestep(...)`
+- `mesh.write_checkpoint(...)`
+
+`write_timestep()` is visualization / remap oriented. Its reload path,
+`MeshVariable.read_timestep(...)`, reads `/fields/<name>` and
+`/fields/coordinates`, builds a KDTree, and maps values onto the current mesh.
+That is flexible, but it is expensive and memory-heavy at high MPI counts.
+
+`write_checkpoint()` writes PETSc DMPlex restart metadata, including section
+and vector information. This is the correct format for exact restart, but UW3
+does not yet expose a clean, validated helper for loading mesh variables from
+this format.
+
+The benchmark scripts should not manually guess PETSc section/SF details. The
+fix belongs in UW3 checkpoint I/O.
+
+## PETSc Requirements
+
+PETSc's DMPlex HDF5 restart workflow requires these steps:
+
+1. Load topology with `DMPlexTopologyLoad(...)`.
+2. Keep the `PetscSF` returned by topology load.
+3. Load coordinates with `DMPlexCoordinatesLoad(...)`.
+4. Load labels with `DMPlexLabelsLoad(...)`.
+5. If the mesh is redistributed after load, compose the topology-load SF with
+   the redistribution SF.
+6. Load the saved section with `DMPlexSectionLoad(...)`.
+7. Load the saved vector with `DMPlexGlobalVectorLoad(...)`.
+8. Scatter the global vector into UW3's local mesh-variable storage.
+
+The critical object identity requirements are:
+
+- The topology DM name used during reload must match the saved topology group.
+- The section DM name must match the saved variable/subDM group.
+- The vector name must match the saved vector group.
+- The `PetscSF` passed to `DMPlexSectionLoad(...)` must describe the mapping
+  from the saved topology points to the current distributed topology points.
+
+## Required File Format Changes
+
+### 1. Write DMPlex HDF5 Storage Version 3.0.0
+
+Checkpoint mesh and field files should be written with PETSc DMPlex HDF5
+storage version `3.0.0`.
+
+Expected implementation direction:
+
+- Set the HDF5 viewer / PETSc option used by DMPlex output so the saved files
+  contain `dmplex_storage_version = "3.0.0"`.
+- Ensure this is used by `Mesh.write_checkpoint(...)`.
+- Prefer a local option scope or explicit viewer setting if available, rather
+  than mutating global PETSc options permanently.
+
+Expected outcome:
+
+- New checkpoint files report storage version `3.0.0`.
+- Files remain readable by PETSc `3.25.x`.
+- Existing `write_timestep()` behavior is not unintentionally changed.
+
+### 2. Preserve Or Expose The Correct Topology SF
+
+When loading a mesh from a PETSc DMPlex HDF5 file, UW3 currently calls the
+PETSc topology load path internally. The reload implementation must preserve
+or expose the exact `PetscSF` returned by `DMPlexTopologyLoad(...)`.
+
+Expected implementation direction:
+
+- Verify that `_from_plexh5(..., return_sf=True)` stores the topology-load SF
+  on the UW3 mesh object.
+- Confirm whether any subsequent `DM.distribute()` call changes the topology.
+- If redistribution happens, capture the redistribution SF and compose it with
+  the topology-load SF as PETSc documents.
+
+Expected outcome:
+
+- The mesh object has a reliable SF suitable for `DMPlexSectionLoad(...)`.
+- This SF is not guessed from `mesh.sf`, `mesh.sf0`, or `dm.getDefaultSF()`
+  unless those names are explicitly verified to mean the PETSc-required SF.
+
+### 3. Add A Supported MeshVariable Reload Helper
+
+Add a public helper on mesh variables, for example:
+
+```python
+mesh_var.load_from_checkpoint(
+    filename,
+    data_name=None,
+)
+```
+
+Expected behavior:
+
+- `filename` points to a `write_checkpoint()` checkpoint file.
+- `data_name` defaults to `mesh_var.clean_name`.
+- The helper loads section metadata and vector data exactly.
+- The helper updates both global and local PETSc vectors.
+- The helper works under MPI.
+- The helper does not use KDTree, coordinate remapping, or interpolation.
+
+Expected internal sequence:
+
+```python
+if mesh_var._lvec is None:
+    mesh_var._set_vec(available=True)
+
+indexset, subdm = mesh_var.mesh.dm.createSubDM(mesh_var.field_id)
+sectiondm = subdm.clone()
+
+mesh_var.mesh.dm.setName("uw_mesh")
+subdm.setName(data_name)
+sectiondm.setName(data_name)
+mesh_var._gvec.setName(data_name)
+
+gsf, lsf = mesh_var.mesh.dm.sectionLoad(
+    viewer,
+    sectiondm,
+    mesh_var.mesh.<checkpoint_topology_sf>,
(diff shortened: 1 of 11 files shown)


gthyagi:
## Checkpoint I/O implementation notes

This PR adds a checkpoint/restart-oriented I/O path for mesh variables. The intent is to make checkpoint output behave like a restart/postprocessing format, while keeping `write_timestep()` / `read_timestep()` as the visualisation and coordinate-remap workflow.

### `write_checkpoint()` changes

- Writes PETSc DMPlex checkpoint data using HDF5 storage version `3.0.0`.
- Sets the mesh DM name consistently to `uw_mesh` while writing, so PETSc stores topology and section data under stable group names.
- Writes the mesh checkpoint as:

```text
<base>.mesh.<index>.h5
```

- Supports `outputPath`, matching the calling style of `write_timestep()`.
- Supports two variable-output layouts:

```text
# default: one file per mesh variable
<base>.<variable>.<index>.h5

# optional combined variable file
<base>.checkpoint.<index>.h5
```

- Defaults to `separate_variable_files=True`, because this is easier to handle for large postprocessing jobs where opening one very large combined checkpoint file is undesirable.
- Writes PETSc section/vector metadata for each mesh variable, not visualisation datasets. This means checkpoint files do not replace XDMF/vertex-field output.

### `read_checkpoint()` changes

- Adds `MeshVariable.read_checkpoint(filename, data_name=None)`.
- Loads mesh-variable data through PETSc DMPlex section/vector metadata rather than coordinate/KDTree remapping.
- Uses the checkpoint mesh topology SF saved during mesh load. If the mesh is redistributed by UW, the topology-load SF is composed with the redistribution SF before variable reload.
- Loads the saved section and local vector using a small Cython wrapper around PETSc DMPlex APIs.
- Copies loaded values into the target UW mesh-variable local vector using PETSc section offsets, then syncs local/global vector state.
- `data_name` defaults to the mesh variable clean name and must match the saved variable name when loading from checkpoint files.

### Relationship to `write_timestep()` / `read_timestep()`

The implementation follows the same high-level design pattern as the timestep API: one mesh-level write method and one mesh-variable read method. The difference is the data model and intended use:

| Workflow | Purpose | Reload mechanism | Output extras |
| --- | --- | --- | --- |
| `write_timestep()` / `read_timestep()` | visualisation and flexible remap | coordinate/KDTree remap | XDMF and vertex-field datasets |
| `write_checkpoint()` / `read_checkpoint()` | restart and exact postprocessing | PETSc DMPlex section/vector metadata | no XDMF; no vertex-field datasets |

So this PR does not replace timestep output. It adds a complementary restart-safe path for cases where exact FE-vector reload and lower memory use are more important than visualisation files.

### Technical requirements

- Checkpoint files are written with PETSc DMPlex HDF5 storage version `3.0.0`.
- This requires PETSc support for DMPlex HDF5 storage version 3; the minimum PETSc version should be treated as PETSc `3.19`.
- The current validation was done with PETSc `3.25.x`.
- The mesh file and variable checkpoint files must be used together. They are not intended to be mixed with unrelated meshes.

### Mesh writing and reload behavior

- The checkpoint mesh file stores DMPlex topology, coordinates, labels, and the metadata required by PETSc to reconstruct the distributed mesh.
- During reload, UW stores the `PetscSF` returned by PETSc topology load. This SF is the key mapping between saved topology points and current distributed topology points.
- If UW subsequently distributes the mesh, the topology-load SF is composed with the redistribution SF before variable reload.
- Variable reload relies on that SF plus PETSc section metadata. It does not infer field ownership from coordinates.

### Parallel reload behavior

- Same-rank MPI checkpoint reload is validated for scalar, vector, and discontinuous mesh variables.
- The checkpoint-specific roundtrip test passes under 2 MPI ranks.
- Because checkpoint reload uses PETSc topology/section/SF metadata, it avoids the KDTree memory growth seen in coordinate-remap reloads for large spherical postprocessing jobs.
- Different-rank reload should be considered a follow-up validation item unless reviewers want it included in this PR. The implementation is designed around PETSc SF composition, but the current test coverage is focused on same-rank reload.

### Compatibility and limitations

- Checkpoint files are restart/postprocessing files, not visualisation files. They do not write XDMF or `/vertex_fields/...` datasets.
- Existing `write_timestep()` behavior is intentionally preserved for visualisation and flexible remapping workflows.
- `read_checkpoint()` expects PETSc DMPlex checkpoint files written with matching mesh-variable names and compatible mesh metadata.
- Older PETSc versions without DMPlex HDF5 storage version 3 support are not expected to support this path.

### Expected use cases

- Large benchmark postprocessing where `read_timestep()` KDTree remapping is too memory-heavy.
- Restart-like workflows where fields should be reconstructed from PETSc FE metadata, not coordinate remapping.
- Workflows that still want `write_timestep()` for visualisation, but need `write_checkpoint()` for scalable metric evaluation or restart/postprocessing.

In the motivating spherical benchmark case, the checkpoint reload path preserved velocity/pressure metrics to roundoff and reduced `1/128` postprocessing memory use from about `3.92 TB` to `1.83 TB` on 1152 Gadi CPUs.

gthyagi:
## Note on KDTree-based reload at large scale

The existing `read_timestep()` reload path uses coordinate/KDTree-based remapping. In the spherical benchmark workflow this method is not reliable at large scale; it has been tested up to 1152 CPUs.

The inconsistent behavior was observed using the same `write_timestep()` HDF5 input files for the `1/128` spherical Thieulot case:

```text
output.mesh.00000.h5              5.8G
output.mesh.Velocity.00000.h5     2.4G
output.mesh.Pressure.00000.h5     350M
```

Representative 1152-rank outcomes from the same checkpoint data:

| Job / run | Stage reached | Exit status | Walltime | Memory requested | Memory used | Outcome |
| --- | --- | ---: | ---: | ---: | ---: | --- |
| `166849820` | metrics completed | `0` | `00:13:55` | `4.5 TB` | `3.92 TB` | success |
| `166856023` | loading checkpoint fields | `137` / SIGKILL | `00:09:00` | `4.5 TB` | `3.77 TB` | killed |
| `166862870` | loading `Velocity` | `137` / SIGKILL | `00:10:14` | `4.5 TB` | `3.75 TB` | killed |

So the issue is not just that the KDTree path is expensive. The behavior is inconsistent: with identical HDF5 input files, one run completed inside the allocation, while later runs exceeded memory or were killed during reload.

That non-deterministic memory behavior makes the coordinate/KDTree remap path unsuitable for robust large-scale production postprocessing. The PETSc DMPlex checkpoint reload path in this PR avoids coordinate/KDTree remapping and is intended to provide a more reproducible restart/postprocessing path for large MPI jobs.

gthyagi:
@lmoresi could you please comment on the API naming before this is finalised?

This PR keeps the existing split between:

- `write_timestep()` / `read_timestep()` for timestep-style visualisation / coordinate-remap I/O
- `write_checkpoint()` / `read_checkpoint()` for PETSc DMPlex checkpoint / restart-style I/O

I would appreciate feedback on whether these names are clear enough, or whether better naming would reduce confusion for users. In particular:

- should `write_checkpoint()` / `read_checkpoint()` remain the restart/postprocessing names?
- should `write_timestep()` / `read_timestep()` be documented more explicitly as visualisation/remap-oriented I/O?
- are there better method names or aliases that would make the distinction clearer?

The documentation added in this PR is intended to make the distinction clean and consistent: checkpoint I/O means PETSc metadata-based restart/postprocessing, while timestep I/O means visualisation and coordinate-based remap. Feedback on wording and naming would be useful before this API becomes more widely used.

lmoresi:
@gthyagi — thank you for the careful work; the at-scale evidence is compelling and the implementation is clean. Before this lands, want to raise an architectural concern about the *split* into two writers, and propose a unification path. Tagging #138 because that's where the design discussion started.

## The concern

What this PR ships are **two file formats** that aren't interchangeable:

- A **timestep file** can be visualised in ParaView and remapped onto a different mesh (different resolution, different decomposition, different rank count) via the KDTree path.
- A **checkpoint file** can be exact-restarted on the same mesh at the same rank count, and is much faster/lighter to reload — but cannot be visualised, and cannot be restarted at a different decomposition or different resolution.

The original ask (issue #138) was for a single, restart-grade output that *also* supported the existing flexible-remap and visualisation use cases. Once a checkpoint has been written under the current proposal, it is locked into:

1. The same mesh resolution
2. The same set of mesh variables (within a per-variable file, fine; but you cannot remap that field to a refined mesh later)
3. The same MPI decomposition (different-rank reload is flagged as not-yet-validated in your own doc)

For production benchmark workflows where postprocessing happens once on the same mesh, that's exactly what you want. For research workflows where users want to (a) postprocess a 1/64 run on a 1/32 mesh to compare convergence, or (b) restart at a different rank count when a queue allocation differs, the checkpoint loses information that the timestep format had.

## Concretely: file-content overlap

The two formats live in **disjoint HDF5 namespaces**:

| Payload | timestep | checkpoint |
|---|---|---|
| `/topologies/uw_mesh/dms/<var>/section` (PETSc DMPlex restart metadata) | ❌ | ✅ |
| `/topologies/uw_mesh/dms/<var>/vecs/<var>` (PETSc DMPlex restart vector) | ❌ | ✅ |
| `/fields/<var>` + `/fields/coordinates` (KDTree remap source) | ✅ | ❌ |
| `/vertex_fields/<var>_P<deg>` + `/cell_fields/<var>` (ParaView) | ✅ | ❌ |
| Companion `.xdmf` | ✅ | ❌ |

Nothing in the HDF5 layout prevents both payloads coexisting in the **same file**. PETSc's restart data lives under `/topologies/`, the KDTree/vis data lives under `/fields/` and `/vertex_fields/`, and they don't collide. If a single writer produced all of them:

- `read_checkpoint()` finds the restart payload → exact, fast reload at the same mesh + rank count.
- `read_timestep()` finds `/fields/coordinates` → KDTree remap onto a different mesh / different decomposition.
- ParaView reads the XDMF companion.

## Proposed unification (option B from the offline discussion)

Keep both `write_timestep()` and `write_checkpoint()` names — minimal API churn — but make each writer emit *both* payloads, so the **file** is no longer use-case-locked. The names then describe the **default reader expectation**, not the file content. A user who has only a checkpoint file can still pass it to `read_timestep()` if they later need to remap; a user who has only a timestep file can pass it to `read_checkpoint()` for fast same-mesh reload.

The implementation cost is small: one extra `_write_compat_groups()` call inside `write_checkpoint()`, and one extra `sectionView()` + `localVectorView()` inside `write_timestep()`. The write-time penalty is the projection-to-vertex pass and the section/vector dump — both of which scale with the variable count, not with the rank count, so they're cheap relative to the solve time.

The reader-side memory win you measured (3.92 TB → 1.83 TB at 1152 ranks) is preserved entirely, because the reader still picks the cheap PETSc-native path when the mesh matches.

Two finer points worth checking before unifying:

- **Different-rank reload of the PETSc section/vector path**: your doc flagged this as needing validation. Is it the redistribution `PetscSF` composition that's untested, or the section reconstruction itself? PETSc DMPlex restart at a different rank count *should* work via `DMPlexSectionLoad`'s SF compose — but if it's not validated, the timestep payload is the safety net for cross-decomposition reload.
- **Different-rank reload of the KDTree path**: this works trivially today because each rank queries its own coordinates against the saved coordinates. So even without different-rank PETSc reload, the unified file still supports cross-decomposition work via the KDTree fallback.

## Summary

Concrete proposal:

1. Both writers emit both payloads. Files become use-case-agnostic.
2. Reader names stay distinct and pick the appropriate path.
3. `separate_variable_files=True` stays the default — it's the right default for both schema flexibility (see below) and selective postprocessing.
4. The performance wins of #146 are preserved.

Does this make sense? Happy to draft the additional writer plumbing as a follow-up commit on this branch if you'd like.

gthyagi:
@lmoresi Thanks for the feedback. I am thinking along similar lines, but there are a few bottlenecks I still need to work through:

1. write_checkpoint() currently uses PETSc HDF5 format v3, while write_timestep() uses the older visualization-oriented format.
2. If we try to combine both restart metadata and visualization data, the HDF5 files can become very large.
3. I am seeing memory spikes when write_checkpoint() and write_timestep() are both called in the same script, so I am still testing the safest workflow.

The points you raised are valid. I did a quick pass through the PETSc documentation, and it looks like some of this flexibility may be possible, but I still need to test it properly.

I am currently fixing the last few issues in the spherical benchmarks. Once that is done, I will look more carefully into making the checkpoint output more flexible.

lmoresi (approved):
Approved — discussed at length; CI green; mergeable cleanly. Provides the on-disk DMPlex reload primitive the snapshot toolkit (#195) v1.1 mode will build on.

PR #152, opened 2026-04-28
Fix swarm.write_timestep parallel deadlock (#151)

### LINKED ISSUE #151: [BUG] - swarm.write_timestep fails in parallel (np > 1)
On latest dev branch. Script to recreate:
```

# %%
import underworld3 as uw
from underworld3.systems import Stokes
import numpy as np
import sympy
import os
from petsc4py import PETSc



# %%
res = 0.2
mesh = uw.meshing.UnstructuredSimplexBox(
    minCoords=(-1.0, 0.0),
    maxCoords=(1.0, 1.0),
    cellSize= res,
    regular=False,
    qdegree=3,
)

# %%
swarm = uw.swarm.Swarm(mesh=mesh)
material      = swarm.add_variable(name="material", size=1, dtype=PETSc.IntType)
swarm.populate(fill_param=4)


swarm.write_timestep(filename='swarm', swarmname='swarm', index=0, outputPath='./', swarmVars=[material]) 
```

lmoresi:
## Summary
- Fixes the silent parallel-HDF5 collective-close hang in \`Swarm.save()\` and \`SwarmVariable.save()\`.
- Two latent bugs in the sequential fallback paths cleaned up along the way.
- Adds \`tests/parallel/test_0790_swarm_write_timestep_mpi.py\` — bknight1's reproducer wrapped as a level_2 MPI regression with a 60s pytest timeout.

Fixes #151.

## Root cause

Both \`Swarm.save()\` and \`SwarmVariable.save()\` parallel paths called

```python
h5f.create_dataset(name, data=local_data)
```

collectively, but each rank passed its own local-sized array. In parallel HDF5 every rank must specify the **same dataset shape** on a collective \`create_dataset\`; passing different shapes leaves HDF5's internal metadata inconsistent so the collective close never synchronises — producing a silent hang. Faulthandler stack trace at the hang point confirmed both ranks stuck inside \`h5py.File.__exit__()\` → \`close()\`.

The bug only manifests when ranks have **unequal local particle counts** — which is the common case for any non-trivial swarm. A perfectly balanced run (equal local sizes by accident) would coincidentally specify the same shape on every rank and complete normally; that's likely why this went unnoticed.

## Fix

Allgather the per-rank local sizes, create the dataset at the global \`(total_n, n_components)\` shape, then have each rank write its own slice via \`dset[offset:offset+local_n] = local_data\`. Standard parallel HDF5 collective pattern.

## Latent bugs fixed along the way

- **\`SwarmVariable.save\` sequential fallback indentation.** The \`comm.barrier()\` and the rank>0 append loop were nested inside \`if comm.rank == 0:\`. Higher-rank slabs were never written, and rank 0 hit the lone barrier with no peers, deadlocking the sequential path too. Moved barrier and loop outside the rank-0 branch.

- **\`Swarm.save\` sequential append branch \`NameError\`.** The per-rank append branch read \`data_copy.shape[0]\`, but the in-scope variable is \`points_data_copy\`. Any zero-particle rank would have raised NameError. Replaced with the correct local variable.

## Test plan

- [x] bknight1's reproducer at np=2: **pre-fix hangs forever (15s faulthandler dump confirms close()), post-fix completes in 0.11s**.
- [x] Same reproducer at np=4: completes in 0.12s.
- [x] New regression test passes at np=2 in 2.2s.
- [x] \`pytest -m \"level_1 and tier_a\"\` on \`amr-dev\`: **56 passed, 3 skipped, 0 failed**. No regressions.

### DIFF
--- a/src/underworld3/swarm.py
+++ b/src/underworld3/swarm.py
@@ -1802,39 +1802,74 @@ class SwarmVariable(DimensionalityMixin, MathematicalMixin, Stateful, uw_object)
         if filename.endswith(".h5") == False:
             raise RuntimeError("The filename must end with .h5")
 
+        local_data = self.data[:]
+        local_n = local_data.shape[0]
+        n_components = self.num_components
+
         if h5py.h5.get_config().mpi == True and not force_sequential:
+            # BUGFIX(#151): the previous parallel path called
+            #   h5f.create_dataset("data", data=self.data[:])
+            # collectively, but each rank passed its own local-sized array.
+            # In parallel HDF5 every rank must specify the *same* dataset
+            # shape on a collective create_dataset; passing different shapes
+            # leaves HDF5's internal metadata inconsistent so the collective
+            # close never synchronises, producing a silent hang.
+            #
+            # Fix: allgather the per-rank sizes, create the dataset at the
+            # global shape, then each rank writes its own slice.
+            sizes = comm.allgather(local_n)
+            total_n = sum(sizes)
+            offset = sum(sizes[: comm.rank])
+
             with h5py.File(f"{filename[:-3]}.h5", "w", driver="mpio", comm=comm) as h5f:
                 if compression == True:
-                    h5f.create_dataset("data", data=self.data[:], compression=compressionType)
+                    dset = h5f.create_dataset(
+                        "data",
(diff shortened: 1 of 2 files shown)


bknight1:
Tested & working

Issue #175, opened 2026-05-07
[BUG] - swarm particles being lost across processors resulting in 'gaps'


bknight1:
I am running a model with swarm particles but large gaps are appearing within the domain, which correlates to processor boundaries. I haven't got a basic script to replicate but can make one if required.

bknight1:
I should add this is when the particles are being advected around the domain

bknight1:
Reproduction script for np > 2:


```
import underworld3 as uw
import numpy as np
from mpi4py import MPI

mesh = uw.meshing.UnstructuredSimplexBox(
    minCoords=(0.0, 0.0),
    maxCoords=(1.0, 1.0),
    cellSize=0.1,
)

swarm = uw.swarm.Swarm(mesh)

# Initialize a variable to track particles
var = uw.swarm.SwarmVariable("test_var", swarm, 1)

swarm.populate(fill_param=1)
var.data[...] = uw.mpi.rank

# Initial count
comm = uw.mpi.comm
initial_count = comm.allreduce(swarm.dm.getLocalSize(), op=MPI.SUM)
if uw.mpi.rank == 0:
    print(f"Initial total particles: {initial_count}")

# Define a constant velocity field to move particles from left to right
# V = (1.0, 0.0)
import sympy
v_fn = sympy.Matrix([1.0, 0.0])

# Advect for a bit
dt = 0.6 # move 0.4 units, should cross x=0.5 if they start at x < 0.5
swarm.advection(v_fn, dt, order=1)

# Final count
final_count = comm.allreduce(swarm.dm.getLocalSize(), op=MPI.SUM)
if uw.mpi.rank == 0:
    print(f"Final total particles: {final_count}")

if uw.mpi.rank == 0:
    if final_count < initial_count:
        print(f"FAILED: Lost {initial_count - final_count} particles!")
    else:
        print("PASSED: No particles lost.")
```
Looks like the `advection` function is forced to use [max_steps=1](https://github.com/underworldcode/underworld3/blob/93d02801ef342adb0367628de71e7daf2aa08430/src/underworld3/swarm.py#L4543), causing issues. Seems a simple fix but will investigate further before finalising

lmoresi:
Definitely a regression- seen this before !


 --- 8< ------------------------------



Prof Louis Moresi

Computational Geophysics

Research School of Earth Sciences

The Australian National University



+61 2 6125 5590

***@***.***



[...]

PR #177, opened 2026-05-10
Fix swarm particle loss across rank boundaries during advection (#175)

lmoresi:
## Summary
- Drops the `max_its=1` override on `Swarm.advection()`'s end-of-step `self.migrate(...)` call. With only one kdtree-centroid try, particles whose owning rank was the 2nd-or-3rd closest failed `points_in_domain` and — because `delete_lost_points=True` — got silently deleted. Reverting to the default `max_its=10` restores the boundary-claim retry that the kdtree migration is designed for.
- Adds `tests/parallel/test_0765_swarm_advection_no_loss.py` (adapted from @bknight1's reproducer): two cases that move particles across processor cuts and assert global particle count is conserved.

The override looks like leftover debugging from `bd209c4` that wasn't dialled back. Existing swarm tests (`test_0760_swarm_cache_migration`) continue to pass.

## Test plan
- [x] `mpirun -n 4 python -m pytest --with-mpi tests/parallel/test_0765_swarm_advection_no_loss.py` — passes
- [x] Same test fails without the fix (order=1: 30/726 lost; order=2 rotation: 66/726 lost on 4 ranks)
- [x] `mpirun -n 4 python -m pytest --with-mpi tests/parallel/test_0760_swarm_cache_migration.py` — passes
- [x] @bknight1 to confirm fix on his original HPC workload at higher proc counts

Reported and diagnosed by @bknight1 — credit and thanks.

### DIFF
--- a/src/underworld3/swarm.py
+++ b/src/underworld3/swarm.py
@@ -4537,10 +4537,14 @@ class Swarm(Stateful, uw_object):
             #
             #
 
-        # Remove points no longer in the domain
+        # Re-route particles to their owning ranks and remove any that
+        # have genuinely left the domain. Use the default max_its so that
+        # boundary particles whose owner is the 2nd/3rd-closest centroid
+        # get reclaimed via the kdtree retry — max_its=1 here was an
+        # accidental regression that deleted boundary particles (issue #175,
+        # reported by @bknight1).
         self.migrate(
             delete_lost_points=True,
-            max_its=1,
         )
 
         return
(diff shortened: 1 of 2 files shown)


bknight1:
All good, confirmed working @lmoresi

bknight1:
This line (https://github.com/underworldcode/underworld3/blob/ddd1e5cfe4fd8074f1ff6befa50946c84ee93fe4/src/underworld3/swarm.py#L3369) returns an index error

Issue #176, opened 2026-05-09
[BUG] - memory leak


bknight1:
I'm running out of memory when running on the HPC:

```
s:qurmstepd: error: Detected 1 oom_kill event in StepId=42492733.0. Some of the step tasks have been OOM Killed.
srun: error: nid001653: task 16: Out Of Memory
srun: Terminating StepId=42492733.0
slurmstepd: error: *** STEP 42492733.0 ON nid001653 CANCELLED AT 2026-05-08T22:47:45 ***
```

I'm running `#SBATCH --mem-per-cpu=1200M` which is close to the max you can request per cpu.

lmoresi:
We are going to need a bit more help on the setup

How many cpu's
Tinestepping ?
Particles or not ?

L


 --- 8< ------------------------------



Prof Louis Moresi

Computational Geophysics

Research School of Earth Sciences

The Australian National University



+61 2 6125 5590

***@***.***



TEQSA Provider ID: PRV12002 (Australian University) | CRICOS Provider Code: 00120C



The Australian National University acknowledges, celebrates and pays our respects to the Ngunnawal and Ngambri people of the Canberra region and to all First Nations Australians on whose traditional lands we meet and work, and whose cultures are among the oldest continuing cultures in human history.

________________________________
From: Ben Knight ***@***.***>
Sent: Saturday, May 9, 2026 1:08:31 PM
To: underworldcode/underworld3 ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [underworldcode/underworld3] [BUG] - memory leak (Issue #176)

[https://avatars.githubusercontent.com/u/55677727?s=20&v=4]bknight1 created an issue (underworldcode/underworld3#176)<https://github.com/underworldcode/underworld3/issues/176>

I'm running out of memory when running on the HPC:

s:qurmstepd: error: Detected 1 oom_kill event in StepId=42492733.0. Some of the step tasks have been OOM Killed.
srun: error: nid001653: task 16: Out Of Memory
srun: Terminating StepId=42492733.0
slurmstepd: error: *** STEP 42492733.0 ON nid001653 CANCELLED AT 2026-05-08T22:47:45 ***


I'm running #SBATCH --mem-per-cpu=1200M which is close to the max you can request per cpu.

bknight1:
[...]

PR #178, opened 2026-05-10
Fix comprehensive memory leaks in solver setup, interpolation caching, and SubDM syncing

bknight1:
Resolves https://github.com/underworldcode/underworld3/issues/176

### Summary of Changes
This PR comprehensively addresses severe memory leaks identified in the solver loop, particularly during dynamic coordinate updates (e.g., swarm advection) and tight solve iterations. 

1. **DMInterpolationCache LRU Policy:** Implemented a Least-Recently-Used (LRU) eviction policy in the `DMInterpolationCache` with a `max_entries` cap (default: 10). This prevents unbounded memory growth when evaluation coordinates change every step.
2. **`SubDM` Cleanup in `MeshVariable`:** Added explicit `subdm.destroy()` calls in `_sync_lvec_to_gvec` and other `MeshVariable` synchronization paths. Previously, temporary `SubDM` objects created during data syncs were orphaned.
3. **Solver Property Guards:** Modified `SymbolicProperty` and solver setters to use `self._needs_function_rewire = True` instead of `self.is_setup = False` when only pointwise functions change. This avoids catastrophic, redundant teardowns and rebuilds of the PETSc `DM` hierarchy.
4. **Stokes & Solver Base Cleanup:** Implemented comprehensive explicit destruction (`.destroy()`) of old PETSc FE objects, `DM` hierarchies, and `SNES` objects within the `_setup_discretisation` and `_build` paths before allocating new ones.

### Verification and Test Findings
A new robust memory leak regression test (`tests/test_0006_memory_leak.py`) was added. The test isolates steady-state leakage by warming up for 30 steps and measuring absolute memory growth over the subsequent 70 steps on a 32x32 mesh.

*   **Before PR (`development` branch):** Memory grew by **77.83 MiB** (~1.1 MiB leaked per step) over the final 70 steps, failing the test. 
*   **After PR (`fix/memory-leak-comprehensive` branch):** Memory growth dropped to just **9.94 MiB** (~0.14 MiB per step) over the same window. This remaining footprint represents high-water mark expansion rather than a leak. The test passes securely.

All standard regression and parallel tests continue to pass.

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -56,6 +56,17 @@ class SolverBaseClass(uw_object):
         self.petsc_options_prefix = self.name
         self.petsc_options = PETSc.Options(self.petsc_options_prefix)
 
+        self.dm = None
+        self.dm_hierarchy = []
+        self.snes = None
+        self.petsc_fe_u = None
+        self.petsc_fe_p = None
+
+        # Solution decomposition caches (IS sets)
+        self._velocity_is = None
+        self._pressure_is = None
+        self._subdict = {}
+
     def _check_expression_meshes(self):
         """Check that all MeshVariable symbols in solver expressions
         belong to this solver's mesh.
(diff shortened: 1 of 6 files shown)


lmoresi (approved):
Looks good, Ben,

I added some diagnostics at the same time you figured out the fix. They should be complimentary and help find the issues quickly next time.

bknight1:
Sounds good, thanks @lmoresi

PR #181, opened 2026-05-11
Core memory leak fixes: Cython deallocation & callback hardening

bknight1:
This PR addresses several memory-related issues and stability concerns identified during OOM investigations on HPC clusters.

### Key changes:
- **Cython Memory Management**: Implemented `__dealloc__` in the `PtrContainer` class to ensure that `malloc`'d function pointer arrays are properly freed when Python objects are garbage collected. Added re-allocation guards to prevent leaks during container reuse.
- **Array Callback Hardening**: Updated mesh and variable array callbacks to verify `array.owner` exists before execution. This prevents `AttributeError` during complex object teardown (e.g., at application exit or during mesh replacement).
- **Recursion Guards**: Integrated explicit recursion guards and `mesh_update_lock` checks in `MeshVariable` callbacks to prevent PETSc synchronization during sensitive coordinate deformations.

### DIFF
--- a/docs/developer/guides/memory-diagnostics.md
+++ b/docs/developer/guides/memory-diagnostics.md
@@ -27,14 +27,22 @@ growing on every step.
 
 | Signal | Source | Cost |
 |---|---|---|
-| Process RSS (MiB) | `resource.getrusage` | free |
+| Process RSS (MiB, current) | `psutil.Process().memory_info().rss`, fallback `/proc/self/statm` (Linux), last resort `resource.ru_maxrss` | free |
 | KDTree live count | `uw.kdtree.live_count()` | free |
 | KDTree total constructed | `uw.kdtree.total_constructed()` | free |
 | Per-class Python instance counts | `gc.get_objects()` walk | slow — gated behind `full=True` |
 
+The RSS source is **current** RSS where possible — it should drop when memory
+is freed, not just rise. The last-resort `resource.ru_maxrss` fallback returns
+the **peak** (high-water-mark) RSS instead and never decreases, so on systems
+where neither psutil nor `/proc/self/statm` is available you'll see growth but
+not recovery; install `psutil` to fix that.
+
 `KDTree` instances are tracked via Cython class counters in `__cinit__` and
-`__dealloc__`. Cython's deterministic destruction makes the live count
-accurate without weak-references.
+`__dealloc__`. CPython refcounting calls `__dealloc__` promptly when the
+refcount hits zero, so the count is accurate for typical use; it can lag if a
+KDTree ends up in a reference cycle that only the cyclic garbage collector
+can break. Call `gc.collect()` before reading if that matters.
 
 PETSc-side object and allocation tracking is **not** parsed from Python —
 PETSc's own `-log_view` and `-malloc_dump` runtime flags give the same
--- a/docs/developer/guides/memory-diagnostics.md
+++ b/docs/developer/guides/memory-diagnostics.md
@@ -1,0 +1,145 @@
+# Memory diagnostics
+
+Long parallel runs occasionally OOM on HPC even when each step looks small.
+The `uw.utilities.memprobe` module gives you a way to "light up" memory
+tracking on demand, sample at regular intervals, and pin which subsystem is
+growing.
+
+## Quick start
+
+```bash
+UW_MEMPROBE=1 mpirun -n 16 python my_long_run.py
+```
+
+With this flag set, the `Stokes.solve()` and `NavierStokes.solve()` paths
+emit a one-line growth report each time they're called:
+
+```
+[memprobe] Stokes.solve:
+  RSS +0.42 MiB
+  kdtree: live +1, total_constructed +1
+```
+
+A clean run shows mostly zero deltas. A leak shows the same component
+growing on every step.
+
+## What's tracked
+
+| Signal | Source | Cost |
+|---|---|---|
+| Process RSS (MiB) | `resource.getrusage` | free |
+| KDTree live count | `uw.kdtree.live_count()` | free |
+| KDTree total constructed | `uw.kdtree.total_constructed()` | free |
+| Per-class Python instance counts | `gc.get_objects()` walk | slow — gated behind `full=True` |
+
+`KDTree` instances are tracked via Cython class counters in `__cinit__` and
+`__dealloc__`. Cython's deterministic destruction makes the live count
+accurate without weak-references.
+
+PETSc-side object and allocation tracking is **not** parsed from Python —
+PETSc's own `-log_view` and `-malloc_dump` runtime flags give the same
+information more reliably. To enable them from Python:
+
+```python
+from underworld3.utilities import memprobe
+memprobe.dump_petsc_leaks_at_finalize()  # equivalent to -malloc_dump -objects_dump
+```
+
+## API
+
+### Snapshots and diffs
+
+```python
+from underworld3.utilities import memprobe
+
+before = memprobe.snapshot()
+do_work()
+after = memprobe.snapshot()
+print(memprobe.format_diff("after-work", memprobe.diff(before, after)))
+```
+
+Add `full=True` to also walk Python-class counts:
+
+```python
+snap = memprobe.snapshot(full=True)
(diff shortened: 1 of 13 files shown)


lmoresi:
Thanks @bknight1 — at a glance this looks orthogonal to the other in-flight work (#179 touched `ckdtree.pyx` + `solvers.py`; this PR is in `petsc_types.pyx` + `discretisation_mesh*.py` + a small `swarm.py` change). No conflicts expected.

### To check before merge

1. **`PtrContainer.__dealloc__` re-allocation guard** — confirm the guard handles the case where a `PtrContainer` is reused with a smaller array (free old before allocating new, no double-free).

2. **`array.owner` existence check in callbacks** — please add a one-line test or comment indicating which teardown path triggers the missing-owner case. That's the kind of guard that's invisible until something breaks, and a test or pointer-comment lets future-us understand why it's there.

3. **`mesh_update_lock`** — is this a new attribute introduced in this PR, or pre-existing? If new, a one-line docstring on it (where set, where cleared, what operations honour it) would help. Recursion guards are easy to leave half-wired.

4. **Cross-validation with #179**: once merged, a `UW_MEMPROBE=1` smoke run should show a clean drop in steady-state RSS after this PR vs. before — that's a nice fingerprint for the `PtrContainer` leak specifically.

Otherwise this looks ready. CI passing + the above confirmations and I'd say merge.

lmoresi:
@bknight1 — CI on this PR is failing, two things to address:

### 1. Rebase onto current `development` (picks up a race fix you don't have)

This PR's base is `848b8b9` (the #178 merge commit). Yesterday after that landed, three more PRs were merged:
- #177 — `Swarm.advection` `max_its=10` fix (your #182 also has this; will conflict).
- #179 — `memprobe` diagnostic module + KDTree counters.
- A 1-line race fix `5f64de2` for `UWexpression._ephemeral_expr_names` iteration mid-finalize (the "dictionary changed size during iteration" race).

Your CI shows the race fix matters here:
```
FAILED tests/test_0810_bc_unit_conversion.py::test_bc_accepts_raw_numbers
  - RuntimeError: dictionary changed size during iteration
```

Rebase:
```bash
git fetch origin
git rebase origin/development
# resolve any conflicts (likely small — your changes are in different files)
git push --force-with-lease
```

### 2. Investigate the integral / swarm-statistics regression

CI also reports:
```
FAILED tests/test_0501_integrals.py::test_integrate_meshvar           - assert 0.6366197... < 0.001
FAILED tests/test_0501_integrals.py::test_integrate_swarmvar_deriv_O1 - assert 2.0 < 0.001
FAILED tests/test_0501_integrals.py::test_integrate_swarmvar_deriv_03 - assert 2.0 < 0.001
FAILED tests/test_0502_boundary_integrals.py::test_bd_integral_meshvar - Expected 0.6366..., got 0.0
FAILED tests/test_0852_swarm_integration_statistics.py::...           - assert 0.0 > 0
FAILED tests/test_0852_swarm_integration_statistics.py::...           - assert 273.0 < 0.0
```

These look like *real regressions caused by this PR*, not pre-existing flakes:
- `test_bd_integral_meshvar` returning `0.0` instead of `0.6366...` is severe — a boundary integral computing zero is the kind of thing you only see if data isn't reaching the integration context at all.
- `test_integrate_meshvar` returning `0.6366` where the test expects `< 0.001` (an error-norm assertion) is the same shape — the *raw value* leaking through where the *residual* should be.

My guess: the `array.owner` existence guard in your `MeshVariable` callbacks is silently early-returning when it shouldn't (e.g. during a teardown-and-rebuild of an integration work-variable, where `owner` is briefly absent but the data sync is still required). Or the `mesh_update_lock` is gating a path that integration relies on.

Worth running these four tests locally with `UW_MEMPROBE=1` and a print/breakpoint in the new callback guard to see which path is being short-circuited.

After the rebase + fix, push and CI should pass — at which point we can merge.

bknight1:
I have addressed the CI failures and the feedback regarding the mesh variable synchronization regression.

### Fixes and Verification:

1.  **Resolved MeshVariable Synchronization Regression**:
    *   **Root Cause**: In the initial PR commit, `MeshVariable._create_canonical_data_array` was missing the `owner=self` argument in the `NDArray_With_Callback` constructor. This caused the PETSc synchronization callback to early-return because `array.owner` was `None`, preventing updates to the underlying PETSc data from the Python `.data` property.
    *   **Impact**: Integration and boundary integral tests were returning `0.0` or residuals that didn't match expectations because the solver was operating on uninitialized PETSc vectors.
    *   **Fix**: Added `owner=self` to ensure the callback has access to the variable for `pack_raw_data_to_petsc` calls.
    *   **Verification**: Verified that `test_0501_integrals.py`, `test_0502_boundary_integrals.py`, and `test_0852_swarm_integration_statistics.py` now pass.

2.  **Resolved Parallel Swarm Advection (Issue #175)**:
    *   **Root Cause**: A regression in `Swarm.advection` (likely from a previous merge or manual override) set `max_its=1` in the final `migrate` call. This prevented the KD-tree from retrying neighboring ranks for particles at process boundaries, leading to silent deletion.
    *   **Fix**: Restored the default `max_its=10` to enable the retry path.
    *   **Verification**: Verified with `tests/parallel/test_0765_swarm_advection_no_loss.py` on 2 and 4 MPI ranks.

3.  **Documentation and Safety Hardening**:
    *   Updated comments in `MeshVariable` and `SwarmVariable` callbacks to clarify that the `owner is None` check is a safety guard for Python garbage collection race conditions during object teardown (e.g., at exit or during mesh replacement).
    *   Documented `_mesh_update_lock` in `Mesh` to clarify its role in preventing PETSc synchronization conflicts during coordinate deformations.

### Rebase:
I have rebased the branch onto the latest `development` to pick up recent race condition fixes and resolve conflicts with the overlapping `advection` fix.

All reported regressions are now resolved and verified locally.

lmoresi:

@bknight1 — the four pre-merge check items from 2026-05-11 are all addressed, and the regression root-cause + fix is well-explained. Going through the verification:

### Check items resolved
1. **`PtrContainer.__dealloc__` reuse guard** — `petsc_types.pyx` now has paired `__cinit__`/`__dealloc__` plus an explicit free-before-reallocate in `allocate()`. Clean.
2. **`array.owner is None` guard** — purpose now documented inline as "Python GC race during object teardown (exit / mesh replacement)". The reason the integral tests failed in the first CI run is the exact teardown path that's now guarded — perfect demonstration of why the comment matters.
3. **`_mesh_update_lock` docstring** — added; explains it's held by `mesh_update_callback` during `_deform_mesh()` and checked by the variable-side callbacks via `var.mesh._mesh_update_lock.acquire(blocking=False)`. Reentrant `RLock` is the right primitive.
4. **`UW_MEMPROBE=1` smoke + CI green** — passing per checks.

### Bug found, root cause documented, fix verified
The `owner=self` regression on `_create_canonical_data_array` is exactly the kind of failure mode the `array.owner` guard is meant to catch — when the callback is plumbed correctly the guard is invisible; the moment you forget to pass `owner`, the guard silently turns off all sync and integrals start returning zero. That's a *useful* failure mode (loud test failures rather than quiet data corruption later) and worth keeping a note about — perhaps a one-line assertion in `_create_canonical_data_array` that `array_obj.owner is self` post-construction, so any future caller that forgets fails immediately rather than via an integral regression three layers up.

### Small leftover from Copilot's review
Copilot flagged unconditional `print()` calls in `mesh_update_callback` (no line pinned). Worth checking whether any `print(...)` statements in the callback path remain unguarded — these fire on every coordinate write and run on all MPI ranks. If any survived the diff, gate behind `uw.pprint(0, ...)` or a verbose flag before merging. Not blocking if they were already removed or gated.

### Cross-validation with #182
Once both #181 and #182 land, the leak-free fingerprint on a tight solver loop should be:
- Flat per-step RSS (this PR)
- Zero `kdtree: total_constructed` delta after warm-up (#182)

Worth a quick `UW_MEMPROBE=1` smoke on the Gadi-class reproducer from #176 to confirm before declaring that issue closed.

LGTM modulo the `print()` gating check.

lmoresi (approved):
Approving per review pass — all four pre-merge check items addressed, regression root-cause clearly explained, CI green. Closes #175 and provides the Cython/callback foundation for #176.

Issue #180, opened 2026-05-11
[BUG] - Matrix Mismatch Regarding NavierStokesSLCN Solver


Chawgai:
I am facing the following problem while using the attached test script. I run like mpirun -n 4 python nstt.py

[nstt.txt](https://github.com/user-attachments/files/27578723/nstt.txt)

**Screenshots**
  File "/Users/u8085704/underworld3/.pixi/envs/dev/lib/python3.12/site-packages/underworld3/systems/solvers.py", line 2414, in <la
    navier_stokes.solve(timestep=delta_t, zero_init_guess=False)
  File "/Users/u8085704/underworld3/.pixi/envs/dev/lib/python3.12/site-packages/underworld3/timing.py", line 326, in timed
    result = routine(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/u8085704/underworld3/.pixi/envs/dev/lib/python3.12/site-packages/underworld3/systems/solvers.py", line 3773, in solve
    self.DFDt.update_pre_solve(timestep, verbose=verbose, evalf=_evalf)
  File "/Users/u8085704/underworld3/.pixi/envs/dev/lib/python3.12/site-packages/underworld3/systems/ddt.py", line 1771, in update_pre_solve
    self._psi_star_projection_solver.solve(verbose=verbose)
  File "/Users/u8085704/underworld3/.pixi/envs/dev/lib/python3.12/site-packages/underworld3/timing.py", line 326, in timed
    result = routine(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/underworld3/cython/petsc_generic_snes_solvers.pyx", line 3863, in underworld3.cython.generic_solvers.SNES_MultiComponent.solve
  File "/Users/umbda>
    lambda self: (self.u.sym - self.uw_function) * self.uw_weighting_function,
                  ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
  File "/Users/u8085704/underworld3/.pixi/envs/dev/lib/python3.12/site-packages/sympy/core/decorators.py", line 118, in binary_op_wrapper
    return func(self, other)
           ^^^^^^^^^^^^^^^^^
  File "/Users/u8085704/underworld3/.pixi/envs/dev/lib/python3.12/site-packages/sympy/matrices/matrixbase.py", line 3045, in __sub__
    return self + (-a)
           ~~~~~^~~~~~
[...]

PR #183, opened 2026-05-11
Fix NavierStokesSLCN DDt projection source shape mismatch (#180)

lmoresi:
## Summary

One-line outdated-API fix. When `SNES_MultiComponent_Projection` was wired into the `SemiLagrangian` DDt path, the `psi_fn` setter and `_setup_projections` were updated to flatten the source tensor to `(1, Nc)` via `_build_projection_source` — but the fallback inside `update_pre_solve` (taken whenever `uw.function.evaluate()` raises on expressions containing derivatives) was missed and kept assigning `self.psi_fn` directly.

Stokes never hits this fallback because its projection source has no derivatives. NavierStokes hits it every solve, producing:

```
sympy.matrices.exceptions.ShapeError: Matrix size mismatch: (1, 3) + (2, 2).
```

— `(1, 3)` is the projector's `Nc=3` symmetric-tensor solution variable in 2D; `(2, 2)` is the raw stress flux tensor that should have been flattened.

### Files
- `src/underworld3/systems/ddt.py` — route the fallback through `_build_projection_source`.
- `tests/test_0610_navier_stokes_slcn_projection.py` — regression test: one `ns.solve()` on a tiny mesh. Reliably raises `ShapeError` without the fix, passes with it.

## Test plan
- [x] `pytest tests/test_0610_navier_stokes_slcn_projection.py` — passes
- [x] Without the fix the same test fails with the user's exact error `(1, 3) + (2, 2)`
- [ ] Reporter @Chawgai to confirm on their `nstt.py` reproducer

Reported by @Chawgai. Diagnosed with @ss2098's investigation pointing at the SLCN projection/history-field path as the root cause.

### DIFF
--- a/src/underworld3/systems/ddt.py
+++ b/src/underworld3/systems/ddt.py
@@ -1765,8 +1765,14 @@ class SemiLagrangian(uw_object):
 
             except Exception:
                 # Fallback to projection solver for expressions that can't be directly evaluated
-                # (e.g., containing derivatives)
-                self._psi_star_projection_solver.uw_function = self.psi_fn
+                # (e.g., containing derivatives — true for the NS viscous flux every step).
+                # Route via _build_projection_source so the (1, Nc) row-matrix flattening
+                # required by SNES_MultiComponent_Projection is applied for tensor vtypes.
+                # Without this, a (dim, dim) tensor function meets a (1, Nc) solver field
+                # and SymPy raises "Matrix size mismatch: (1, Nc) + (dim, dim)" (issue #180).
+                self._psi_star_projection_solver.uw_function = self._build_projection_source(
+                    self.psi_fn
+                )
                 self._psi_star_projection_solver.smoothing = 0.0
                 self._psi_star_projection_solver.solve(verbose=verbose)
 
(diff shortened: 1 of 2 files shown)


lmoresi:

**Scope:** the DDt-fallback path was assigning `self.psi_fn` directly to `_psi_star_projection_solver.uw_function`, which collided with `SNES_MultiComponent_Projection`'s `(1, Nc)` row-matrix contract — producing the `ShapeError("(1, 3) + (2, 2)")` reported in #180. The fix routes through `_build_projection_source(self.psi_fn)` so the same flattening used at setup time applies to the runtime fallback, and adds the matching flat→tensor fan-out so subsequent advection sees the freshly-projected values.

**Verification:**
- Confirmed `_build_projection_source` already exists and is what `_setup_projections` and the `psi_fn` setter use — the fallback was the only place that had been missed in the multicomponent migration.
- Fan-out mirrors `initialise_history()` (~line 1540) — same indexing, same symmetric fill.
- The new regression test (`tests/test_0610_navier_stokes_slcn_projection.py`) reproduces the exact `ShapeError("(1, 3) + (2, 2)")` from the issue, and is correctly tagged `level_2 / tier_a` so it runs in standard CI.
- @Chawgai already reported on #180 that this branch resolves the crash on his `nstt.py` reproducer.

**Copilot note on stale state:** Copilot's earlier comment about the missing fan-out is now stale — the current diff includes exactly the fan-out it asked for (lines 1779–1785). Worth marking resolved.

**Suggested follow-up:** the fallback exists at all because `uw.function.evaluate` couldn't handle tensor expressions with derivatives. PR #185 removes that limitation, which means in steady-state operation this fallback path should become rare. Both PRs together close out #180.

LGTM modulo standard maintainer eyeball.

PR #185, opened 2026-05-13
Implement tensor evaluate path in _project_to_work_variable (#180 followup)

lmoresi:
## Summary

Replaces dead matrix-projection code in ``_project_to_work_variable`` (``_function.pyx``) so ``uw.function.evaluate`` works directly on tensor expressions containing derivatives (e.g. strain rate, NS viscous flux). Follow-on to #183 — together they fix issue #180 from both sides:

- **#183** makes the DDt fallback path do the right shape when it's entered.
- **#179** (this PR) prevents the fallback from being entered in the first place for tensor-evaluate cases.

## The bugs (both marked by a pre-existing ``TODO(BUG)``)

1. ``MeshVariable(num_components=rows*cols)`` (flat int) — base class can't infer vtype, raises ``ValueError`` on every call. This is what Mohammed sees in his trace.
2. ``SNES_Projection(scalar_component=c)`` — that keyword doesn't exist on ``SNES_Projection.__init__``; the loop would have ``TypeError`` if it ever reached.

So the entire matrix branch was unreachable. Every tensor-evaluate call (e.g. ``evaluate(NS.flux, coords)``) bounced out via bug 1 and callers like ``SemiLagrangian.update_pre_solve`` fell back to their own projector.

## Fix

Replaces the dead branch with a real multi-component projection that mirrors the pattern already used in ``DDt._setup_projections``:
- Cache a flat ``(1, Nc) MATRIX`` work var as the projector target.
- Use ``SNES_MultiComponent_Projection`` (one solve, shared DM) instead of per-component cycling.
- Cache a tensor-shaped ``(rows, cols)`` work var so the caller's ``work_var.sym`` keeps the original shape.
- After solve, fan flat var values into the tensor work var.

## Tests

``tests/test_0506_tensor_evaluate.py`` (3 tests):
- ``evaluate(strain_tensor, pts)`` returns shape ``(n_pts, dim, dim)`` — rotation field gives zero strain (correct).
- Simple shear field produces the expected off-diagonal strain entries ~0.5.
- A NavierStokes solve completes without entering the DDt fallback.

## Test plan
- [x] ``pytest tests/test_0506_tensor_evaluate.py`` — 3 passed
- [x] ``pytest tests/test_0503_evaluate.py tests/test_0503_evaluate2.py tests/test_0504_projections.py`` — 18 passed (no regression on existing evaluate paths)
- [x] Instrumented trace confirms ``_project_to_work_variable`` is now reached and succeeds with ``(2, 2)`` shaped tensors on a NS solve
- [ ] Reviewer to confirm the cache scheme (``mesh._eval_work_<rows>x<cols>`` etc.) is OK for long-running models

### DIFF
--- a/src/underworld3/function/_function.pyx
+++ b/src/underworld3/function/_function.pyx
@@ -549,44 +549,73 @@ def _project_to_work_variable(expr, mesh, smoothing=1e-6):
         Work variable containing projected nodal values
     """
     import underworld3 as uw
+    import sympy
 
-    # Handle matrix expressions - need multi-component work variable
-    # TODO(BUG): This fails for dim×dim matrices (e.g. 2×2 stress tensor)
-    # because MeshVariable can't infer vtype from num_components=4 (flat int).
-    # Needs: pass num_components=(rows,cols) with vtype=TENSOR, and use
-    # Tensor_Projection instead of per-component scalar Projection.
-    # Currently, callers like SemiLagrangian.update_pre_solve fall back to
-    # their own projection solver via except-clause when this fails.
+    # Handle matrix/tensor expressions — need a multi-component projection.
+    # Implementation: project the (rows, cols) expression into a flat
+    # (1, Nc) MATRIX work variable using SNES_MultiComponent_Projection
+    # (one solve, shared DM), then fan the result out into a (rows, cols)
+    # shaped work variable so the caller's ``work_var.sym`` preserves the
+    # original tensor shape.
     if hasattr(expr, 'shape') and expr.shape != (1, 1):
         rows, cols = expr.shape
         n_components = rows * cols
 
-        # Get or create multi-component work variable
-        cache_key = f'_eval_work_{n_components}'
+        shape_key = f'{rows}x{cols}'
+        cache_key = f'_eval_work_{shape_key}'
+        flat_key = f'{cache_key}_flat'
+        projector_key = f'{cache_key}_projector'
+
         if not hasattr(mesh, cache_key):
+            # Tensor-shaped result variable that the caller sees via .sym
+            if rows == mesh.dim and cols == mesh.dim:
+                vtype_out = uw.VarType.TENSOR
+            else:
+                vtype_out = uw.VarType.MATRIX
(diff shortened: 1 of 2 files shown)


lmoresi:

**Scope:** `_project_to_work_variable` was effectively dead code for matrix expressions — the matrix branch tried `MeshVariable(num_components=rows*cols)` (a flat int), which can't infer `vtype`, so every call raised `ValueError` and bounced into `DDt.update_pre_solve`'s outer `except Exception` clause. That made the NS solver pay a try-fail-fallback cost on every step. This PR makes the matrix branch actually work, using `SNES_MultiComponent_Projection` against a flat `(1, Nc) MATRIX` var and fanning the result back into a `(rows, cols)` tensor work-var so the caller's `work_var.sym` preserves the original shape.

**Verification:**
- Shape detection: `if rows == mesh.dim and cols == mesh.dim → vtype_out = TENSOR` is the right default; non-square stays `MATRIX`.
- Flat→tensor fan-out indexing: `idx = i*cols + j` matches the comprehension order in `flat_source`, so `divmod(idx, cols) → (i, j)` is correct.
- Cache key `_eval_work_{rows}x{cols}` is shape-aware, so different tensors on the same mesh don't collide.
- Tests at three levels: shape-correctness on zero-strain rotation, value-correctness on simple shear, and the end-to-end NS solve.

**Copilot's two notes:**
1. *Symmetric tensors are projected redundantly* — true but optional. Current code projects all `rows*cols` independently and then writes both off-diagonals; for a symmetric source `expr == expr.T` the noise is bounded by the projection tolerance (1e-6). Worth a follow-up only if the SYM_TENSOR path shows measurable benefit.
2. *Test name asserts the fallback isn't triggered but only checks `solve()` completes* — fair point. Worth either monkeypatching the fallback to assert non-entry, or renaming to `test_navier_stokes_solve_with_tensor_evaluate_path`. Not blocking.

**Together with #183, this closes out #180.** #183 hardens the fallback path; this PR removes the reason the fallback gets entered. Belt + braces.

LGTM modulo standard maintainer eyeball.

PR #182, opened 2026-05-11
Consolidate and unify cached spatial indexing (KDTree)

bknight1:
This PR refactors how spatial indices (KDTree) are managed across Underworld3 to eliminate redundant construction overhead in solver loops.

### Key changes:
- **Unified Caching**: Introduced a consistent `_get_kdtree()` pattern across `Swarm`, `PICSwarm`, and `MeshVariable`. This ensures that a single C++ index is built per step and reused for all property updates.
- **Automatic Invalidation**: Spatial indices are now automatically invalidated and rebuilt only when necessary (e.g., after particle migration or mesh deformation).
- **Parallel Routing Fixes**:
  - Fixed an `IndexError` in `Swarm.migrate()` where a 1D rank array was incorrectly indexed as 2D during centroid-based routing.
  - Restored default `max_its=10` in `Swarm.advection` migration to ensure particles are correctly routed across multiple MPI partitions in high-velocity flows.

PR #191, opened 2026-05-15
_deform_mesh: bump _mesh_version + clear runner nav caches

lmoresi:
## Summary

Two cache-invalidation gaps in `Mesh._deform_mesh`, both triggered by direct `_deform_mesh(coords)` calls (which bypass the `mesh.X.coords` NDArray callback that normally performs this hygiene — e.g. every free-surface RK stage in the convection benchmark):

- **`_mesh_version` not bumped.** PR #182 introduced version-keyed kdtree navigation caches (`_BaseMeshVariable._get_kdtree`, `Mesh._get_domain_kdtree`) gated on `_mesh_version`. The `mesh.X.coords` callback bumps it; direct `_deform_mesh()` did not — so navigation kdtrees stayed frozen on the undeformed mesh. PR #188 added `_topology_version` invalidation here but missed `_mesh_version`.
- **Runner coord-identity nav caches not cleared.** A runner's `restore_points_to_domain` caches a kdtree keyed on `id(mesh.X.coords)`. `_deform_mesh` replaces `self._coords`, but CPython reuses freed ids → a fresh array can collide with the old id() and the staleness check false-negatives. Explicitly clearing `_restore_kdt` / `_restore_coords_id` defeats the id()-reuse hazard.

Brings `_deform_mesh` into line with the cache hygiene `mesh.adapt()` and `_legacy_access` already perform.

## Test plan

- [x] `_mesh_version` increments on a direct `_deform_mesh()` call (was frozen at 0 before the fix)
- [ ] Existing mesh/smoother test suites pass
- [ ] Parallel smoke (np=2) of a deforming-mesh case

Note: this is an independent correctness fix. It does **not** by itself resolve the separate free-surface convection feedback regression under investigation (verified — both fixes present in a clean build still reproduce the damped regime).

### DIFF
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -1844,6 +1844,37 @@ class Mesh(Stateful, uw_object):
                 self._dminterpolation_cache.invalidate_all(
                     reason="mesh deformed")
             self._topology_version += 1
+            # Bump the geometry-version counter so version-keyed
+            # kdtree navigation caches rebuild against the new DOF
+            # positions: _BaseMeshVariable._get_kdtree and
+            # Mesh._get_domain_kdtree both gate their rebuild on
+            # `_mesh_version`. PR #182 introduced those version-keyed
+            # caches; the mesh.X.coords callback path bumps
+            # _mesh_version, but direct _deform_mesh() calls (every
+            # free-surface RK stage) bypass that callback. Without
+            # this bump the navigation kdtrees stay frozen on the
+            # undeformed mesh — back-advected SL samples land at the
+            # wrong DOFs, corrupting the temperature field. PR #188
+            # added _topology_version invalidation but missed this.
+            self._mesh_version += 1
+            # Also nuke any *user-installed* navigation caches that
+            # key off coord-array identity. A runner's
+            # restore_points_to_domain typically installs a
+            # `_restore_kdt` / `_restore_coords_id` pair on the mesh,
+            # rebuilding the kdtree only when `id(mesh.X.coords)`
+            # changes. _deform_mesh replaces self._coords with a new
+            # object, but CPython reuses freed ids, so the fresh
+            # coords array can land on the SAME id() as the old one —
+            # the staleness check then false-negatives and the stale
+            # restore-kdtree is reused, snapping back-advected SL
+            # particles to pre-deform DOF positions (smeared T,
+            # damped free-surface convection feedback). Explicitly
+            # clearing these defeats the id()-reuse hazard. PR #188
+            # captured the generic cache invalidation above but
+            # dropped this runner-navigation-cache clear.
+            for _nav_attr in ("_restore_kdt", "_restore_coords_id"):
+                if hasattr(self, _nav_attr):
+                    setattr(self, _nav_attr, None)
 
             # Propagate coordinate changes to registered submeshes
             for submesh in self._registered_submeshes:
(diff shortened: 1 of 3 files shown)


lmoresi:
## Diagnosis — bundles a clean cache fix with a version-counter design issue

The red CI here is **not** flaky — it's a real semantic collision. The 3 failures are all in the snapshot suite (`test_0007_snapshot_inmemory`, `test_0008_snapshot_realsolver`):

```
SnapshotInvalidatedError: _mesh_version moved from 0 to 1 since snapshot.
mesh.adapt() rebuild on restore is scheduled for v1.2; v1 refuses rather than corrupt the DOF arrays
```

The snapshot system treats **any** `_mesh_version` bump as a DOF-invalidating topology change and refuses to restore. This PR's `self._mesh_version += 1` in `_deform_mesh` fires on node *movement* (ALE deform), which keeps the same topology — so the snapshot should still be restorable (coords are part of the payload). `_mesh_version` is overloaded across two different consumers:
- kdtree navigation caches want a **geometry** version (bump on node movement)
- the snapshot system wants a **topology** version (bump on adapt/rebuild) — PR #188 already added `_topology_version`

**Recommended resolution:** give node-movement its own counter (e.g. have the kdtree nav caches gate on a `_geometry_version` that `_deform_mesh` bumps, and leave the snapshot-watched `_mesh_version`/`_topology_version` for true topology changes). That also cleanly provides the version hook #216's projector-cache fix needs.

### Splitting out the clean parts now
This PR also carries two **independent, uncontroversial** changes that don't touch `_mesh_version`:
- the `_expr_hash` `sympy.Dummy` canonicalisation (fixes #194 — `evaluate()` minted a fresh Dummy per call so the lambdify cache **never hit**; real correctness/perf bug)
- removal of the flaky timing-based `test_lambdify_caching`

I'm extracting those into a separate small PR so they land for the release. Leaving this PR open for the geometry-vs-topology version split.

lmoresi:
## Correction — the clean parts already landed independently

Scratch the extraction plan: `development` already has both clean changes via commit `ca5c3ef` ("Fix lambdify cache never hitting on evaluate() — **closes #194**"). The `_expr_hash` `Dummy` canonicalisation and the flaky `test_lambdify_caching` removal are present on current `development`.

So this PR's **only remaining unique content** is the `_mesh_version += 1` + `_restore_kdt`/`_restore_coords_id` clear in `_deform_mesh` — i.e. exactly the change that collides with the snapshot invalidation semantics (the 3 red tests).

**Re-scoping this PR to the version-counter design only.** It needs the geometry-vs-topology split before it can land:
- a `_geometry_version` (bumped by `_deform_mesh` on node movement) that the kdtree nav caches gate on, and
- leave the snapshot-watched `_mesh_version` / `_topology_version` for true topology changes.

That also gives #216's projector-cache fix a clean staleness hook. I'd suggest a fresh branch off current `development` for that rather than rebasing this stale one. Leaving open as the tracking PR for the design.

lmoresi:
Closing as superseded. The core change here — bumping `_mesh_version` when the mesh deforms — has already landed on development via the #246 deform-gate (`deform()` bumps the version, lines 863-865 of discretisation_mesh.py). The remaining piece in this PR (version-bump on the direct `_deform_mesh` path) now *collides* with the later-merged snapshot version-gating: CI fails with `SnapshotInvalidatedError` because a deform-only change trips the snapshot refuse.

The correct fix is a geometry-vs-topology version split (deform bumps geometry, only adapt/refine bumps topology → snapshots restore across a deform), tracked in https://github.com/underworldcode/underworld3/issues/296. Closing this PR in favour of that focused work.

PR #200, opened 2026-05-21
projection: unit-aware smoothing_length API on the four Projection classes

lmoresi:
@benknight — this is the bit you asked me about: a length-scale-aware way to set the smoother on `Projection` (and friends) instead of the unit-less `smoothing` knob. Splitting it out of a larger meshing branch so it can land independently.

## Summary

Adds an L-valued, **unit-aware** `smoothing_length` accessor to the four projection classes:

- `SNES_Projection`
- `SNES_Vector_Projection`
- `SNES_Tensor_Projection`
- `SNES_MultiComponent_Projection`

The existing `smoothing` knob is preserved unchanged (units of length², historically used as a tiny numerical regulariser); `smoothing_length` is a parallel view with the physical convention `smoothing = smoothing_length²`. The two stay consistent: setting one updates both.

## Why — the mathematics

The L2-projection solvers all minimise a Tikhonov-regularised misfit

$$\min_u\; \tfrac12\!\int_\Omega (u-\tilde f)^2 \;+\; \tfrac{\alpha}{2}\!\int_\Omega |\nabla u|^2 ,$$

whose Euler–Lagrange equation is the **screened-Poisson** equation

$$u \;-\; \nabla\!\cdot\!\bigl(\alpha\,\nabla u\bigr) \;=\; \tilde f .$$

The free-space Green's function of $\bigl(1 - \alpha\nabla^2\bigr)$ decays as $\exp(-r/L)$ where

$$L \;=\; \sqrt{\alpha},$$

so the projection acts as a Gaussian-like convolution of width $L$, obtained implicitly by one elliptic solve — no kernel is ever assembled. In Fourier space the transfer function is

$$\widehat{u}(k) \;=\; \frac{1}{1 + k^{2} L^{2}}\,\widehat{\tilde f}(k),$$

so features at scale $\ll L$ are damped, features at $\gg L$ pass through, and features at $\sim L$ are attenuated by ≈ 1/2.

This makes $L$ — not $\alpha$ — the natural physical knob: you ask for "smooth this field at scale $L$" and the solver delivers it.

## API

```python
proj = uw.systems.Projection(mesh, V)

# L-valued, unit-aware
proj.smoothing_length = 0.05                        # plain ND scalar
proj.smoothing_length = 5 * uw.scaling.units.km     # Pint Quantity → ND'd

print(proj.smoothing)         # α = L²  (sympy)
print(proj.smoothing_length)  # Pint Quantity in m when a scaling
                              # context is set; otherwise √α
```

[...]

PR #201, opened 2026-05-21
Add SNES_Scalar.petsc_use_constant_nullspace for singular Poisson

lmoresi:
## Summary

Adds an opt-in `petsc_use_constant_nullspace` flag on `SNES_Scalar` that attaches a constant `MatNullSpace` to the Jacobian. Needed for scalar problems whose linear operator has a constant kernel and no Dirichlet BC to break it: pure-Neumann domains, closed manifolds (sphere, torus, periodic boxes).

Without it, PETSc either diverges on the singular system or returns an arbitrary solution within the constant null space, dependent on initial guess and solver path. With it, PETSc projects each KSP right-hand side onto the orthogonal complement of the nullspace and returns the minimum-norm (near-zero-mean) solution.

Default `False`, so every existing scalar solver is unaffected.

## Implementation

`src/underworld3/cython/petsc_generic_snes_solvers.pyx`:

- `_petsc_use_constant_nullspace = False` field in `SNES_Scalar.__init__`
- `petsc_use_constant_nullspace` property + setter (setter invalidates `is_setup` so the rebuild attaches the nullspace)
- `_attach_constant_nullspace()` helper: calls `snes.setUp()` to materialise the Jacobian, then sets `MatNullSpace().create(constant=True)` on operator + preconditioner + their transposes
- Single call site in `SNES_Scalar._setup_solver`

Mirrors the existing Stokes saddle-point nullspace pattern (`_attach_stokes_nullspace`) but specialised for the scalar case (no field decomposition, just the one Jacobian Mat).

## Example

```python
poisson = uw.systems.Poisson(box_mesh, u_Field=T)
poisson.constitutive_model = uw.constitutive_models.DiffusionModel
poisson.constitutive_model.Parameters.diffusivity = 1.0
poisson.petsc_use_constant_nullspace = True   # opt-in
poisson.f = x - 0.5    # zero-mean source — equation is solvable
poisson.solve()
# T has near-zero discrete mean and matches the analytic answer
# u(x,y) = -x^3/6 + x^2/4 + C  with C ≈ 0
```

## Why this matters / context

Surfaced while testing the `SphericalManifold` work (cd-1 mesh, separate WIP branch). Surface Poisson `-Δ_S T = z` on a closed sphere fails `DIVERGED_LINE_SEARCH` at iteration 0 without the nullspace declaration; with the flag it solves cleanly and recovers `T = z/2 + const` to discretisation error with monotone refinement convergence. The same flag works on any pure-Neumann scalar problem, manifold or not — flagging this as a small standalone change keeps the API addition independent of the larger manifold-mesh work.

## Test plan

- [x] `tests/test_1055_snes_scalar_constant_nullspace.py` — 2 tests:
  - `test_constant_nullspace_canonical_solution`: pure-Neumann Poisson on a 2-D box; verifies convergence on both paths, shape-correctness matches analytic to FE accuracy on both paths, and the flag pulls the discrete mean toward zero.
  - `test_constant_nullspace_property_setter`: property round-trip, setup invalidation.
- [x] Regression: `test_0001_meshes.py`, `test_0506_tensor_evaluate.py`, `test_1010_stokesCart.py` (25/25 Stokes tests) all pass.

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -2133,7 +2133,26 @@ class SNES_Scalar(SolverBaseClass):
 
         Used by scalar Poisson on a fully-Neumann domain or a closed
         manifold. See :attr:`petsc_use_constant_nullspace`.
+
+        The minimum-norm property holds for a solve from a zero initial
+        guess; with a warm start (``zero_init_guess=False``) the result
+        may differ by an additive constant in the nullspace.
         """
+        # The constant vector is only a genuine nullspace of the Jacobian when
+        # nothing pins the solution. Dirichlet (essential) BCs break the constant
+        # mode, so attaching the nullspace then is at best a no-op and at worst
+        # misleading. Warn rather than fail, since the flag may be set before BCs.
+        if len(getattr(self, "essential_bcs", [])) > 0 and uw.mpi.rank == 0:
+            import warnings
+            warnings.warn(
+                f"SNES_Scalar ({self.name}): petsc_use_constant_nullspace=True with "
+                f"{len(self.essential_bcs)} essential (Dirichlet) BC(s). The constant "
+                "vector is generally not a nullspace of the Jacobian when Dirichlet BCs "
+                "are present; use the constant nullspace only on fully-Neumann or "
+                "closed-manifold problems.",
+                stacklevel=2,
+            )
+
         self.snes.setUp()
         jacobian = self.snes.getJacobian()
         operator_matrix = jacobian[0]
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -1646,6 +1646,15 @@ class SNES_Scalar(SolverBaseClass):
         self.boundary_conditions = False
         # self._constitutive_model = None
 
+        # Constant-nullspace handling. When True, attach a MatNullSpace
+        # with the constant mode to the Jacobian operator before each
+        # KSP solve. Needed for scalar problems on a closed manifold or
+        # a fully-Neumann domain where the operator (e.g. Laplacian)
+        # has a 1-D constant kernel and the linear system is otherwise
+        # singular. Default False — every solver with any Dirichlet BC
+        # or reaction term is non-singular and shouldn't pay the cost.
+        self._petsc_use_constant_nullspace = False
+
         self.verbose = verbose
 
         self._rebuild_after_mesh_update = self._build  # Maybe just reboot the dm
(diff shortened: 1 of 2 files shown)


lmoresi:
Addressed the review on the constant-nullspace path:

- **Constant nullspace + Dirichlet BCs (the main concern)** — now emits a rank-0 warning when `petsc_use_constant_nullspace=True` while essential BCs are present, since the constant vector is generally not a nullspace of the Jacobian in that case. Warns rather than raises, because the flag may be set before BCs are added. Verified it fires on the misconfiguration.
- **Min-norm docstring** — clarified that the minimum-norm property holds for a zero-init-guess solve; a warm start can differ by a constant in the nullspace.

Deliberately **not** done:

- **Caching the `NullSpace` object** — unlike the Stokes nullspace (basis vectors, costly to rebuild), a constant nullspace is trivial to create, and a cached handle risks going stale across a DM/comm rebuild. Not worth the micro-optimisation, and slightly riskier, so I left creation inline.
- **MPI-safe test mean** — the rank-local `T.data[:,0].mean()` assertion would need a global reduction to be MPI-correct; flagging for a follow-up since the test is currently serial-only.

lmoresi:
## Investigation: this branch doesn't survive contact with current development

I rebased this branch onto current `development` (155 commits ahead). The rebase was **textually clean (no conflicts)**, but the feature is semantically stale and **fails on current development**:

- **`test_1055` (this PR's own feature test) fails.** With the nullspace toggled on vs off, the solve is **bit-identical** (`|mean_off| == |mean_on| == 3.743e-16`) — i.e. `petsc_use_constant_nullspace` is a **no-op** on current development, and the solution mean is already ~zero without it. The May-era premise (the system is singular and the nullspace fixes the drift) no longer holds; development's solver setup appears to handle the constant mode already.
- **`test_0004` (the new FMG checkpoint test, from #230) diverges** with this branch present (`getConvergedReason() == -6`), deterministically (3/3), even though it never enables the flag — while it **passes on plain development** (4/4). The clean rebase fuzzy-matched this PR's guarded `_attach_constant_nullspace` hunks into an evolved setup flow, so the feature code is active where it shouldn't be.

**Conclusion:** a mechanical rebase is unsafe here (clean but wrong). This needs re-implementing the feature by hand on current development — and first confirming it's still needed, since `test_1055`'s bit-identical result suggests development may already produce the minimum-norm solution for these systems.

The earlier CI red on this PR is from the same root cause (base moved since May), not from the review-nit warning commit — a `warnings.warn()` can't change convergence.

Recommend not merging as-is. Happy to help re-implement on a fresh branch off `development` (as we did for #200 → #234) once you've decided whether the feature is still required.

lmoresi:
Closing as redundant. Development already provides this exact capability as `constant_nullspace` (added in 074a660 for the Monge-Ampere smoother's pure-Neumann solve), including the cached nullspace object suggested in review here. Verified equivalent: `constant_nullspace=True` gives the minimum-norm solution (mean 3.7e-16) vs -0.224 drift when off. Use `solver.constant_nullspace = True` instead of `petsc_use_constant_nullspace`. Salvaging this PR's test_1055 to cover `constant_nullspace` (currently untested) in a follow-up.

PR #203, opened 2026-05-24
fix(interpolation): bypass DMLocatePoints when caller supplies a verified hint

lmoresi:
## Summary

When the caller of `DMInterpolationSetUp_UW` (UW3's fork of stock `DMInterpolationSetUp`) supplies a per-query cell hint, PETSc re-verifies the hint via `DMPlexLocatePoint_Internal` inside `DMLocatePoints`. On cell shapes where its in-cell test is geometrically subtle — notably 2-D simplex cells embedded in 3-D space — verification can reject a correct hint and fall back to a wrong-cell match, yielding wild FE evaluations.

This PR skips `DMLocatePoints` when:

  - the mesh is simplex (`mesh.dm.isSimplex()`), where the face-containment test is exact for the affine reference map, or
  - the mesh is a manifold (`mesh.dim != mesh.cdim`), where the kdtree-nearest cell on a closed surface correctly contains every query.

Non-simplex volume meshes (quads, hexes) keep the original `DMLocatePoints` path, since deformed cells can have non-planar faces where the kdtree-nearest cell can be wrong and PETSc's more rigorous search is needed.

## What's in the patch

- `src/underworld3/function/_function.pyx` — hint source in `petsc_interpolate` is `Mesh.get_closest_cells` (kdtree-nearest, no in-cell test). Comment explains the role.
- `src/underworld3/function/_dminterp_wrapper.pyx::create_structure` — gates whether the hint is passed to PETSc on `mesh.dm.isSimplex() or mesh.dim != mesh.cdim`. On non-simplex volume meshes the hint is suppressed (`NULL`) and PETSc's `DMLocatePoints` runs unchanged.
- `src/underworld3/function/petsc_tools.c::DMInterpolationSetUp_UW` — when `owning_cell` is provided, skip `DMLocatePoints` entirely. The `MPI_Allreduce(MIN, foundProcs)` after each rank claims its hinted points assigns each global point to the lowest-rank claimant, preserving the prior ownership convention. Lazily builds `pointVec` / `globalPointsScalar` only on the DMLocatePoints path.

Net: ~60 / -40 lines across the three files.

## Behaviour

| Mesh type | Path | Result |
|---|---|---|
| Non-simplex volume (StructuredQuadBox, hexes) | `DMLocatePoints` (unchanged) | No regression |
| Simplex volume (UnstructuredSimplexBox 2D/3D) | Bypass | Single-rank: same cell id as before. Multi-rank: ownership collapses to rank-0-claims-all (pre-existing parallel-evaluate work tracks this) |
| Manifold (`SphericalManifold`, when PPE lands) | Bypass | Eliminates wrong-cell matches; SLCN advection runs without `_evalf=True` RBF workaround |

## Test plan

- [x] `pytest tests/test_0820_deform_mesh_solver_rebuild_regression.py::test_stokes_velocity_updates_after_second_deform` — passes (this was the regression in the previous push; quad mesh now goes through DMLocatePoints unchanged)
- [x] `pytest tests/test_0000_imports.py tests/test_0001_meshes.py tests/test_0004_pointwise_fns.py` — 28/28
- [x] Spot-check `uw.function.evaluate` on quad and simplex meshes — analytic match
- [x] Validated on `feature/parallel-point-eval`: `probe_manual_p1_eval.py` on `SphericalManifold` — wild-evaluation rate 16% → 0
- [x] Validated on `feature/parallel-point-eval`: `probe_manifold_advection.py` — SLCN advection runs without `_evalf=True`
- [ ] Full Tier-A / Tier-B regression (CI)

## Copilot review feedback (addressed)

- ✅ Moved `pointVec` / `globalPointsScalar` construction into the `else` branch so the bypass path skips them.
- ✅ Sentinel test changed from `(PetscInt)owning_cell[p] >= 0` to `owning_cell[p] != (size_t)-1` to avoid implementation-defined unsigned→signed conversion.

## Known limitations / out of scope

- Multi-rank ownership on simplex / manifold meshes collapses to rank-0-claims-all because `get_closest_cells` doesn't carry an ownership filter. This is symmetric with the pre-existing behaviour for any single-rank UW3 evaluate (typical use). The general parallel-point-eval architecture is being addressed separately on `feature/parallel-point-eval`.
- The pre-existing `owning_cell[p]` access for `p >= n_local` in `redundantPoints=FALSE` mode (where N from `MPI_Allgatherv` can exceed each rank's hint-array length) is a separate latent issue, not touched by this PR.
- A separate upstream PETSc MR could add `DMInterpolationSetCellHints` to the stock `DMInterpolation` API (the stock wrapper unconditionally drops caller hints). Useful for non-UW3 PETSc users but unrelated to this UW3 bypass.

### DIFF
--- a/src/underworld3/function/_dminterp_wrapper.pyx
+++ b/src/underworld3/function/_dminterp_wrapper.pyx
@@ -135,12 +135,23 @@ cdef class CachedDMInterpolationInfo:
                 DMInterpolationDestroy(&self._ipInfo)
                 raise RuntimeError(f"DMInterpolationAddPoints failed with error {ierr}")
 
-        # Set up — calls DMLocatePoints which is COLLECTIVE on the mesh DM.
+        # Set up — calls DMLocatePoints (or, on simplex / manifold meshes,
+        # the DMLocatePoints-bypass path) which is COLLECTIVE on the mesh DM.
         # All ranks must call this, even with zero local points.
         # ignoreOutsideDomain=1: PETSc silently skips points it cannot locate
         # rather than crashing. This is essential — points_in_domain() uses a
         # kd-tree heuristic that can misclassify distant points as interior.
-        if n_points > 0:
+        #
+        # Hint policy: pass the UW3 kdtree-nearest-cell hint to PETSc only on
+        # meshes where it can be trusted as authoritative — i.e. simplex
+        # cells (planar faces, affine map, exact face-containment test) and
+        # manifold meshes (dim != cdim, where PETSc's own in-cell test is
+        # unreliable near 2-manifold simplex edges). On non-simplex meshes
+        # (quads, hexes), deformed cells can have non-planar faces and the
+        # kdtree-nearest cell can be wrong; pass NULL so PETSc's
+        # DMLocatePoints runs its own (more rigorous) cell-location search.
+        cdef bint use_hint = bool(mesh.dm.isSimplex()) or (mesh.dim != mesh.cdim)
+        if n_points > 0 and use_hint:
             cells_view = np.ascontiguousarray(self.cells)
             ierr = DMInterpolationSetUp_UW(self._ipInfo, dm, 0, 1, <size_t*> &cells_view[0])
         else:
--- a/src/underworld3/function/_function.pyx
+++ b/src/underworld3/function/_function.pyx
@@ -1151,8 +1151,13 @@ def petsc_interpolate(   expr,
             # CACHE MISS - Create structure and cache it
             cached_info = CachedDMInterpolationInfo()
 
-            # Get cell hints
-            # coords is already np.ndarray type (function signature ensures this)
+            # Cell hints from a kdtree-nearest-cell lookup. The downstream
+            # `create_structure` decides — based on the mesh's cell shape —
+            # whether to pass these to PETSc as authoritative (bypass path,
+            # simplex / manifold meshes whose face containment is exact for
+            # affine transforms) or as a guess for `DMLocatePoints` to
+            # verify (non-simplex meshes whose deformed faces can be
+            # non-planar and require PETSc's pseudo-inverse projection).
             cells = mesh.get_closest_cells(coords)
 
             # Create and set up DMInterpolation structure (EXPENSIVE)
--- a/src/underworld3/function/petsc_tools.c
+++ b/src/underworld3/function/petsc_tools.c
@@ -123,8 +123,17 @@ PetscErrorCode DMInterpolationSetUp_UW(DMInterpolationInfo ctx, DM dm, PetscBool
   */
   PetscInt *recovery_cells = NULL;
   PetscCall(PetscMalloc1(N, &recovery_cells));
+  /*
+    Handle the (size_t)-1 sentinel BEFORE casting to PetscInt. On builds where
+    sizeof(size_t) > sizeof(PetscInt) (32-bit-indices PETSc) the cast of
+    SIZE_MAX to a signed PetscInt is implementation-defined; rejecting the
+    sentinel via unsigned comparison first guarantees the cast only ever
+    runs on a valid non-negative cell id.
+  */
   for (PetscInt k = 0; k < N; ++k)
-    recovery_cells[k] = owning_cell ? (PetscInt)owning_cell[k] : -1;
+    recovery_cells[k] = (owning_cell && owning_cell[k] != (size_t)-1)
+                          ? (PetscInt)owning_cell[k]
+                          : -1;
   for (p = 0; p < numFound; ++p) {
     if (foundCells[p].index >= 0) {
       PetscInt orig_p = foundPoints ? foundPoints[p] : p;
--- a/src/underworld3/function/petsc_tools.c
+++ b/src/underworld3/function/petsc_tools.c
@@ -22,18 +22,17 @@ PetscErrorCode DMInterpolationSetUp_UW(DMInterpolationInfo ctx, DM dm, PetscBool
   PetscScalar       *a;
   PetscInt           p, q, i;
   PetscMPIInt        rank, size;
-  Vec                pointVec;
+  Vec                pointVec = NULL;
   PetscSF            cellSF;
   PetscLayout        layout;
   PetscReal         *globalPoints;
-  PetscScalar       *globalPointsScalar;
+  PetscScalar       *globalPointsScalar = NULL;
   const PetscInt    *ranges;
   PetscMPIInt       *counts, *displs;
   const PetscSFNode *foundCells;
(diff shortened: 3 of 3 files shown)


lmoresi:
Force-pushed an updated design (commit `17a5a8d`) and retargeted base back to `development`. 

The earlier push had a regression on quad meshes — `_get_closest_local_cells_internal` rejects boundary vertices on bounded meshes (test_0820 failed). Pivoted the design: the bypass is now gated on `mesh.dm.isSimplex() or mesh.dim != mesh.cdim`. Non-simplex meshes (quads, hexes) keep the original DMLocatePoints path; simplex and manifold meshes use the bypass.

Both Copilot review comments addressed in the new version (pointVec lazy-built, unsigned-sentinel comparison).

lmoresi:
**Update — precursor needed**

Discussed offline: `get_closest_cells` is by design a heuristic (always returns SOME cell, suitable for hinting PETSc, not authoritative). The current PR uses it as authoritative for the bypass — that's wrong in general; works on `SphericalManifold` only coincidentally.

The right authoritative hint source is `_get_closest_local_cells_internal` with looser semantics (`>= 0` instead of strict `> 0` in the in-cell test) so it accepts on-face queries — i.e., every mesh vertex, every point on a shared face. Opened as **PR #207** (`bugfix/in-cell-test-loose-semantics`).

Once #207 merges, I'll rebase this PR onto it and switch the hint source in `_function.pyx` from `get_closest_cells` back to `_get_closest_local_cells_internal`. The bypass will then be authoritative for simplex + manifold meshes; non-simplex meshes keep the original DMLocatePoints path.

lmoresi:
Update on the precursor: #207 now uses an opt-in design — `on_boundary=True` to get loose semantics. The first attempt (loose-by-default) broke an AdvDiff test that depends on boundary vertices being routed to RBF in `points_in_domain`. Opt-in keeps that behaviour and makes the loose path available where it's actually needed.

When #207 merges, I'll update this PR's `_function.pyx` to call `mesh._get_closest_local_cells_internal(coords, on_boundary=True)` — gives the authoritative per-rank cell-id the bypass needs without touching the default path used by everything else.

lmoresi:
Updated precursor #207 design: loose is now the default in `_get_closest_local_cells_internal`. So once #207 merges, the update to this PR is just:

```python
# in _function.pyx
cells = mesh._get_closest_local_cells_internal(coords)  # authoritative, no kwarg needed
```

No `on_boundary=True` opt-in plumbing in the bypass path. Cleaner.

lmoresi:
## Disposition of Copilot review comments (after `7168a0a`)

| # | Copilot concern | Status |
|---|---|---|
| 1 | `globalPointsScalar` + `VecCreateSeqWithArray()` still done unconditionally on the bypass path — wasteful allocations | **Already fixed** in current HEAD — the scalar buffer + `pointVec` are now inside `else { /* !owning_cell */ ... }` at lines 95–106, and the matching cleanup at 183–186 is guarded `if (!owning_cell)`. The hot bypass path now does only the unsigned-sentinel claim loop |
| 2 | `(PetscInt)owning_cell[p] >= 0` is implementation-defined when `size_t` holds `SIZE_MAX` | **Fixed in `7168a0a`** — reject the sentinel via unsigned comparison *before* the cast, so `(PetscInt)owning_cell[k]` only ever runs on a valid non-negative cell id. Mirrors the pattern already used at line 93 for `foundProcs` |

### Verification

Smoke test against the PR's intent — `uw.function.evaluate(v.sym, pts)` with a tri-simplex 2D mesh, including a corner vertex and an on-boundary point, returns the expected `x` values to 1e-9:

```python
mesh = uw.meshing.UnstructuredSimplexBox(minCoords=(0,0), maxCoords=(1,1), cellSize=0.2)
v = uw.discretisation.MeshVariable('v', mesh, 1, degree=2)
v.array[:, 0, 0] = v.coords[:,0]
pts = np.array([[0.3, 0.7], [0.5, 0.5], [0.0, 0.0], [1.0, 1.0]])
# → [0.3, 0.5, 0.0, 1.0]  ✓
```

Plus `test_0001_meshes.py`, `test_0004_pointwise_fns.py`, `test_0503_evaluate.py` — 25 passed.

### Note on namespace

This branch still calls `mesh.get_closest_cells(coords)` in `_function.pyx:1161`, which is independent of #207's `_get_closest_local_cells_internal(on_boundary=True)` change — so this PR doesn't need rebasing on #207 to merge. The loose-by-default semantics in #207 will apply automatically to any future caller that switches to `_get_closest_local_cells_internal` for cell hinting.

PR is ready to merge.

lmoresi:
## Why the FE trace-back matters — quantitative demo

Solid-body rotation of a Gaussian on a `SphericalManifold`, one full revolution (cs=0.075, P3, identical `monotone_mode="clamp"` for both runs — the **only** difference is the semi-Lagrangian trace-back interpolation):

| trace-back | rel-L2 vs analytic after 360° | amplitude retained |
|---|---|---|
| **FE** (`_evalf=False`, enabled by this bypass) | **1.3 %** | 98.4 % |
| **RBF** (`_evalf=True`, the prior workaround) | **20.4 %** | 96.1 % |

FE error stays essentially flat across the whole rotation; the linear RBF-Shepard fallback climbs steadily to ~20%. Visually, the FE blob returns to its start indistinguishable from the analytic Gaussian, while the RBF blob smears and distorts. That ~15× accuracy gap is the payoff of the `DMLocatePoints` bypass — it restores genuine high-order FE interpolation on the 2-manifold instead of falling back to O(h) RBF.

Reproducer (on the `feature/parallel-point-eval` base where `SphericalManifold` lives): `docs/examples/parallel_point_eval/compare_fe_rbf_lonlat.py`. An animated equirectangular FE | RBF | analytic comparison is attached below.


https://github.com/user-attachments/assets/eb292752-a7e9-4656-99f7-ca0623e1318b

lmoresi:
Closing as superseded.

The core bypass-DMLocatePoints change from this PR (`17a5a8d`) has been independently ported into `development` via PR #213 (commit `c52201b` "remesh field-transfer redesign (Phase 1 + Phase 2) + parallel locator hardening"). The merged version is a refinement of the same C-side bypass with updated docs referencing the parallel mesh-adapt and `SphericalManifold` use cases.

What's left in this branch that *isn't* in development:

- `7168a0a` — guards the `(size_t)-1` no-claim sentinel via unsigned comparison before the `(PetscInt)owning_cell[k]` cast in `petsc_tools.c`. The cast UB only bites on 32-bit-indices PETSc builds (uncommon for UW3) — portability/future-proofing rather than a production bug. If we want this on `development`, the cleanest path is a fresh 1-line PR off `development`, not a rebase of this branch.

PR #205, opened 2026-05-24
Fix XDMF output to explicitly write cell-to-vertex topology and ensure strict ParaView compatibility

tiannh7:
This PR resolves a critical issue where generating XDMF files could silently output PETSc DMPlex's internal point numbering instead of zero-indexed vertex connectivity. This malformed connectivity causes abrupt crashes in ParaView 5.13+ and renders corrupted meshes in older versions. 

**Key changes:**
* **Standard-compliant `/viz` generation:** Ported the `_write_mesh_viz_groups` generation logic into `write_timestep`. It now correctly gathers and constructs true `/viz/geometry/vertices` and `/viz/topology/cells` across MPI ranks. **This restores strict XDMF schema compliance and is 100% backward-compatible with all older versions of ParaView.**
* **Proactive validation:** Added safety checks in `checkpoint_xdmf` to explicitly warn users if the connectivity goes out of bounds (`cells.max() >= numVertices`), if `numCorners` mismatches the topology type, or if Field attribute lengths (`Node`/`Cell` center) are malformed.
* **Unit tests:** Added a new test case to enforce that `/viz` groups are consistently generated and XDMF reference paths correctly point to them without triggering validation warnings.

While this PR hardens our existing XDMF/HDF5 pipeline and ensures stability, it might be worth exploring alternative or complementary formats (such as native `.vtu` / `.vtk`) down the road. Integrating tools like **`pyvista`** or **`meshio`** could potentially offer a more integrated, single-file visualization workflow. This is just a thought for the future.

### DIFF
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -4117,7 +4117,7 @@ def _local_viz_cell_connectivity(mesh):
     """Return local cell-to-vertex connectivity in global vertex ids."""
 
     dm = mesh.dm
-    pStart, _ = dm.getDepthStratum(0)
+    pStart, pEnd = dm.getDepthStratum(0)
     cStart, cEnd = dm.getHeightStratum(0)
     vertex_numbering = dm.getVertexNumbering().getIndices()
     vertex_gids = _petsc_numbering_to_global_ids(vertex_numbering)
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -4197,19 +4197,43 @@ def checkpoint_xdmf(
     numCorners = cells.shape[1]
     cellDim = topo["cells"].attrs["cell_dim"]
 
+    import warnings
+    import numpy as np
+
+    if topoPath == "topology":
+        warnings.warn(
+            "Using raw '/topology/cells' for XDMF. This may not be Paraview-compatible. "
+            "Expected '/viz/topology/cells'."
+        )
+
+    cells_data = cells[...]
+    c_min, c_max = cells_data.min(), cells_data.max()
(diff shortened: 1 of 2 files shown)


bknight1:
@lmoresi I'm also experiencing ParaView crashes when viewing mesh xdmf output, which this PR fixes. I've tested and confirmed working. 

commit 22dd9ab2 on feature/annulus-spherical-benchmark-debug takes a different approach (adds `petsc_format` parameter to `write()` to switch between HDF5_VIZ and HDF5_PETSC). Worth comparing the two strategies.

lmoresi:
Merged at `29a2029` — thanks @tiannh7! 🙏 First contribution and it lands a real ParaView-crash fix for everyone, which is excellent.

### For context — there's a complementary approach worth keeping visible

@bknight1's earlier comment correctly noted that @gthyagi's commit `22dd9ab2` (on the now-deleted `feature/annulus-spherical-benchmark-debug` branch) addresses the same bug from a different direction:

- Adds `petsc_format: bool` kwarg to `Mesh.write()` and uses PETSc's `HDF5_VIZ` viewer format
- PETSc itself writes `/viz/topology/cells` natively — ~10 lines of behavior change vs. the ~115-line Python+MPI helper landed here
- Also tightens up `checkpoint_xdmf` (integer topology precision from dtype, removes HyperSlab wrappers)

It hadn't become a PR before the branch was deleted, and we didn't want to block the working fix here on reviving someone else's work.

### Tracking issue

I've opened **#217** to keep the alternative visible and document what's needed to evaluate / revive it. @gthyagi — if you're up for cherry-picking `22dd9ab2` onto current `development` as a fresh PR, the manual `_write_mesh_viz_groups` helper from this PR could come back out and the topology path would be entirely PETSc-managed.

tiannh7:
@lmoresi Thanks

PR #208, opened 2026-05-25
Promote monotone interpolation to a universal evaluator flag

lmoresi:
## Summary

Adds an opt-in `monotone` keyword (default `False`) to `uw.function.evaluate` and `global_evaluate`. When truthy (`"clamp"` / `"pick"`, with `True` aliasing `"clamp"`), the interpolated result is bounded to the local data range of the source field: for each query point the `mesh.dim + 1` nearest source DOFs define `[min, max]`; `"clamp"` clips into that range, `"pick"` keeps in-bounds values and re-evaluates only the out-of-bounds subset via bounded RBF.

This lifts the kNN clamp/pick limiter that previously lived **only** inside the semi-Lagrangian trace-back (`SemiLagrangian.update_pre_solve`) into a shared evaluator helper (`_apply_monotone_limit`), so any resampling path (remesh re-interpolation, projection, restart, the SL DDt itself) can request the same bounded result from one place. The source field is recovered from the expression via `meshVariable_lookup_by_symbol`.

This is the standalone **axis C** of the free-surface mesh-motion work — independent of the ALE/mesh-motion axes (A/B), and lands on its own.

## Bit-identical guarantee

`monotone=False` is **provably** bit-identical to today: the existing `evaluate` / `global_evaluate` bodies are renamed unchanged to `_evaluate_impl` / `_global_evaluate_impl`, and the public functions are thin wrappers that return the impl result untouched unless `monotone` is truthy. `ddt.py`'s inline limiter block is deleted and routed through the new option (`monotone=monotone_mode`).

A full `AdvDiffusionSLCN` trajectory (steep blob, rotational advection, P3) is bit-identical pre/post refactor — `max|Δ| = 0` for **all** three modes (`None`, `"clamp"`, `"pick"`), not just the default.

## Scope & flagged follow-ups (MVP)

- **Single-MeshVariable expressions only**; composites raise `ValueError` (the neighbour bound across fields is ill-defined).
- `TODO(parallel)`: the neighbour bound is rank-local, and `"pick"`'s collective re-eval is gated on a rank-local mask → serial-only. Lifted verbatim from the validated (serial) limiter; the nav-clone overlap machinery is noted as the hardening hook.
- `TODO(units)`: bounds come from `var.data` (always non-dimensional) vs the dimensional value — a pre-existing latent mismatch, harmless while scaling is inactive (as in the validated baseline). Flagged, not fixed.

## Tests

- New `tests/test_0760_evaluate_monotone.py` — locks the API: default no-op bit-identical, `clamp`/`pick` bounding against independently recomputed neighbour ranges, composite / unknown-option refusal.
- `tests/test_1054_advdiff_monotone_mode_kwarg.py` — gains end-to-end solver-integration coverage.
- `pytest -m "level_1 and tier_a"`: 158 passed, 5 skipped (MPI), 0 failed.

### DIFF
--- a/src/underworld3/function/functions_unit_system.py
+++ b/src/underworld3/function/functions_unit_system.py
@@ -628,7 +628,8 @@ def _normalize_monotone(monotone):
     )
 
 
-def _apply_monotone_limit(expr, coords, value, mode, evalf=False):
+def _apply_monotone_limit(
+    expr, coords, value, mode, coord_sys=None, other_arguments=None):
     """Bound an interpolated result to the local data range of its source
     field (monotone / bounded interpolation).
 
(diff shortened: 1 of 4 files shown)


lmoresi:
Thanks @Copilot — addressed in e171fb2. Point by point:

- **Fail-fast validation** (`evaluate` / `global_evaluate` ran the impl before validating `monotone`): fixed. `monotone` is now normalized at the top of both wrappers, and the unsupported `monotone` + `check_extrapolated` combination in `global_evaluate` is rejected *before* the impl call — an unknown option or bad combo no longer does a full, wasted evaluation.

- **`pick` + MPI deadlock** (collective `global_evaluate` gated on a rank-local `oob_mask.any()`): the bounded RBF re-evaluation is now guarded — `monotone="pick"` raises a clear `NotImplementedError` under MPI pointing to `monotone="clamp"` (which is rank-local and parallel-safe), instead of hanging. The `TODO(parallel)` for a true global-neighbour bound (via the nav-clone overlap machinery) stays.

- **`pick` not propagating evaluation settings**: `coord_sys` / `other_arguments` are now propagated into the re-evaluation so it stays in the caller's frame. `evalf=True` remains a *deliberate* override — the FE result is exactly what overshot, so the re-eval intentionally uses the bounded RBF path and does not honour `mode`/`rbf`/`force_l2`; this is now documented in the helper.

- **`coords[oob_mask]` assuming an ndarray**: investigated — the premise doesn't hold for this path. `_evaluate_impl` rejects non-array coords upstream (`non_dimensionalise` raises on a plain `list`/`tuple`) *before* monotone runs, so the mask only ever sees an `ndarray` / `UnitAwareArray`. No coercion needed; added a clarifying comment rather than dead defensive code.

- **Rank-local neighbour bound**: already flagged `TODO(parallel)`; the true global-neighbour (halo-aware) bound is deferred to a follow-up per the PR plan, with the nav-clone overlap clone noted as the hook.

Defaults stay bit-identical (full `AdvDiffusionSLCN` trajectory `max|Δ|=0` across `None`/`clamp`/`pick`). `level_1 and tier_a`: 158 passed; `test_0760` + `test_1054`: 14 passed.

PR #216, opened 2026-05-29
[BUG] - Fix three stale-cache bugs after swarm particle addition

### LINKED ISSUE #215: [BUG] - Three stale-cache issues after swarm particle addition
## Summary

Three independent stale-cache issues were found when a swarm is modified via `add_particles_with_coordinates()` and then used for interpolation or projection. All three cause silent data corruption or MPI deadlock (PETSc < 3.24) and affect the common pattern of re-adding particles to empty cells each timestep.

---

### Bug 1 — Stale kd-tree in `add_particles_with_coordinates`

**File:** `src/underworld3/swarm.py:3571-3577`

`add_particles_with_coordinates()` calls `self.dm.migrate()` (the raw PETSc DMSwarm migration) rather than `self.migrate()` (the UW3 wrapper). The manual cache invalidation at lines 3574-3577 nils `_particle_coordinates._canonical_data` and each variables `_canonical_data`, but **misses `self._kdtree`**:

```python
# Lines 3574-3577 (BEFORE fix)
self._particle_coordinates._canonical_data = None
for var in self._vars.values():
    if hasattr(var, "_canonical_data"):
        var._canonical_data = None
# missing: self._kdtree = None
```

By contrast, `Swarm._invalidate_canonical_data()` at line 2648 correctly sets `self._kdtree = None`. It is called by `self.migrate()` (line 3463), but `add_particles_with_coordinates` bypasses that path.

**Effect:** After adding particles, `swarm._get_kdtree()` returns a kd-tree built from OLD particle coordinates. RBF interpolation (both for proxy mesh variables and for `uw.function.evaluate()` with `rbf=True`) looks up particle indices from the stale tree, accessing wrong PETSc memory locations. On PETSc 3.22.2 this produces an MPI deadlock inside the kd-tree query → `rbf_evaluate` → `update_lvec` path; on PETSc 3.24.2 it silently returns wrong interpolated values.

**Fix applied:** Added `self._kdtree = None` after line 3573.

---

### Bug 2 — Stale cached projector in `_project_to_work_variable`

**File:** `src/underworld3/function/_function.pyx:529-642`

`_project_to_work_variable()` caches `Projection` solver instances on the mesh object as `_eval_projector_scalar` (scalar) or `_eval_{shape}_projector` (tensor). The solver is created once and reused across all subsequent `evaluate()` calls on that mesh:

```python
if not hasattr(mesh, "_eval_projector_scalar"):
    mesh._eval_projector_scalar = uw.systems.Projection(mesh, ...)
projector = mesh._eval_projector_scalar
projector.uw_function = scalar_expr
projector.solve(zero_init_guess=False)  # no _force_setup
```

When a Stokes solve (or any other solver modifying the DM) runs between two `evaluate()` calls, the cached projectors PETSc solver state (SNES/KSP/matrix decomposition) is stale. On PETSc 3.22.2 the `projector.solve()` deadlocks because the cached matrix doesnt match the current DM state. PETSc 3.24.2 tolerates this (silently returns wrong results).

**Fix applied:** Changed both the scalar projector (line 640) and the tensor projector (line 613) to pass `_force_setup=True`:
```python
projector.solve(zero_init_guess=False, _force_setup=True)
```

**Note:** The same stale-cached-projector pattern exists in user code that reuses `Projection` solver instances across timesteps or after Stokes solves. Any cached projection solver should either (a) pass `_force_setup=True` on every solve, or (b) track a DM version counter and auto-rebuild when the DM changes.

---

### Bug 3 — Stale proxy mesh variable data after swarm write

**File:** `src/underworld3/swarm.py:1034-1087` (proxy update pipeline)

When a `SwarmVariable` has `proxy_degree > 0` (the default is `proxy_degree=2`), a proxy `MeshVariable` is created that stores RBF-interpolated values from the swarm. The update is lazy:

1. `swarm.access(var)` modifies the canonical data array
2. On exit, `delay_callbacks_global` fires the data callback
3. The callback calls `pack_raw_data_to_petsc()` (line 478), which writes to PETSc and calls `self._update()` (line 1291), setting `self._proxy_stale = True`
4. The actual re-interpolation (`_rbf_to_meshVar`) happens only when `material.sym` is accessed or `_update_proxy_if_stale()` is called

**The problem:** If code reads the proxys `MeshVariable` DM directly (e.g., a `Projection` solver that evaluates its `uw_function` at quadrature points), it reads STALE data from the proxys PETSc DM — the lazy update hasnt fired yet.

**Concrete scenario:**
```python
material = swarm.add_variable("material", 1, dtype=int, proxy_degree=2)
meshMat.uw_function = material.sym[0]  # triggers proxy update, stores symbol

# ... add particles and set new material values ...
meshMat.solve(_force_setup=True)
# ^ evaluates stored proxy symbol at quadrature points
# ^ proxy DM still contains data from the FIRST sym access — STALE
```

**Why `uw.function.evaluate(material.sym[0], ...)` works:** It re-accesses `material.sym`, which calls `_update_proxy_if_stale()` and re-interpolates from the current swarm.

**Fix needed:** Either:
- (a) Document that `_update_proxy_if_stale()` must be called before using the proxy MeshVariable DM after a swarm write
- (b) Make the evaluation pipeline check for stale proxies and auto-update before reading
- (c) Remove the lazy proxy update pattern and update immediately on data write
- (d) Add proxy update hooks in `add_particles_with_coordinates` and other swarm-mutating methods

---

## Reproduction

The test file `tests/test_0112_swarm_add_particles.py` contains `test_proxy_updates_after_add_particles` which reproduces Bug 1 (kd-tree) and Bug 3 (proxy staleness). Bug 2 was reproduced on Setonix HPC (PETSc 3.22.2) and confirmed locally on macOS (PETSc 3.24.2).

## Environment

- PETSc 3.22.2 (Setonix HPC) — deadlocks on Bugs 2
- PETSc 3.24.2 (macOS) — silently returns wrong values on Bugs 2
- Underworld3 `development` branch (as of 2026-05-29)

## Related Files

- `src/underworld3/swarm.py` — lines 3501, 3571-3577 (Bug 1), lines 1034-1087 (Bug 3)
- `src/underworld3/function/_function.pyx` — lines 529-642 (Bug 2)

bknight1:
## Summary

Fixes three stale-cache bugs that occur when a swarm is modified via `add_particles_with_coordinates()` and then used for interpolation or projection. These affect the common pattern of re-adding particles to empty cells each timestep.

Fixes #215

### Bug 1 — Stale kd-tree in `add_particles_with_coordinates`

`add_particles_with_coordinates()` calls `self.dm.migrate()` directly, bypassing `Swarm.migrate()` which calls `_invalidate_canonical_data()`. The manual invalidation in `add_particles_with_coordinates` set `_canonical_data = None` for coordinates and variables but missed `self._kdtree`. After particle addition, `swarm._get_kdtree()` returned a kd-tree built from old coordinates.

**Fix:** Added `self._kdtree = None` in the invalidation block in `add_particles_with_coordinates`.

### Bug 2 — Stale cached projector in `_project_to_work_variable`

`_project_to_work_variable()` caches `Projection` solver instances on the mesh object (`_eval_projector_scalar` and `_eval_{shape}_projector`). When reused across `evaluate()` calls after a Stokes solve modifies the DM, the cached projectors PETSc solver state is stale. On PETSc 3.22.2 (Setonix HPC) this causes an MPI deadlock; on PETSc 3.24.2 (macOS) it silently returns stale results.

**Fix:** Added `_force_setup=True` to both the scalar (line 640) and tensor (line 613) `projector.solve()` calls.

### Bug 3 — Stale proxy mesh variable after swarm write (documented, no fix)

The lazy proxy update pattern marks the proxy as stale on data write (`_proxy_stale = True`), but the actual re-interpolation (`_rbf_to_meshVar`) only fires when `material.sym` is accessed or `_update_proxy_if_stale()` is called. Code that reads the proxy MeshVariable DM directly (e.g. a `Projection` solver evaluating its `uw_function` at quadrature points) reads stale data.

**Workaround:** Call `material._update_proxy_if_stale()` (or access `material.sym`) after modifying a swarm variable and before using its proxy mesh variable.

### Test

Added `tests/test_0112_swarm_add_particles.py` with `test_proxy_updates_after_add_particles` that verifies the proxy mesh variable correctly reflects swarm changes after adding particles.

### Platform-dependent behavior

See issue #215 comment for details. PETSc 3.22.2 deadlocks on these bugs; 3.24.2 silently returns wrong values.


lmoresi:
## Review — good diagnosis, but the projector fix is too blunt (memory regression)

The three stale-cache bugs in #215 are real and the swarm-side work here is good. One change needs rework before merge.

### ✅ Good
- **`_invalidate_canonical_data()` consolidation** — replaces the repeated `_canonical_data = None` loops in `add_particles_with_coordinates`, `apply_snapshot_payload`, and `disk_snapshot._read_swarm_from_sidecar` with one method that also marks proxies stale via `var._update()`. Clean DRY win and fixes Bugs 1 & 3's invalidation path.
- The Bug 3 `TODO(BUG)` marking the lazy proxy-DM staleness is a good signpost.

### ⚠️ Blocker — `_force_setup=True` in `_project_to_work_variable` (`_function.pyx`)
The CI failure is **not** in the swarm tests (those pass) — it's `test_0006_memory_leak.py::test_stokes_advdiff_memory_leak`: **56.98 MiB growth vs a 30 MiB threshold**.

Root cause: the projector in `_project_to_work_variable` is **cached on the mesh** (`mesh._eval_work_{shape}_projector`, built once behind the `if not hasattr(mesh, cache_key)` guard) and reused across every `uw.function.evaluate()` that resolves a derivative/tensor expression. Passing `projector.solve(..., _force_setup=True)` tears down and rebuilds that cached projector's PETSc DM/SNES/matrices on **every** call. UW3's DM destroy/recreate is a known object-leak path (issue #176), so a Stokes+advdiff timestepping loop that evaluates each step accumulates exactly the kind of steady growth the test caught. It's also a performance regression (full re-assembly per evaluate).

### Suggested fix
Gate the rebuild on an actual staleness signal rather than forcing it unconditionally — e.g. store the `mesh._mesh_version` (and/or a swarm change counter) at projector-setup time and only `_force_setup=True` when it has advanced since the cached projector was built. That preserves the Bug 2 correctness fix (stale projector after a Stokes/DM modification) without rebuilding when nothing changed. Alternatively, ensure the `_force_setup` rebuild path fully destroys the prior DM/SNES/Mat/Vec before recreating.

### Logistics
This PR's CI last ran 2026-05-29 against a `development` that has since moved a lot (several PRs merged today). Please **rebase on current `development` and re-run CI** — then we can confirm whether the memory failure persists after a version-gated fix.

Happy to push the version-gated projector invalidation if you'd like. Leaving #216/#215 open.

lmoresi:
Updated this PR: merged current development (168 commits, clean) and resolved the CI failure (test_0006_memory_leak: 128 MiB growth → **16.9 MiB**, under the 30 MiB ceiling).

**Diagnosis** — both cache fixes re-introduced the projection/proxy leak that test_0006 guards:

1. **`_project_to_work_variable` `_force_setup=True`** (Bug 2) forced a full DM+SNES rebuild on *every* derivative `evaluate()` (reached via AdvDiffusion's semi-Lagrangian flux), ~75 MiB. The staleness it guarded is already handled by the current cache machinery — a plain solve refreshes correctly against changed field data (verified: `dT/dy` 1 → 3 across a data change). Reverted to a plain solve.
2. **`_invalidate_canonical_data()` calling `var._update()`** — that method is *eager* for IndexSwarmVariable (`_update_proxy_variables` → RBF re-interpolation), so a proxy was re-interpolated on every `swarm.access` write, ~35 MiB. The comment's intent was to *mark* the proxy stale; now sets `_proxy_stale = True` directly (lazy).

**Preserved:** Bug 1 (the `_kdtree` invalidation in `add_particles` — the real #215 fix) and the `_invalidate_canonical_data()` consolidation. Correctness verified: test_0112 15/15, a same-expression/changed-data staleness probe passes.

Thanks @bknight1 — the cache-invalidation diagnosis and consolidation are yours; this just swaps the two over-eager rebuilds for lazy invalidation.

PR #219, opened 2026-06-05
Constrained free-slip via recoverable Lagrange multiplier (dynamic topography)

lmoresi:
## Summary

Adds `uw.systems.Stokes_Constrained` (`SNES_Stokes_Constrained`): enforce `u·n = g`
on curved boundaries with a **recoverable Lagrange multiplier** instead of a tuned
penalty. The converged multiplier is the normal traction holding the boundary — a
direct **dynamic-topography** estimate `h = λ/(Δρ g)`.

Motivated by the fragility of penalty/Nitsche free-slip on annulus/spherical
boundaries (the magnitude must be tuned against forcing strength and viscosity; too
weak gives spurious radial throughflow, too strong diverges).

## How it works

An **augmented-Lagrangian (ALG2) outer loop** reusing the existing penalty BC: each
Stokes solve carries `t = [λ + r(u·n − g)]·n`, and `λ ← λ + r(u·n − g)`. The penalty
term preconditions every boundary mode (fast convergence); the multiplier removes the
penalty's accuracy bias (exact constraint from a *moderate*, well-conditioned `r`).

**Purely additive** — the validated 2×2 saddle-point assembly and fieldsplit
configuration are untouched (honours "solver stability is paramount"). The outer loop
lives inside `solve()`.

## Hands-off `solve()` (no tuning)

- **Relative tolerance**: `RMS(u·n − g) < rtol·RMS|u|` (default `rtol=1e-3`) — problem-independent.
- **Viscosity-weighted augmentation**: default `r = 10³·μ(x)`, so high-viscosity
  boundary regions are constrained as well as low-viscosity ones (a flat `r` is not).

```python
stokes = uw.systems.Stokes_Constrained(mesh, velocityField=v, pressureField=p)
stokes.constitutive_model = uw.constitutive_models.ViscousFlowModel
stokes.constitutive_model.Parameters.shear_viscosity_0 = mu
lam  = stokes.add_constraint_bc("Upper", g=0.0)   # free-slip, no penalty coefficient
stokes.solve()
topo = stokes.topography("Upper", buoyancy_scale=delta_rho_g)
```

## Validation

- **Annulus** (`tests/test_1061_constrained_freeslip_annulus.py`, 5 tests, level_2/tier_b):
  `RMS(u·n)=6.5e-5`, velocity within 0.3% of the penalty reference, 2 outer iterations,
  interior λ exactly 0, boundary `corr(λ, −n·σ·n) = 0.9999`.
- **SolCx** (variable viscosity): ~2e-4 vs Dirichlet reference in 3 outer iterations across
  contrast 1 → 10⁶, no manual tuning.

## Trade-offs and roadmap

The augmentation `r` is a *speed* knob, not an *accuracy* knob (accuracy is
`r`-independent — the AL property): too small just costs iterations, too large stiffens
the inner solve, the answer is never wrong. True-work is U-shaped with an efficient
basin `r ∈ [300, 10⁴]`.

Deferred (see `docs/developer/design/CONSTRAINED_FREESLIP_MULTIPLIER.md`):
adaptive `r₀` for very coarse meshes; **parallel** (mask + nodal update are serial);
rotation-null-space removal for both-boundaries free-slip; and the monolithic
`P'=[p,h]` fieldsplit ("arbitrary constraints in the saddle point") as a separate
feasibility spike.

Design note: `docs/developer/design/CONSTRAINED_FREESLIP_MULTIPLIER.md`.

### DIFF
--- a/docs/developer/design/CONSTRAINED_FREESLIP_MULTIPLIER.md
+++ b/docs/developer/design/CONSTRAINED_FREESLIP_MULTIPLIER.md
@@ -194,6 +194,42 @@ evidence): the monolithic 3-field / co-dim-1 representation; 3D spherical shells
 space needing explicit removal); and live free-surface equilibrium integration
 (pass `λ` as the target normal-stress end-state).
 
+## Monolithic `P'=[p,h]` fieldsplit — feasibility spike
+
+A future direction (and a general "inject arbitrary constraints into the saddle
+point" capability): rather than a third field forcing a nested 3-way Schur, group
+pressure and the multiplier into a composite `P' = [p, h]` and keep a **2-way
+`u | P'`** split.
+
+**Spike result (confirmed):** a 3-field DM `(u, p, h)` on a real mesh, with the `p`
+and `h` index sets grouped (`pc_fieldsplit_1_fields 1,2`, or an explicit
+concatenated IS), produces exactly a 2-block `u | [p,h]` Schur fieldsplit
+(block sizes 84 | 84 = u | (p+h) on a coarse test). The nested-Schur /
+KSP-reconfiguration concern is therefore moot — the split structure is identical to
+the current `u | p` solver. (`/tmp/spike_pph.py`.)
+
+**Remaining work (bounded, ~2 weeks of Cython, behind a subclass):**
+- Register `h` as field 2 (one `dm.setField`).
+- `h`-equation residual: boundary part `∫_Γ ψ(n·u − g)` (the existing Nitsche path
+  already registers field-1 boundary residuals — same pattern) plus a small interior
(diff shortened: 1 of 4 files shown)


lmoresi:
Superseded by #224. The constrained free-slip work is replaced by the **in-saddle (single coupled solve)** `Stokes_Constrained` solver, which is more production-ready: one mesh-independent solve (no fragile outer loop), validated against the exact SolCx analytic solution, with a lossless boundary-only multiplier reduction. The augmented-Lagrangian outer-loop variant introduced here is removed in #224 (it's easy to reproduce in Python if wanted). Closing in favour of #224.

PR #222, opened 2026-06-06
global_evaluate: faithful parallel evaluate() (fix out-of-domain mislocation)

lmoresi:
## Problem
`global_evaluate`'s parallel path silently returned **wrong** values for query points outside the (old) domain. The evaluation-swarm `migrate` routes unclaimed points by nearest rank centre-of-mass and strands them on an arbitrary rank, which extrapolates from a geometrically-far local cell. Deterministic reproduction (rotation gate, **linear** field `T=x`, np=5): a point at `x=+0.64` read `−0.42` (opposite side), `max_err=1.06`. This corrupted mesh-variable transfer on parallel mover-adapted meshes.

## Fix
Make `global_evaluate` a faithful parallel `evaluate()` — interpolate inside, extrapolate from the **true nearest cell** outside, flag inside/outside. A best-claim out-of-domain fallback in `global_evaluate_nd`:
- `allgather` the (small, boundary-layer) extrapolated set,
- each rank reports nearest-local-cell distance + its **local rbf** extrapolation,
- `Allreduce(MIN dist / MIN rank / SUM winner-value)` picks the globally-nearest rank's value.

Only unconditional collectives + a local `rbf_evaluate` (never the *collective* FE interpolation, which would desync per-rank → hang). O(boundary points), no dense global tree. **Deadlock-safe by construction.**

Default on; `GE_LOCAL_FALLBACK=0` restores legacy. Serial path unchanged (gated `mpi.size>1`).

## Validation
Rotation gate, `T=x`, np=5: `max_err 1.06 → 0.003` (bit-identical to serial `0.003`); turning the fallback off reproduces `1.06`. tier-A green; used throughout the parallel adaptive-convection runs.

### DIFF
--- a/src/underworld3/function/_function.pyx
+++ b/src/underworld3/function/_function.pyx
@@ -359,6 +359,7 @@ def global_evaluate_nd(   expr,
                 check_extrapolated=False,
                 force_l2=False,
                 smoothing=1e-6,
+                local_fallback=True,
             ):
 
     """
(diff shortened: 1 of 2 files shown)


lmoresi:
Pushed fixes for the review nits:

- **`GE_LOCAL_FALLBACK` case-sensitivity** — now normalized with `.strip().lower()`, so `False`/`OFF`/`No` disable the fallback. (Also: `local_fallback` is now a proper kwarg, not env-only.)
- **"true nearest cell" docstring** — reworded to reflect the actual centroid-distance heuristic (lowest rank wins ties in parallel).

On the other two threads:

- **`ext_flag` always-true** — this is correct by construction, not a bug: the points fed to the rbf fallback (`all_ext`) are exactly the points no rank could locate in-cell, so flagging them all as extrapolated is right, and writing `return_mask` back to True for them is a no-op. Happy to drop the redundant `ext_flag` plumbing as cleanup if preferred.
- **Missing regression test** — agreed this is the real gap. A parallel out-of-domain test (asserting rank-independence) would lock the behaviour; flagging for a follow-up unless we want it in this PR.

PR #224, opened 2026-06-08
Stokes_Constrained: in-saddle multiplier free-slip + topography (supersedes #219)

lmoresi:
## Summary

Makes the **in-saddle (single coupled solve)** constrained free-slip solver the production `uw.systems.Stokes_Constrained`, and **removes the augmented-Lagrangian outer-loop variant** (PR #219). The constraint `u·n = g` is enforced by a Lagrange multiplier carried inside the saddle point; the converged boundary multiplier is the normal traction = dynamic topography (`multiplier()` / `topography()`).

This supersedes #219 — the outer-loop solver is closer to half-formed and fragile on hard constraints (it explodes on viscosity jumps), whereas the in-saddle solver is one mesh-independent solve and is now validated against an exact analytic solution.

## What's here

- `SNES_Stokes_Constrained` is the in-saddle solver; single public export `uw.systems.Stokes_Constrained` (the outer-loop class and `_ConstraintBC` are removed — the Uzawa scheme is easy to reproduce in Python if wanted).
- **Lossless boundary-only multiplier reduction (default on)**: interior multiplier DOFs are constrained directly in the PetscSection, collapsing the `[p,h]` block DOF premium from ~3× to ~1.1× Dirichlet with no accuracy loss.
- Default `augmentation_base` raised 1e3 → 1e4 (accuracy is independent of the AL value; larger sits in the low-iteration plateau, well below roundoff).
- **Validated against the exact SolCx analytic** (`tests/test_1062_constrained_solcx.py`): free-slip via four in-saddle multipliers matches the analytic to `8.7e-6` — indistinguishable from a Dirichlet free-slip reference — with the constraint enforced to machine precision (`RMS(u·n) ≈ 1.6e-10`).
- Retargeted constrained test (box + annulus); outer-loop annulus test removed; design doc updated.

## Merge note

Please **squash-merge** — the branch was rebased onto `development` carrying #219's commits underneath, so squash collapses them into one clean commit. The net file diff is the in-saddle solver only (no outer-loop).

Depends on #223 (SolCx analytic), already merged.


lmoresi:
Addressed the Copilot review (commit 9863744):
- **Dead code**: removed `_BlockConstraintBC.interior_mask` and its construction (a P1 marker field + `evaluate`) — unused since the section-based interior-multiplier reduction; this also drops the avoidable per-constraint setup cost.
- **Tests**: switched to the public `petsc_use_pressure_nullspace` property (its setter resets cached nullspace state); fixed the stale `run` path in `test_1061`.
- **Docs**: rewrote the outdated outer-loop sections of `CONSTRAINED_FREESLIP_MULTIPLIER.md` to the shipped in-saddle formulation (no outer iteration; `r` is augmented-Lagrangian stabilisation only; `augmentation_base=1e4`; boundary-only reduction), corrected the Files list and option-table verdicts, and marked the `r`-sweep table as historical outer-loop data.

12/12 constrained tests still green.

PR #230, opened 2026-06-11
Restore FMG mesh hierarchy transparently on checkpoint reload

lmoresi:
## Summary

A mesh built with `refinement` carries a geometric-multigrid hierarchy in `dm_hierarchy`. Until now the checkpoint reload (`Mesh(file)`) rebuilt only the fine DMPlex, so a reloaded mesh had a single level and geometric **Full Multigrid** silently fell back to **GAMG** after a restart.

This makes the existing reload restore the hierarchy **transparently** — same `Mesh(file)` call, no new function or pattern, deformed coordinates unchanged, in **serial and parallel**.

## How it works

- `Mesh.write` persists the **coarsest** level in one sidecar (`<base>.hierarchy.L0.h5`) plus a `hierarchy_coarse_levels` depth in the mesh metadata. One extra file regardless of refinement depth.
- `Mesh(file)` loads the coarsest, rebuilds the intermediate coarse levels by `refine()`, distributes the fine **and** every coarse level with PETSc's **Simple** partitioner, links them, and flags the fine levels as regular refinements.
- **Why Simple co-locates:** the fine carries canonical refinement numbering (coarse cell `c` -> fine cells `c*numSubcells+r`, contiguous), so equal contiguous splits put each rank's coarse cells and their fine children on the same rank. The multigrid interpolation is rank-local and the exact **nested** operator applies — this avoids the cross-partition hang that a graph partitioner produces on independently-loaded levels.
- The working mesh stays the loaded fine, so coordinates come back exactly as a non-hierarchical reload.

## Backward compatibility

Old checkpoints (no `hierarchy_coarse_levels`) and plain (non-hierarchy) meshes write no sidecar and reload byte-for-byte as before, serial and parallel.

## Trade-off

Hierarchy meshes reload with a Simple (contiguous) partition rather than the default graph partition. Field reload is coordinate-matched (partition-agnostic), so correctness is unaffected; partition-quality tuning can come later. Plain meshes are untouched.

## Validation

| case | result |
|---|---|
| serial suite (refinement=2 -> one sidecar) | 4/4 tier_a |
| parallel np=2 | 3 levels restored, FMG `its=31` |
| parallel np=3 (uneven partition) | 3 levels restored, FMG `its=32` |
| old-format reload (serial + parallel) | 1 level, no hang |
| mesh + save/load regression | 26/26 |

## Changes

- `discretisation_mesh.py` — sidecar write (coarsest only), `.h5` reload rebuild + Simple co-location + splice.
- `petsc_discretisation.pyx` / `petsc_extras.pxi` — `petsc_dm_{set,get}_regular_refinement` wrappers (`DMPlexSetRegularRefinement` is not exposed by petsc4py).
- Tests: `tests/test_0004_checkpoint_fmg_hierarchy.py` (serial), `tests/parallel/ptest_0004_checkpoint_fmg_hierarchy.py` (np=2, np=3, registered in `mpi_runner.sh`).
- Design note: `docs/developer/design/fmg-checkpoint-hierarchy.md`.

> **Docs preview (after merge to `development`):** [Persisting the FMG mesh hierarchy across checkpoints](https://underworld3.readthedocs.io/en/development/developer/design/fmg-checkpoint-hierarchy.html) — the design note (reachable by direct link; it is a developer design doc, not in the nav toctree). ReadTheDocs builds the `development` version on merge.

## Reviewer note

Intended to be exercised by the MG testing machinery. The Simple-partition co-location is the load-bearing trick for parallel; the design note records the dead ends (reconstruct-from-label, `coarsen()`, regularRefinement-flag-alone) and why this path was chosen.

### DIFF
--- a/docs/developer/design/fmg-checkpoint-hierarchy.md
+++ b/docs/developer/design/fmg-checkpoint-hierarchy.md
@@ -1,0 +1,117 @@
+---
+title: "Persisting the FMG mesh hierarchy across checkpoints"
+---
+
+# Persisting the geometric-multigrid hierarchy across checkpoints
+
+## Problem
+
+A mesh built with `refinement=N` carries a geometric refinement hierarchy in
+`mesh.dm_hierarchy`, which the Stokes/scalar/vector solvers use for geometric
+**Full Multigrid (FMG)** — the anisotropy-robust preconditioner of choice on
+adapted meshes. But the checkpoint load path (`_from_plexh5`) reconstructs only
+the single saved DMPlex, so a **reloaded mesh has `dm_hierarchy = [dm]`** (one
+level) and FMG silently falls back to GAMG after a restart. This note records the
+design that restores the hierarchy on reload, and the experiments that shaped it.
+
+## Why store the coarse levels (not reconstruct, not refine-from-label)
+
+Several approaches were prototyped and rejected:
+
+- **Reconstruct the coarse mesh from a per-node "level" label.** The coarse
+  *topology* is recoverable bit-exact from labelled nodes (the all-midpoint
+  "central" fine cells map 1:1 to coarse cells). **But** a coarse DMPlex built
+  from scratch (`createFromCellList`) does not reproduce the original's internal
+  cone-orientation / edge ordering, and PETSc's nested multigrid interpolator
+  needs the **canonical `refine()` numbering**. Splicing a reconstructed coarse
+  in throws `PETSc has generated inconsistent data` (err 77).
+- **Store only the coarsest level + `refine()` back up on reload.** Works for the
+  hierarchy, but then the rebuilt fine has to be reconciled (numbering + field
+  data) with the saved deformed mesh — fragile.
+
+The winning insight: a **loaded** coarse DM preserves the canonical numbering
+(`topologyLoad` is faithful), so it can be `setCoarseDM`-linked directly under the
+working fine and FMG just works — **no refine, no node-moving, no reconstruction**.
+This is exactly the live `clone_dm_hierarchy` pattern with *load* swapped for
+*clone*. Validated: `refine(stored L0) == saved fine` bit-exact in 2D and 3D, and a
+reloaded hierarchy drives `pc_type=mg` to convergence.
+
+## On-disk format: a single coarsest sidecar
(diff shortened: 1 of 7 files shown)


lmoresi:
Validated #230 end-to-end against an adaptive-convection **restart** (the case that motivated it) — it does what it says: a checkpoint reload restores `dm_hierarchy` and FMG runs on the restarted mesh instead of silently falling back to GAMG.

**Confirmed working**
- Sidecar reload restores `len(mesh.dm_hierarchy) == 3` — in isolation at **np=1, np=2, and np=5** (minimal repro: build with `refinement=2`, write, reload, check levels).
- End-to-end FMG restart at **np=1 (serial) and np=2**: resume → hierarchy restored → FMG runs flat (`vel_inner_its` ~5–6, `levels=3`, outer Schur KSP = 1) → completes cleanly. (Without #230 the same restart silently drops to GAMG at ~75–130 inner iters — i.e. this PR is exactly what makes a restarted run keep its mesh-independent solve.)

**One finding at np=5 — likely worth a look**
The resumed run dies with **signal 9 (macOS jetsam / memory pressure)** during the *initial Stokes solve* on the restored-hierarchy mesh — *after* the reload and FMG setup both succeed. It is **not a correctness bug**: np=2 completes the identical solve fine. It reads as a **memory-footprint** difference between the *freshly-built* hierarchy (refinement at construction — np=5 solves without trouble) and the *reloaded* one (OOMs at np=5). Candidates: the restored coarse-level partition, or `mg_coarse_pc_type=redundant` replicating a larger coarse operator per rank than the freshly-built case. Happy to share the repro or dig further.

**Dependency for the parallel path**
Before reaching that solve, the np>1 restart also hits an empty-partition `reshape size 0` deadlock in `swarm._pack_array_to_data_format` (a rank with zero particles after the reload/redistribution) — which is exactly **#221**. So a parallel FMG restart wants **#221 landed alongside #230**.

Config: Annulus, Ra 1e7, Δη 1e3, res 32, R=8, mmpde + tangent-slip. (The harness's own `--resume` bugs — a 10-vs-16-column history row and a corrupt-`history.npz` crash — were unrelated and fixed separately.)

PR #228, opened 2026-06-09
Anisotropic mesh-adaptation movers + mesh-owned boundary tangent-slip + MPI robustness fixes

lmoresi:
## Architectural Review Submission

**Review Type**: Code Implementation / Performance  
**Priority**: HIGH  
**Review Period**: 2026-06

---

### 📄 Review Document
- **File**: `docs/reviews/2026-06/ANISOTROPIC-MOVER-ADAPT-TRANSFER-REVIEW.md`
- **Tracking Index**: `docs/reviews/2026-06/REVIEW-TRACKING-INDEX.md`

### 📋 Executive Summary
This PR delivers anisotropic mesh-adaptation updates (default MMPDE mover, metric and slip-path refactors) and hardens parallel correctness in adapt/remap workflows. In addition to the original transfer clamp fix, follow-up corrections ensure deterministic MPI behavior by clearing reused global vectors before additive assembly and by using globally weighted edge-length reductions for characteristic mesh size (`width='auto'` and `_mesh_h0`). Together these changes remove partition-dependent behavior while preserving existing physics outcomes.

### 🎯 Scope
**Systems Affected**:
- [x] Core solvers (Stokes, Poisson, Advection-Diffusion)
- [x] Variable system (Mesh/Swarm variables, arrays)
- [ ] Units & non-dimensionalization
- [x] Parallel operations (MPI, collective operations)
- [x] Testing infrastructure
- [x] Documentation & examples
- [x] Other: Meshing/mover boundary-slip architecture

**Files Changed**:
- New files: multiple tests/docs
- Modified files: meshing/remesh/smoothing/surfaces and related tests
- Total LOC: substantial net refactor + targeted bug fixes

### ✅ Testing Status
- [x] All tests passing: `pixi run underworld-test` (targeted and tier-A validation reported for changed areas)
- [x] New tests added: mover/slip/fault-metric coverage expanded
- [ ] Test coverage: N/A
- [x] Regression tests validated
- [x] Performance benchmarks run (if applicable)

**Test Results**:
```bash
# Command used
pixi run -e default pytest tests/test_0762* tests/test_0763* tests/test_0855* -v
# plus tier-A / level_1 and parallel adapt validation runs (np=5)

# Results summary
Targeted slip/mover tests green; tier-A (level_1) 200 passed; parallel adapt runs clean.
```

### 📊 Key Metrics & Impact
[...]

PR #231, opened 2026-06-11
Add `preconditioner` property: automatic geometric FMG on refinement meshes

lmoresi:
## Summary

UW3 Stokes / scalar / vector solvers defaulted to **GAMG** (algebraic multigrid), which is sensitive to mesh anisotropy — exactly what mesh adaptation produces. Geometric **Full Multigrid (FMG)** is anisotropy-robust but required hand-setting ~8 cryptic `fieldsplit_velocity_*` PETSc options.

This adds a `preconditioner` property so building a mesh with `refinement=N` and solving uses FMG **automatically**, with no PETSc options to set.

## API

```python
stokes.preconditioner = "auto"   # default: FMG if the mesh has a refinement hierarchy, else GAMG
stokes.preconditioner = "fmg"    # force geometric multigrid
stokes.preconditioner = "gamg"   # force algebraic multigrid (historical default)
```

On the SNES base class, inherited by `SNES_Stokes`, `SNES_Scalar`, `SNES_Vector`. The choice is resolved at build time (re-evaluated after a remesh), keyed by a per-class option prefix (`""` for scalar/vector, `"fieldsplit_velocity_"` for Stokes).

## Conservative `auto`

`"auto"` only *adds* geometric MG on top of an untouched framework default — it never rewrites pc options a solver or user set directly (e.g. the tuned GAMG in the OT / mesh-smoother, or an explicit `pc_type=lu`). This was caught by regression: the first cut clobbered those and broke `test_0760_mesh_ot_adapt` and `test_snes_vector_asymmetric_jacobian`; the conservative resolution fixes both.

## Why it matters (benchmark)

Annulus Ra=1e7, dEta=1e3, res32, R=8, mode-1, np=5, MMPDE mover, 50 steps of a continuously deforming mesh — inner velocity-block KSP iterations:

| velocity-block KSP iters | FMG | GAMG |
|---|---|---|
| range over 50 steps | 3-6 (median 5) | 64-131 (median ~114, ~23x) |

FMG stays mesh-independent at ~5; GAMG is volatile but does **not** cliff at R=8. Wall-clock gap is only ~1.8x (the cold-start Stokes solve after each adapt, common to both engines, dominates the time), so the value is **predictability/mesh-independence, not raw speed**. The benchmark figure + per-step CSV are now in `docs/advanced/multigrid-preconditioning.md`. Hierarchy survives the coordinate-deforming mover; a true remesh collapses it and `"auto"` reverts to GAMG.

## Validation

- `tests/test_1014_stokes_multigrid.py` — 8 tests, tier_a (FMG engages on refinement meshes, gamg override, auto-fallback without hierarchy, fmg-without-hierarchy warns + falls back, scalar Poisson, invalid value raises).
- `test_1010_stokesCart.py`: 6/6. `level_1 and tier_a` sweep: 177 passed (the OT-adapt and asymmetric-jacobian regressions from the first cut are resolved).

## Docs

> **Docs preview (after merge to `development`):** [Multigrid Preconditioning (FMG vs GAMG)](https://underworld3.readthedocs.io/en/development/advanced/multigrid-preconditioning.html) — the how-to with the benchmark figure. ReadTheDocs builds the `development` version on merge.

`docs/advanced/multigrid-preconditioning.md` (user how-to with the benchmark), `solver-strategies-catalogue.md` GMG section marked landed.

## Relationship to PR #230

Complementary and independent. #230 makes the FMG hierarchy survive a checkpoint reload; this makes solvers *use* it automatically. Either can merge first.


lmoresi:
## Hardening: FMG broke on adapted meshes in parallel — root cause + fix (5aec2b2)

A parallel FMG benchmarking session found that geometric FMG diverged on a mover-adapted (mmpde) mesh at np>1, while it stayed mesh-independent on the anisotropic-mover branch (#228) that lacks this property. Root cause was three issues in the `preconditioner` property's auto/FMG path — all now fixed with regression tests (test_1014 now 11/11).

**1. Override clobber on rebuild.** `_apply_preconditioner_options()` runs on every `_build` (so `auto` can re-resolve after a remesh). The "respect an explicit user `pc_type`" back-off only held on the *first* build: once the user's `mg` was adopted as the managed value, a later rebuild (which a mover deform triggers) could no longer tell "user set mg" from "we set mg", and overwrote the user's tuned smoother / coarse-solver with the framework bundle. So step 1 used the good options and worked; step 2+ (post-adapt) got clobbered. Fixed with a durable `_pc_user_override` latch.

**2. Non-parallel-safe defaults.** The bundle shipped a bare serial `mg_coarse_pc_type=lu`, which cannot factor a distributed coarse matrix and fails at np>1 (`DIVERGED_LINEAR_SOLVE` after 0 iterations), plus `chebyshev` level smoothing (eigen-estimate-fragile on the indefinite/variable-viscosity velocity block). Changed to `redundant`+`lu` (np-safe, identical to lu in serial) and `richardson`+`sor` (the benchmark-validated mesh-independent choice). The property's own defaults now converge unaided at np=1 and np>1.

**3. Missing-Galerkin footgun.** UW3's geometric MG *requires* Galerkin RAP coarse operators — it installs no operator callbacks on the coarse DMs, so PETSc cannot re-discretise there. Selecting `pc_type=mg` without `pc_mg_galerkin` failed cryptically as PETSc **error 73** in serial or **`DMCoarsen -> DMAdaptMetric -> ParMmg`** (3D-only) in parallel — note the ParMmg is the MG trying to *generate* coarse grids, not the mover adapting topology. Added `_enforce_galerkin_for_geometric_mg()` which forces `pc_mg_galerkin=both` (with a warning) on any managed block using geometric MG, so this can't be tripped silently. (A bare `=None` flag already engages RAP and is left alone; only genuinely-unset or `none` is overridden.)

PR #237, opened 2026-06-14
Land manifold-mesh PDE support + Mesh.extract_surface (supersedes #202)

lmoresi:
## Summary

Lands the manifold-mesh PDE support and `extract_surface` work from #202 onto the
**current** `development`, with the reconciliation fixes that integration
surfaced. #202's branch is on a very old base (fork point ~#146, 194 dev commits
behind); this branch is that work *merged onto current development* and made
green, so it supersedes #202 as the mergeable landing.

Motivating consumer: the free-surface MMPDE-mover work needs a codim-1 surface
submesh (`mesh.extract_surface`) for the surface-following smoother. Landing the
interface first (per the branching discipline) lets that work build on a
released API.

## What's here

**From #202 (manifold / cdim stack):** `SphericalManifold`, dim≠cdim plumbing
through the solver/JIT/constitutive/swarm stack, parallel point-eval + scalar PDE
solving on codim-1 meshes, `SNES_MultiComponent_Projection`, and the
`extract_surface` prototype + manifold-PDE example suite. All cdim changes are
gated behind `dim != cdim`; volume (2D/3D) paths are unchanged.

**Reconciliation fixes (this branch):**
- **Constant-nullspace de-duplication.** Both `development` and #202 had
  independently added a scalar constant-nullspace feature. The merge left two
  `_attach_constant_nullspace` definitions; Python's last-wins took #202's
  guard-less one, and development's *unconditional* caller then attached a
  constant nullspace to **every** scalar solve — making Dirichlet problems
  singular (`DIVERGED_LINE_SEARCH`; `test_1004` Darcy). Kept development's
  guarded implementation, dropped the duplicate, and made
  `petsc_use_constant_nullspace` a back-compat alias to `constant_nullspace`
  (the manifold-PDE examples still work).
- **`Mesh.extract_surface`.** Promoted the `docs/examples` prototype to a real
  `Mesh` method (third submesh flavour alongside `extract_region`): codim-1
  surface mesh (`dim = parent.dim-1`, `cdim` preserved) via
  `DMPlexCreateSubmesh` + `DMPlexFilter`, composed subpoint maps, standard
  lineage, KDTree-on-`_coords` vertex map. Serial contract tests in
  `tests/test_0772_surface_extract.py` (3D shell → 2-manifold; 2D annulus →
  1-manifold loop).
- **`ddt.py` merge resolution** — kept development's `RemeshPolicy.CARRY` setup
[...]

PR #238, opened 2026-06-14
fix(meshing): clear dm_plex_gmsh options after each import (spacedim leak)

lmoresi:
## Summary

A pre-existing core-meshing bug: the `dm_plex_gmsh_*` options are global,
import-time scratch on the PETSc options database, but they were **not cleared
after a gmsh import**. A manifold-surface generator sets
`dm_plex_gmsh_spacedim = 2` and, if it raises before cleanup (e.g.
`SegmentedSphericalSurface2D` errors in `set_periodicity` *after* the import),
that option **leaks** — and the next gmsh import (say a `SphericalShell`) is read
with `spacedim=2`, producing a 2-D mesh in a 3-D `cdim`.

Symptom (found via cross-test pollution): after a 2-D manifold generation, a
fresh `SphericalShell` loads as 2-D and crashes with
`cannot reshape array of size 856 into shape (3)` — and the corrupt 2-D mesh is
even written to the shell's on-disk cache, so it persists.

## Fix

- **A (root):** clear the whole `dm_plex_gmsh_*` namespace in a `finally` after
  every import, centrally in `_from_gmsh`. These options are meaningful only for
  the single import that follows, so a value set for one import can no longer
  leak into the next — on success or failure.
- **B (defence-in-depth):** in the `Mesh` constructor, validate the DM
  coordinate array is divisible by `cdim` before reshaping; if not, raise a
  clear, actionable error (stale/corrupt cache → delete `.meshes/*.msh(.h5)`)
  instead of an opaque numpy reshape failure.

## Validation

Verified on the integration branch where it was found: a failed manifold
construct followed by a `SphericalShell` now yields a correct 3-D mesh; full
`level_1 + tier_a` is green; no false positives on valid meshes.

Cherry-picked from validated work on the manifold/extract_surface integration
branch (PR #237), where it gates the surface-extraction tests' CI.

### DIFF
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -45,6 +45,35 @@ def _temporary_petsc_option(key, value):
             opts[key] = old_value
 
 
+# The ``dm_plex_gmsh_*`` options configure how PETSc imports a Gmsh ``.msh``
+# file (coordinate space dimension, label handling, …). They are GLOBAL on the
+# PETSc options database and are *import-time scratch* — meaningful only for the
+# single import that follows. A mesh constructor that sets one (notably the
+# manifold-surface generators set ``dm_plex_gmsh_spacedim = 2``) and then raises
+# before cleanup leaks it, and the stale option silently corrupts the NEXT gmsh
+# import (e.g. a subsequent SphericalShell is read with spacedim=2 → a 2-D mesh
+# in a 3-D cdim → a "cannot reshape" crash, or a corrupt cache). ``_from_gmsh``
+# therefore clears the whole namespace after every import (see below).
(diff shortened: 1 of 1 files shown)


lmoresi:
## Review — approved ✅

Clean, well-documented single-file fix. Verified:
- `_clear_gmsh_import_options()` runs in a `finally` *after* `createFromFile`, so the import still sees `dm_plex_gmsh_spacedim`; only the leak into the *next* import is prevented. Idempotent, runs on all ranks (safe).
- Coordinate-dimension validation is guarded with `if self.cdim and ...` (no div-by-zero) and turns an opaque numpy "cannot reshape" into an actionable message naming the stale cache.

No behavioural change to valid imports. Merging for the release cleanup.

PR #240, opened 2026-06-15
Stokes_Constrained: unlock parallel (remove over-conservative serial guard)

lmoresi:
## What

Removes the serial-only guard on `SNES_Stokes_Constrained` (the in-saddle Lagrange-multiplier free-slip solver). It is parallel-safe; the guard was over-conservative.

## Why it was guarded, and why that was unnecessary

The multiplier is carried as a full-domain field whose interior DOFs are pinned away by `_constrain_interior_multipliers_in_section`. That reduction is **rank-local section surgery** — it uses the *distributed* boundary-label `IS` (`getStratumIS`) and iterates the *local* chart, constraining DOFs in the local `PetscSection` per rank. So the assembled global system is partition-independent; there is no single-rank enumeration to MPI-decompose.

## Validation

Bit-identical serial-vs-parallel (np=1/2/4), isotropic **and** transverse-isotropic, on the enclosed free-slip annulus:

| metric | np1 | np2 | np4 |
|---|---|---|---|
| velocity `L2(v)` | 0.39245136 | = | = |
| free-slip `∫(v·n)²` | 1.3e-9 | = | = |
| gauge-fixed topography `∫(h−h̄)²` | 37.0335397 | 37.0335204 | 37.0335210 |

`BdIntegral(fn=1)` (boundary length) is also bit-identical, confirming the surface integrator is parallel-safe. New regression test: `tests/parallel/test_1063_constrained_freeslip_parallel.py` (`mpi(min_size=2)`).

## The gauge constant

On an enclosed problem the raw multiplier `h` is determined only up to the `[p,λ]` gauge constant, and the solver lands on a *partition-dependent representative* of it (the velocity and the *deviation* of `h` are unaffected). `topography()` gains a `reference="mean"` option to subtract the boundary mean for a reproducible, gauge-fixed readout. The **default is unchanged** (`reference=None`, raw multiplier) — correct for problems with no gauge freedom (e.g. an open boundary), where the mean of `h` is the physical mean traction.

## Changes

- `src/underworld3/systems/solvers.py`: remove the serial guard; add `topography(reference=)`.
- `tests/parallel/test_1063_constrained_freeslip_parallel.py`: new np≥2 regression.

## Tests

- Serial constrained suite (`test_1061`, `test_1062`): 16 passed (no regression).
- New parallel test: passes at np2 and np4.

### DIFF
--- a/src/underworld3/systems/solvers.py
+++ b/src/underworld3/systems/solvers.py
@@ -2149,12 +2149,18 @@ class SNES_Stokes_Constrained(SNES_Stokes):
         h : MeshVariable
             The scalar multiplier field.
         """
-        # Serial only for now: the boundary mask (later milestones) is not
-        # MPI-decomposed.
-        if uw.mpi.size > 1:
-            raise NotImplementedError(
-                "SNES_Stokes_Constrained is serial-only for now."
-            )
+        # Parallel-safe: the interior-multiplier reduction
+        # (_constrain_interior_multipliers_in_section) is rank-local section
+        # surgery (it uses the distributed boundary label IS and iterates the
+        # local chart), so the global system — and hence the velocity solve and
+        # the gauge-invariant boundary traction — are partition-independent.
+        # Validated bit-identical at np=1/2/4 (velocity L2 and mean-stripped
+        # boundary topography) in
+        # tests/parallel/test_1063_constrained_freeslip_parallel.py.
+        # NOTE: on enclosed problems the raw multiplier h carries the [p,λ] gauge
+        # constant, of which the solver lands on a partition-dependent
+        # representative — strip its boundary mean for a reproducible topography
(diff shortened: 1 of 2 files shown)


lmoresi:
Addressed the Copilot review in a510445:
- `topography(reference="mean")` docstring now notes it runs **collective** `BdIntegral` reductions (call on all ranks); `reference=None` stays a pure symbolic accessor.
- The parallel test now computes the gauge-fixed topography via `solver.topography("Upper", reference="mean")`, so that new code path is exercised (golden values unchanged; np2/np4 still pass).
- Reworded "bit-identical" → "tight tolerance" in the test docstring: the residual serial-vs-parallel difference is the parallel reduction order, not the solver, so the `np.isclose` tolerances are intentional (velocity 1e-9, topography 1e-6).

PR #241, opened 2026-06-16
Surface.influence_function: respect finite edges (no bleed past the end)

lmoresi:
## Problem

`Surface.influence_function` built its influence on `Abs(self.distance.sym[0])` — the `Abs` of the **signed** distance field. The signed distance changes sign across the surface and its zero-contour follows the surface's **infinite** line/plane, so it extends *past* a finite edge. Any non-nodal evaluation interpolates the signed field through ~0 along that extension, and `Abs()` of that leaves a **spurious high-influence streak well beyond the end of the surface**.

Concretely: a dipping weak-fault zone (finite segment) lit up a weak band along the fault's line-extension, far beyond the fault tip — contaminating the viscosity field where the fault does not exist.

The per-node distances are already edge-clamped (the segment/surface distance falls back to the nearest endpoint), so the magnitude is correct *at the nodes* — the artifact is purely the **interpolation of the sign-flipping field**, which `Abs` cannot undo (abs of an interpolated zero-crossing is still ~0).

## Fix

Store a companion **unsigned, edge-clamped** distance field `Surface.abs_distance` (populated as `|signed distances|` in `_compute_distance_field`) and build `influence_function` on it. The unsigned field is `>= 0` everywhere and, beyond an edge, is simply the radial distance to the endpoint, so its interpolant never crosses zero there — the influence decays correctly past a finite edge.

Signed `distance` is unchanged and still available for sided (`d > 0`) properties.

## Test

`tests/test_0851_surface_influence_edge.py` embeds a finite segment and checks the influence decays beyond the tip. Along the line extension (>= 3·width past the tip) the **old** basis spikes to influence ~1.0 on the mesh edges crossing the extension; the **fix** keeps it ~0. A second test confirms `abs_distance` tracks the geometric edge-clamped distance between nodes.

Existing surface tests (`test_0850_surfaces.py`, `test_0762_fault_metric_tensor.py`) pass unchanged (82 passed, 1 pre-existing xfail).

### DIFF
--- a/src/underworld3/meshing/surfaces.py
+++ b/src/underworld3/meshing/surfaces.py
@@ -704,6 +704,10 @@ class Surface:
 
         # Level 3: Cached proxy MeshVariable for distance
         self._distance_var: Optional[uw.discretisation.MeshVariable] = None
+        # Companion UNSIGNED (edge-clamped) distance field. influence_function()
+        # interpolates THIS rather than Abs() of the signed field — see
+        # _compute_distance_field for why the signed field is unsafe to abs.
+        self._abs_distance_var: Optional[uw.discretisation.MeshVariable] = None
 
         # Dimension (2 or 3) - detected from mesh or control points
         self._dim = None
@@ -759,6 +763,7 @@ class Surface:
             # If distance var exists, it needs to be recreated with new symbol
             if self._distance_var is not None:
                 self._distance_var = None
+                self._abs_distance_var = None
                 self._distance_stale = True
 
     def _dimensionalise_coords(self, coords: np.ndarray) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


lmoresi:
## Review — approved ✅

Correct, complete, and well-tested. Verified:
- **Root cause is real**: `Surface` stored the *signed* distance and `influence_function` took `Abs()` of its **interpolant**. The signed field's zero-contour follows the surface's infinite line/plane past a finite edge, so non-nodal evaluation interpolates through ~0 there and `Abs()` lights up a spurious weak zone beyond the tip.
- **Fix is consistent with existing design**: `SurfaceCollection.influence_function` already stored an *unsigned* nodal distance (`compute_distance_field`, `np.abs(...)` min) and used it directly without `Abs()`. This PR brings the single-`Surface` path in line — both now interpolate a `>=0` edge-clamped nodal field whose interpolant never crosses zero past the edge.
- **No lifecycle hazard**: `_distance_var` is created inside `_compute_distance_field` (not the `distance` property), so accessing `abs_distance` before `distance` creates both fields — no None-deref. `_abs_distance_var` is reset alongside `_distance_var` in the `symbol` setter.
- **Regression test** `test_0851` probes a band along the segment's line extension at non-nodal points — exactly where the artifact appeared.

Merging for the release cleanup.

PR #243, opened 2026-06-16
fix(darcy): correct sign conventions for Darcy flux and velocity

### LINKED ISSUE #214: [BUG] - Sign mismatch between docstrings and implementation in SNES_Poisson and SNES_Darcy
These are minor bugs, but may be confusing to users due to mismatches between documentation and code behaviour. 

**Bug 1 — SNES_Poisson docstring: wrong sign**

The class docstring states:

∇·[κ∇u] = f

But the code comment below (line 218 of solvers.py) says:

```python
# For the Poisson equation: -∇·(k∇u) = f
```

Numerical tests also confirm that this code comment is correct.

**Bug 2 — SNES_Darcy / DarcyFlowModel: v_Field sign**

The SNES_Darcy docstring states:

v = κ∇h − s

The DarcyFlowModel docstring states:

q_i = κ_ij (∂p/∂x_j − s_j)

But in solvers.py, the projected velocity is set as:

self._v_projector.uw_function = -self.darcy_flux

so the actual v_Field = −κ·(∇h − s) has opposite sign to both docstrings.

- What version code?
Latest development branch

jcgraciosa:
## Summary

Fixes sign-convention inconsistencies in the Darcy and Poisson solvers.
Changes are in `solvers.py` and `constitutive_models.py`:

1. **SNES_Poisson docstrings** — restored missing minus sign: `-∇·[κ∇u] = f`
2. **v_Field** — removed double negation; now correctly stores the physical
   Darcy velocity `q = -κ(∇h - s)` (inflow positive)
3. **DarcyFlowModel.flux** — now returns physical velocity `q = -κ(∇h - s)`;
   permeability should be set as a positive value (matching literature convention)
4. **F1 Template** — uses `darcy_flux = q` consistent with uw3's natural BC
   convention, where `add_natural_bc(g, boundary)` specifies inward Darcy flux
5. **TransientDarcy DFDt** — negated to compensate for Fix 3, restoring the
   same net `K(∇h−s)` that DFDt tracked before any fixes

Validated against the Henry (1964) saltwater intrusion benchmark and the
Badon-Ghijben–Herzberg freshwater lens benchmark — both produce physically
correct results with the correct freshwater inflow direction and wedge geometry.

## Tests

- `test_1004_DarcyCartesian`
- `test_1005_TransientDarcyCartesian` 
- `test_0820_template_parameter_propagation`

## Follow-up

The natural BC sign convention for `SNES_TransientDarcy` (when `f ≠ 0` or
when natural BCs are used on transient problems) will be addressed in a
separate PR.

Closes #214 

---

### DIFF
--- a/src/underworld3/constitutive_models.py
+++ b/src/underworld3/constitutive_models.py
@@ -2410,7 +2410,7 @@ class DarcyFlowModel(Constitutive_Model):
 
         ddu = self.grad_u - self.Parameters.s.sym
 
-        return self._q(ddu)
+        return -self._q(ddu)
 
 
 class TransverseIsotropicFlowModel(ViscousFlowModel):
--- a/src/underworld3/systems/solvers.py
+++ b/src/underworld3/systems/solvers.py
@@ -164,7 +164,7 @@ class SNES_Poisson(SNES_Scalar):
 
     .. math::
 
-        \nabla \cdot \left[ \boldsymbol{\kappa} \nabla u \right] = f
+        -\nabla \cdot \left[ \boldsymbol{\kappa} \nabla u \right] = f
 
     where :math:`\mathbf{F} = \boldsymbol{\kappa} \nabla u` relates the flux to
     gradients in the unknown :math:`u`.
@@ -257,7 +257,7 @@ class SNES_Poisson(SNES_Scalar):
         The source term :math:`f` appears on the right-hand side:
 
         .. math::
-            \nabla \cdot (\kappa \nabla u) = f
+            -\nabla \cdot (\kappa \nabla u) = f
 
         Returns
         -------
@@ -434,7 +434,7 @@ class SNES_Darcy(SNES_Scalar):
 
     # =========================================================================
     # PETSc Residual Templates
-    # For Darcy flow: -∇·(K∇p) = f, with velocity v = -K∇p
+    # For Darcy flow: -∇·(κ(∇h - s)) = f, with velocity q = -κ(∇h - s)
     # =========================================================================
 
     F0 = Template(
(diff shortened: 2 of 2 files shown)


lmoresi:
## Review — sign-convention change needs rework before merge

Thanks for chasing this down — there **is** a real bug here, but the current fix trades it for a different (untested) one. Recommend reworking before merge.

### What the PR does
It flips `DarcyFlowModel.flux` from `+κ(∇h−s)` to `−κ(∇h−s)` and adjusts the call sites. Working through the PETSc weak residual `∫ (f0·v + F1·∇v) = 0` → strong form `f0 − ∇·F1 = 0`, with `f0 = −f`:

| quantity | `development` | this PR |
|---|---|---|
| `F1` (assembly flux) | `+κ(∇h−s)` | `−κ(∇h−s)` |
| **pressure PDE actually solved** | `−∇·(κ(∇h−s)) = f` ✓ (all `f`) | `−∇·(κ(∇h−s)) = −f` ✗ (only correct for `f=0`) |
| steady velocity (`SNES_Darcy.solve`) | `−darcy_flux = q` ✓ | `darcy_flux = q` ✓ |
| transient velocity (`SNES_TransientDarcy.solve`) | `+darcy_flux = −q` ✗ ← **the actual bug** | `darcy_flux = q` ✓ |

So the genuine defect was only the **transient velocity projection sign** (steady used `−self.darcy_flux`, transient used `self.darcy_flux`). Flipping `flux` to fix that also flips `F1`, which silently inverts the source term for `f≠0`. The PR's own `F1` docstring acknowledges this ("for f≠0 a sign fix is needed (tracked separately)").

CI stays green because the steady Darcy regression test (`test_1004` line 96) sets `darcy.f = 0.0`, so the broken path has **no coverage**.

### Suggested minimal fix
Keep assembly `F1 = κ(∇h−s)` (correct for all `f`); separate the **physical velocity** from the **assembly flux**:

1. In `SNES_TransientDarcy.solve`, project `-self.darcy_flux` (match the steady solver) — this is the one-line behavioural fix.
2. Fix the docstrings to the physically correct Darcy velocity `q = −κ(∇h−s)` (issue #214 Bug 2 is a *docstring* error — the steady code was already right).
3. Keep the SNES_Poisson docstring sign corrections from this PR — those are correct (#214 Bug 1).

A regression test with `f ≠ 0` (e.g. a manufactured solution) would lock the assembly sign and is worth adding alongside.

Happy to push these changes if useful. Leaving #214 open until the rework lands.

lmoresi:
Superseded by #255 (merged to development). #255 takes the minimal route — fixes only the genuine bug (the transient-velocity projection sign) and corrects the docstrings, leaving the assembly flux `F1 = κ(∇h−s)` intact (so the `f≠0` source response is unchanged/correct). This PR's approach of flipping the constitutive `flux` introduced an untested `f≠0` source-sign regression (see review above). Thanks for surfacing the issue — closing in favour of #255.

PR #255, opened 2026-06-19
fix(darcy): correct transient velocity sign + sign-convention docstrings (#214)

### LINKED ISSUE #214: [BUG] - Sign mismatch between docstrings and implementation in SNES_Poisson and SNES_Darcy

lmoresi:
## Summary
Resolves the Darcy/Poisson sign-convention issues in #214 with a **minimal** fix (no change to the assembly flux), superseding #243.

### The real bug (behavioural)
`SNES_TransientDarcy.solve` projected `+darcy_flux` for the velocity field. `darcy_flux` is the **assembly flux** `F = κ(∇h−s)`; the physical Darcy velocity is `v = −F = −κ(∇h−s)`. Steady `SNES_Darcy` already used `−darcy_flux`, so **transient runs reported a sign-flipped velocity**. Fixed to `−darcy_flux` to match steady.

### Docstrings (#214)
- **SNES_Poisson** (Bug 1): wrote `+∇·[κ∇u]=f` but the solver assembles `−∇·[κ∇u]=f` → corrected both instances.
- **SNES_Darcy / DarcyFlowModel** (Bug 2): stated `v = +κ(∇h−s)`; the code correctly returns `v = −κ(∇h−s)`. Clarified that the constitutive `flux` property is the assembly flux `F` and the physical velocity is `−F`.

### What is NOT changed
The assembly flux `F1 = κ(∇h−s)` is untouched — it already solves `−∇·(κ(∇h−s)) = f` correctly for **all** `f`. (#243 instead flipped the constitutive flux, which fixed the velocity sign but introduced an *untested* `f≠0` source-sign regression — see my review there.)

## Tests (`tests/test_1004b_darcy_sign_regression.py`)
- `v·∇h < 0` (flow down-gradient) for **steady** and **transient** — the transient assertion **fails on the pre-fix `+darcy_flux`** (verified by temporary revert: ∫v·∇h = +0.25).
- `f≠0` **manufactured-solution** test locking the source/assembly sign (existing Darcy tests all use `f=0` and would not catch a flux-sign flip).

Local (amr-dev): test_1004b 3/3, test_1004 3/3, test_1005 1/1 — all green.

Closes #214.

### DIFF
--- a/src/underworld3/constitutive_models.py
+++ b/src/underworld3/constitutive_models.py
@@ -2276,15 +2276,21 @@ class DarcyFlowModel(Constitutive_Model):
     r"""
     Darcy flow constitutive model for porous media flow.
 
-    Relates the Darcy flux to pressure gradients and body forces:
+    The ``flux`` property returns the assembly flux
+    :math:`\mathbf{F}` that enters the PDE as :math:`-\nabla\cdot\mathbf{F} = f`:
 
     .. math::
 
-        q_{i} = \kappa_{ij} \left( \frac{\partial p}{\partial x_j} - s_j \right)
+        F_{i} = \kappa_{ij} \left( \frac{\partial p}{\partial x_j} - s_j \right)
 
     where :math:`\kappa` is the permeability (or hydraulic conductivity),
     :math:`p` is the pressure (or hydraulic head), and :math:`s` is the
-    body force term (e.g., gravity: :math:`s = \rho g`).
+    body force term (e.g., gravity: :math:`s = \rho g`). The **physical Darcy
+    velocity** is minus this (flow runs *down* the head gradient):
+
+    .. math::
+
+        q_{i} = -F_{i} = -\kappa_{ij} \left( \frac{\partial p}{\partial x_j} - s_j \right)
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


lmoresi:
Self-review: minimal sign fix, assembly flux untouched. Transient-velocity regression test verified to fail on the pre-fix code (∫v·∇h flips +0.25). test_1004b 3/3 + existing Darcy 4/4 green in amr-dev. Merging for the release.

PR #227, opened 2026-06-09
Fix units setup in vector projection Poisson test

ss2098:
## Summary

This PR fixes `test_vector_projection_after_poisson_with_units()` by setting model reference quantities before creating meshes and `MeshVariable` objects with physical units.

The test previously failed in strict units mode with:

```text
ValueError: Strict units mode: Cannot create variable 'T' with units='kelvin' when model has no reference quantities.
```

## What changed

Added reference quantity setup at the start of `test_vector_projection_after_poisson_with_units()`:

```python
uw.reset_default_model()
model = uw.get_default_model()
model.set_reference_quantities(
    domain_depth=uw.quantity(1.0, "m"),
    plate_velocity=uw.quantity(1.0, "m/s"),
    mantle_viscosity=uw.quantity(1.0, "Pa*s"),
    temperature_difference=uw.quantity(1.0, "K"),
)
```

## Rationale

Strict units mode requires model reference quantities to be configured before creating variables with units. This updates the test setup to match the current units-system behavior. The Poisson solve and vector projection logic are unchanged.

## Testing

Ran:

```bash
pixi run python tests/test_1450_poisson_vector_projection.py
```

Output:

```text
Structured box element resolution 16 16
✓ test_vector_projection_after_poisson passed
✓ Units system active with automatic non-dimensionalization
Structured box element resolution 16 16
✓ test_vector_projection_after_poisson_with_units passed
Structured box element resolution 16 16
✓ test_multiple_solvers_sequence passed

✓ All tests passed!
```

PR #257, opened 2026-06-19
Tidy docstring placement in vector projection Poisson test

ss2098:
This follow-up tidies the docstring placement in test_vector_projection_after_poisson_with_units() so that the docstring remains the first statement in the function.

No solver or units-system logic is changed.

Tested with:

```bash
pixi run python -m pytest tests/test_1450_poisson_vector_projection.py -q
```

Result:

```text
3 passed in 8.72s
```


lmoresi (approved):
Thanks @ss2098 — this is exactly the docstring-placement tidy I flagged on #227. Docstring is now the function's first statement again. Green CI. Merging.

Issue #239, opened 2026-06-15
Nitsche free-slip BC: inconsistent Jacobian for anisotropic (transverse-isotropic) constitutive models


lmoresi:
### Summary

`SNES_Stokes(_SaddlePt).add_nitsche_bc` (and the `SNES_Vector` version) produces a **boundary Jacobian that is inconsistent with its residual when the constitutive model is anisotropic** (e.g. `TransverseIsotropicFlowModel`). The residual is correct, so solves still reach the right answer — but Newton degrades to a slow defect-correction (tens of oscillating SNES steps instead of 1–2), because the linearisation it follows is wrong by a few percent. The bulk Jacobian, `derive_by_array`, and the JIT are all correct; the error is confined to the Nitsche boundary terms for anisotropic stress.

### Reproduce

`snes_test_jacobian` on a minimal TI Stokes (tilted director, `η₁ ≠ η₀`), varying only the boundary condition / geometry:

| configuration | `‖J − J_fd‖/‖J‖` |
|---|---|
| box + TI | `3e-11` ✓ |
| annulus + Dirichlet + TI | `2e-9` ✓ |
| annulus + field-varying viscosity + TI | `3e-11` ✓ |
| annulus + **penalty** free-slip (`add_natural_bc`) + TI | `6.7e-11` ✓ |
| annulus + Nitsche + **isotropic** | `2e-9` ✓ |
| **annulus + Nitsche + TI** | **`0.069` (7%) ✗** |

The failure is precisely the **Nitsche × anisotropy** intersection. `Amat` and `Pmat` are equally wrong, and the ratio is invariant to pressure-nullspace on/off, monolithic-`lu` vs fieldsplit-Schur — so it is the tangent the BC contributes, not solver configuration.

A self-contained reproducer is `scripts/jac_minimal_box.py` (on the `feature/fault-convection` worktree); it sweeps `GEOM`/`BC`/`VISC`/`NOWEAK` and runs `snes_test_jacobian`. Happy to upstream a trimmed version as a regression test.

### Mechanism (likely a small fix)

[...]

PR #258, opened 2026-06-19
Consistent Jacobian tangent (unwrap-before-differentiate fix) — opt-in, default-off

lmoresi:
### The bug
SNES Jacobian assembly differentiated the residual flux `F1` **while the effective
viscosity was still a wrapped `UWexpression` atom**, so `∂η/∂(grad v)` was silently dropped
from every Jacobian. UW3 viscoplastic Stokes was therefore running an accidental **Picard /
defect-correction** tangent, not full Newton — the unwrap happened *after* the derivative
instead of *before* it. Constant-viscosity problems were unaffected, which is why it stayed
hidden behind the "≈20 Picard iterations is intrinsic" folklore.

### The fix (default-off, bit-identical by default)
- **`symbolic_keep_constants` unwrap mode** — expands `UWexpression` atoms down to (but not
  including) truly-constant atoms (η₀, τ_y stay symbolic for the `constants[]` mechanism).
  The keep-symbolic predicate is the *same* `_is_truly_constant` used by `constants[]`
  extraction, so they cannot drift (drift-guard test).
- **`solver.consistent_jacobian`**: `False` (default, frozen/Picard, **bit-identical**) /
  `True` (full Newton) / `"continuation"` (Picard→Newton via an `α`-blend routed through
  `constants[]`, so switching costs no JIT recompile and `α=0` is bit-identical).
- **`Constitutive_Model.flux_jacobian`** hook (default `None`) for a model to supply a smooth
  tangent law; `_newton_flux` is guarded so the default path never evaluates it.

The residual never goes through the new path, so a converged solution always satisfies the
exact constitutive law.

### Non-regression evidence
- Constant-viscosity Jacobian **symbolically bit-identical** (0/N blocks change).
- `test_1010`, SolCx `test_1015`, `test_0610`, asymmetric-Jacobian guard, units (64) pass;
  `level_1 tier_a` 225/225.
- Crash-isolated (`--forked`) `level_2 tier_a` failure set is **identical to pristine
  `development`** — the pre-existing reds (`test_1012` gmsh crash + 3 `test_1052` VEP) are not
  introduced here.

### Scope / why a smooth Jacobian is not the VEP fix
For a hard-`Min` residual, a smooth Jacobian is the consistent tangent of a *different*
(harmonic) problem — it diverges more than Picard. The robust route for hard-yield VEP is
**problem-space homotopy** (ramp the softmin softness δ→0; δ=0 is identically `Min`), which is
a **separate follow-up PR** that reuses this PR's `constants[]`-ramp machinery. See
`docs/developer/design/jacobian-consistent-tangent.md` for the full tangent hierarchy and the
successor plan.

### DIFF
--- a/docs/developer/design/jacobian-consistent-tangent.md
+++ b/docs/developer/design/jacobian-consistent-tangent.md
@@ -1,0 +1,99 @@
+# Consistent Jacobian tangent for nonlinear (viscoplastic) solves
+
+## The bug
+
+The SNES Jacobian assembly in `src/underworld3/cython/petsc_generic_snes_solvers.pyx`
+formed the `G0`–`G3` blocks by calling `sympy.derive_by_array` / `sympy.diff` on the
+residual flux `F1` **while the effective viscosity was still a wrapped `UWexpression`
+atom**. `derive_by_array` therefore treated the viscosity as a constant, so the term
+`∂η/∂(grad v)` was **silently dropped from every Jacobian**.
+
+Consequence: UW3 viscoplastic Stokes was running an accidental **Picard /
+defect-correction** tangent, not full Newton — not by design, but because the unwrap
+happened *after* the derivative instead of *before* it. Constant-viscosity problems were
+unaffected (η has no `grad v` dependence), which is why the entire constant-viscosity
+suite passed bit-identically and the bug stayed hidden behind the "≈20 Picard iterations
+is intrinsic" folklore.
+
+## The fix (this PR — opt-in, default-off)
+
+A new unwrap mode and a gate, so the default path is **bit-identical** to the historical
+(Picard) behaviour and the consistent tangent is **opt-in**.
+
+- **`symbolic_keep_constants` unwrap mode** (`function/expressions.py`): expands every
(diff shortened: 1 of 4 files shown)


bknight1:
@lmoresi I've had a look at this PR and it's vastly improved the convergence of the VP models I've been working on, both when using continuous and True. The only issue is compilation time with multimaterials, which can be >5 mins, with 'continuous' taking longer than True. Another issue is the compilation appears to occur twice after updating the constitutive model (one to check the hash is different then a rebuild the jacobians). 

I have made some modifications to fix these issues, happy to discuss committing these changes.

lmoresi:
Reviewed + rebased onto current development (resolved the petsc_generic_snes_solvers.pyx conflict — kept both this PR's Jacobian methods and development's set_custom_mg; builds clean, Stokes 6/6 passes). 

Safe to merge: the consistent tangent is opt-in and default-off (`consistent_jacobian = False`), so the default assembly path is bit-identical to prior behaviour (green CI confirms no regression). The opt-in Newton benefit and the deferred harmonic-Min flux_jacobian are follow-up work under #239. Merging via admin (team-authored).

PR #260, opened 2026-06-20
Skip mesh view length-scale test without pyvista

ss2098:
This updates test_mesh_view_displays_length_scale() to skip when optional visualization dependencies are unavailable.

The test exercises mesh.view(), which depends on PyVista/trame visualization support. Without those optional dependencies, the length-scale logic tests pass, but the visualization-specific test fails with ModuleNotFoundError: No module named 'pyvista' or a pyvista[jupyter]/trame dependency error.

No solver or units-system logic is changed.

Tested with:

```bash
pixi run python -m pytest tests/test_0815_mesh_length_scale.py -vv
```

Result:

```text
10 passed, 1 skipped in 2.37s
```

### DIFF
--- a/tests/test_0815_mesh_length_scale.py
+++ b/tests/test_0815_mesh_length_scale.py
@@ -236,6 +236,11 @@ def test_length_units_matches_coordinate_units():
 )
 def test_mesh_view_displays_length_scale():
     """Test that mesh.view() displays length scale information."""
+    pytest.importorskip("pyvista")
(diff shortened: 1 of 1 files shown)


lmoresi (approved):
Correct use of `pytest.importorskip` for the pyvista/trame visualization deps that `mesh.view()` needs — the test now skips cleanly in headless/minimal environments. Green CI. Merging.

PR #261, opened 2026-06-20
Fix dimensionality demo for strict units

ss2098:
This updates docs/examples/Dimensionality-Demo.py so that it runs with the current strict-units behavior.

The example previously failed because it:
1. created unit-bearing MeshVariable objects before reference quantities were set,
2. assigned plain dimensional values directly into .array, and
3. used older .to_nd(), .from_nd(), and UWQuantity reference-scaling calls that are not available in the current API.

The updated example keeps the dimensionality-tracking demonstration, but uses the current strict-units pattern:
- set reference quantities before creating unit-bearing variables,
- write non-dimensional values into .data,
- use scaling coefficients for dimensional/non-dimensional conversion.

Tested with:

pixi run python docs/examples/Dimensionality-Demo.py

### DIFF
--- a/docs/examples/Dimensionality-Demo.py
+++ b/docs/examples/Dimensionality-Demo.py
@@ -14,6 +14,14 @@ import numpy as np
 # All variables now track their dimensionality automatically from units.
 
 # %%
+model = uw.get_default_model()
+model.set_reference_quantities(
+    domain_depth=uw.quantity(3000, "km"),
+    plate_velocity=uw.quantity(5, "cm/year"),
+    density=uw.quantity(3.3, "g/cm**3"),
+    temperature_diff=uw.quantity(1000, "kelvin"),
+)
+
 mesh = uw.meshing.UnstructuredSimplexBox(
     minCoords=(0.0, 0.0),
     maxCoords=(3000.0, 3000.0),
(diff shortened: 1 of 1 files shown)


lmoresi (approved):
Thanks @ss2098 — verified it runs end-to-end on current `development` (the dead `to_nd`/`nd_array`/`from_nd` API is correctly removed — those were dropped in the units simplification). Merging since a running demo is a clear win over the broken one.

**Follow-up suggestion (non-blocking):** since this is specifically the *Dimensionality* demo, it would showcase the units interface better to assign **dimensional quantities via `.array`** rather than manual `value / scaling_coefficient` → `.data`. I verified this works once reference quantities are set:
```python
with uw.synchronised_array_update():
    T.array[...] = uw.quantity(1300.0, "kelvin")   # dimensional assignment
print(T.data[...])   # -> 1.3  (non-dimensional, == 1300/1000)
```
i.e. `.array` = dimensional (UnitAware), `.data` = non-dimensional — the manual scaling then isn't needed. Happy to push that polish as a follow-up if you'd like; not blocking this merge.

ss2098:
Thanks Professor @lmoresi  that makes sense. I created a small follow-up PR with the polished version here: #268.

It updates the demo to assign dimensional quantities through `.array` and then shows the corresponding non-dimensional values stored in `.data`, following your suggestion.

Kindly check when you have a chance and see if it works as intended.

Thanks,
Saurabh

PR #262, opened 2026-06-20
Fix unit-aware coordinate evaluation example

ss2098:
This updates docs/examples/example_unit_aware_coordinates.py so that it runs with the current evaluate() behavior.

The example previously failed because it:
1. assigned mesh vertex coordinates to a degree=2 MeshVariable, causing a shape mismatch, and
2. passed a Python list of UWQuantity coordinate tuples directly to uw.function.evaluate(), which is not accepted by the current non-dimensionalisation path.

The updated example:
- initializes the degree=2 MeshVariable using its own coordinates,
- converts unit-aware coordinate values explicitly to SI base units,
- passes numeric coordinate arrays to uw.function.evaluate(),
- preserves the original purpose of demonstrating coordinate evaluation with unit-aware inputs.

Tested with:

pixi run python docs/examples/example_unit_aware_coordinates.py

### DIFF
--- a/docs/examples/example_unit_aware_coordinates.py
+++ b/docs/examples/example_unit_aware_coordinates.py
@@ -1,13 +1,68 @@
 """
 Example: Using unit-aware coordinates with uw.function.evaluate()
 
-This demonstrates the built-in unit-aware coordinate handling in Underworld3's
-evaluate() function. As of the latest version, evaluate() automatically handles
-unit conversion, so you can pass coordinates with units directly!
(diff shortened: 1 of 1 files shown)


lmoresi (approved):
Verified it runs end-to-end on current `development`. Good update — it now documents the *correct* current patterns (numeric arrays via a `coords_to_si()` helper / `T.coords` before `uw.function.evaluate`, and the 'don't assign `mesh.X.coords` to a degree>1 MeshVariable' guidance). Thanks @ss2098. Merging.

PR #264, opened 2026-06-20
fix(mesh): boundary normals and domain membership must track a deformed mesh

lmoresi:
Two related bugs where **analytic/cached *undeformed* geometry survived a `mesh.deform()`**, corrupting Nitsche/penalty free-slip BCs and semi-Lagrangian advection on free (deforming) surfaces. Surfaced while building a free-slip + stress-driven dynamic-topography free-surface convection scheme.

## Bug 1 — projected boundary normals stayed radial on a deformed surface
`mesh.Gamma`'s base scalars carry the C-code `petsc_n[]` (the PETSc facet normal), which is only defined inside surface-integral kernels. `_update_projected_normals` *point-evaluated* it, so it fell back to the coordinate → `(x,y)/r` = **radial**. `Gamma_P1` thus never tracked deformation and `add_nitsche_bc` constrained `v·r̂` instead of `v·n̂_true`, leaking throughflow ∝ surface tilt (verified ~9° off at 15% topography).

**Fix:** new `Mesh.boundary_normal(label)` assembles the exact PETSc facet normals (`computeCellGeometryFVM`, outward, area-weighted) from **only that boundary's faces** → per-boundary (a corner shared with another boundary keeps its own one-sided normal, never an average across the discontinuity — handles the box free-slip/no-slip corner *and* a Cartesian free-surface top-meets-wall corner), deformation-tracking, smooth on a smooth boundary. `add_nitsche_bc` (Stokes + Scalar) now uses it; the global `Gamma_P1` is **unchanged for back-compat**. The field is refreshed on every `deform` so BCs that captured its `.sym` at setup read the new geometry. Parallel-safe (guards `getStratumIS` against dead label values on ranks holding no faces of a boundary).

## Bug 2 — `points_in_domain()` / `function.evaluate()` used the original boundary
The boundary-skeleton kd-tree is built from `mesh._nav_coords`, which was captured as a *reference to the original coords* in `__init__` and never refreshed in `nuke_coords_and_rebuild` (only the per-cell control-point arrays were, on `adapt`). On a volume mesh `_nav_coords` stayed at the undeformed boundary, so a region that bulged past the original surface (`r > r_o` on a free surface) read as **exterior** — stranding SL trace-back feet there and mis-locating evaluations to the old boundary, injecting the cold boundary value at topographic highs.

**Fix:** invalidate the boundary kd-tree **and refresh `_nav_coords`** from the current DM coordinates on every `deform`.

## Tests
- `test_0056` — `boundary_normal` tracks a deformed annulus + a Cartesian free-surface corner (tier_a).
- `test_0057` — deformed-domain membership + evaluation (tier_a).
- `test_1060` Nitsche free-slip passes **serial and np2**; broad serial regression (meshes, pointwise, evaluate, slip, Nitsche) green.

## Known follow-up
ADD-reduce the unnormalised facet contributions at partition-seam boundary vertices before normalising — those few vertices currently use a this-rank-only stencil (parallel-safe and correct everywhere else; flagged with a `TODO` in `_assemble_boundary_normal`).

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -4723,8 +4723,8 @@ class SNES_Stokes_SaddlePt(SolverBaseClass):
             a fault orientation field).
         normal : sympy.Matrix or list, optional
             Boundary unit normal used in the Nitsche consistency, symmetry,
-            and pressure-coupling terms. Default ``None`` uses the PETSc
-            boundary facet normal ``mesh.Gamma_N``.
+            and pressure-coupling terms. Default ``None`` uses the per-boundary,
+            deformation-tracking ``mesh.boundary_normal(boundary)``.
         gamma : float, default=10.0
             Dimensionless stabilisation parameter. Typical values 5--20
             for P2 elements.
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -2928,10 +2928,11 @@ class SNES_Vector(SolverBaseClass):
         # in the embedding space with an implicit tangency constraint.
         dim = mesh.cdim
 
-        # Surface normal components — use projected P1 normals by default.
-        # These are smooth, consistently oriented, and converge in 3D.
-        Gamma_P1 = mesh.Gamma_P1
-        n = [Gamma_P1[i] for i in range(dim)]
+        # Surface normal components — use this boundary's own deformation-
+        # tracking facet normal (see Mesh.boundary_normal); the legacy global
+        # mesh.Gamma_P1 stays radial on a deformed surface.
+        bnorm = mesh.boundary_normal(boundary)
+        n = [bnorm[i] for i in range(dim)]
 
         # Constraint direction: defaults to surface normal
         if direction is not None:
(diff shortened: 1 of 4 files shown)


lmoresi:
This fix is a **critical component of the adaptive-meshing / free-surface work**, not just the free-slip BC path.

While validating mmpde adaptive convection we traced the long-standing "holes / giant empty cells in the adapted mesh + wrecked adaptive convection" failure to **Bug 2 here** (deformed-mesh point location). The chain:

- `metric_density_from_gradient` builds a P1 density with strictly-positive nodal values, but on a *deformed* mesh `uw.function.evaluate` mis-located points (stale `_nav_coords`) and returned **negative** values *even at the field's own DOF coords* (measured min ≈ −3.0).
- Negative density → non-SPD metric → the mover either NaN-bailed (a protective stall that *hid* the problem) or, after the SPD-floor in #259, acted on garbage-magnitude metric → intermittent giant cells / holes / divergence.

With this PR, the metric evaluates correctly/positive on the deformed mesh, so the mover is fed a valid metric and adaptation is stable.

**Validation (with this fix):** forced adaptation *every step* (39/39) at R=5 under vigorous stagnant-lid convection (Ra1e6, Δη1e3) → mesh clean every frame (folded=0, cell-area-ratio ~14 flat), vrms→18, Nu→1.86. The same configuration without the fix produced area-ratio spikes 189/96/875 (holes). So this unblocks adaptive convection (and, as noted, the free-surface dynamic-topography work).

A small complementary hardening (mover bakes the metric from positive nodal values via monotone RBF/Shepard, so it can never go negative regardless of eval precision) will follow as a separate bugfix PR.

lmoresi:
Addressed the Copilot review in e086e72a:

- **`nverts == cdim` (the substantive one):** now counts P1 vertices **per facet** from its closure (`vStart..vEnd`), so 3D non-simplex boundary faces (quads on a hex mesh = 4 verts) are handled correctly, not just simplex edges/triangles.
- Dropped the unused `cKDTree` import in `boundary_normal()` and the unused `import underworld3 as uw` in `_assemble_boundary_normal()`.
- Fixed the `add_nitsche_bc` `normal` docstring (default is now `mesh.boundary_normal(boundary)`, not `mesh.Gamma_N`).
- Test tidy-ups: removed the unused `surf_now`; cleaned the editorial comment fragment.

`test_0056` + `test_0057` pass (3 passed). CI will re-run on the new commit.

lmoresi:
I'm going to merge this in - this bug is being rediscovered endlessly and it's partly because we keep these PRs alive while working on other stuff. Keep raising issues !

Issue #254, opened 2026-06-18
Stokes_Constrained: parallel-correctness residuals (partition-dependent [p,λ] gauge + constraint-assembly drift + lossy knockout)


lmoresi:
## Summary

While root-causing the #248 Zhong-benchmark velocity catastrophe (now fixed by #242 — it was a broken `SphericalShellInternalBoundary` `Lower`/`Internal` label bug), three smaller but real parallel-correctness issues in `Stokes_Constrained` (in-saddle multiplier free-slip) were isolated with reliable integral diagnostics. Tracking them here for a focused fix.

**Method note:** point evaluation (`uw.function.evaluate`) is unreliable for serial-vs-parallel field comparison here (gave 134%-of-|v| pointwise noise while volume L2 differed 0.4%). Use partition-independent **integrals** only.

**Ruled out:** velocity rigid-rotation nullspace (angular momentum ∫r×v dV reproducible to ~1e-5 serial vs np=8); iterative under-convergence (serial & np=8 each bit-identical at tol 1e-7 vs 1e-11 — both fully converge, to *different* answers).

## 1. `[p,λ]` / topography gauge is partition-inconsistent (the nullspace bug)
The combined constant-(pressure, multiplier) gauge (`_build_block_gauge_nullspace_vector`) is a true velocity null mode, but its **level is partition-dependent: mean pressure = 3.333 (serial) vs 2.983 (np=8), 10.5%**. Because topography is read from the multiplier, this corrupts topography across ranks unless mean-stripped. `set_pressure_gauge("Upper", 0.0)` (#250, global `BdIntegral` callback) pins it consistently → 0.4%, bit-identical on velocity.
**Fix:** make the gauge partition-reproducible by construction (default-on pin, or enforce `topography(reference="mean")`); add a parallel regression asserting `meanP`/topography match serial to ~1e-6. **Highest priority — topography correctness, fix already exists.**

## 2. Residual ~0.4% velocity partition-dependence in constraint/augmentation assembly (separate, NOT a nullspace)
[...]

PR #265, opened 2026-06-21
Stokes_Constrained parallel correctness: gauge, convergence, knockout, rotation-gauge

lmoresi:
Parallel-correctness work for `SNES_Stokes_Constrained` (the in-saddle Lagrange-multiplier free-slip solver), bringing the constrained free-slip solve to round-off partition-independence.

## What's here

**1. Auto pressure gauge.** Guarded `auto_pressure_gauge` (default on) pins the surface-mean pressure on the first constraint boundary when a pressure null space is active and the user set no gauge. Physics-neutral; raw mean pressure 10.5%→0.4% at np=8, velocity bit-identical. Topography is read gauge-invariantly via `topography(reference="mean")` — the raw multiplier keeps its own independent gauge freedom the pressure pin does not touch.

**2. Solver convergence.** The grouped `u | [p,h]` Schur preconditioner uses inner *iterative* sub-solves = a variable preconditioner; plain `gmres` is invalid for that and false-converged (preconditioned norm reported CONVERGED while the true residual blew up), landing on a wrong, partition-dependent answer. Fixed with `ksp_type=fgmres`, `ksp_norm_type=unpreconditioned`, a tightened Eisenstat–Walker default, and a defensive `ksp_max_it`. Now genuinely converges (true resid ~1e-13) and reproduces the Zhong benchmark.

**3. Knockout audit + LU guard.** Confirmed the interior-multiplier knockout is lossless (the earlier "2–5% off" was iterative conditioning, not lost physics). Added a warning that a monolithic direct solve (`pc_type lu/cholesky`) of the constrained saddle point is a serial diagnostic only.

**4. Rigid-rotation gauge fix (the main new result).** On a fully free-slip velocity the rigid rotations are a true nullspace of the velocity block (`A_uu·rotation = 0`). The monolithic nullspace attached for the solve is **not** removed inside the fieldsplit/Schur iteration — the inner velocity KSP has no rotation nullspace, so it leaves an unconstrained, **partition-dependent** rigid rotation in the velocity. The operator is blind to it (the true residual still converges to machine precision), so the tangential velocity differed serial-vs-parallel by ~0.1–1% while the physical (radial) flow was already clean. `SNES_Stokes_SaddlePt.solve` now projects the velocity rotation modes out of the converged solution via `MatNullSpaceRemove` (`_remove_velocity_rotation_gauge`) — the velocity analogue of the constant-pressure gauge. Because the rotation is a genuine nullspace, removing it does not change the residual and yields the same rotation-free solution on any decomposition.

Validated on the constrained free-slip problem: tangential surface velocity serial-vs-np8 relative difference **1.2% → 2.5e-14**, iteration count unchanged. New serial guard `tests/test_1065_rotation_gauge_freeslip.py` (the existing `test_1063` cases all pin the rotation nullspace with a Dirichlet BC, so they did not cover this path).

## Known PETSc limitation (documented, not fixed here)

`add_natural_bc` on an **internal** surface assembles a slightly partition-dependent load: `DMPlexComputeBdResidual_Single_Internal` attributes each interior facet's integral to a partition-dependent `support[0]` cell closure, so a seam facet's non-owned closure DOFs are dropped at global assembly (F0 differs ~3e-4). The refined `TODO(BUG)` records this and the workarounds that were tested and ruled out (owner-only label strip, 1-cell overlap, codim-1 submesh). Once the rotation gauge is removed, this leaves only ~3e-6 in the velocity (the elliptic solve smooths the localized load error); applying the internal traction as a volume body force makes it round-off.

## Tests

- `test_1064` (constrained spherical shell) + `test_1010` (Cartesian Stokes): 9 passed, 2 pre-existing xfails.
- `test_1063` (parallel constrained free-slip, np=2): passed.
- `test_1065` (new rotation-gauge guard): passed.

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -1008,6 +1008,8 @@ class SolverBaseClass(uw_object):
                 self._stokes_nullspace = None
             if hasattr(self, "_stokes_nullspace_basis"):
                 self._stokes_nullspace_basis = ()
+            if hasattr(self, "_velocity_rotation_nullspace"):
+                self._velocity_rotation_nullspace = None
             if hasattr(self, "_constant_nullspace_obj"):
                 self._constant_nullspace_obj = None
 
@@ -4697,6 +4699,7 @@ class SNES_Stokes_SaddlePt(SolverBaseClass):
         self._petsc_velocity_nullspace_basis = ()
         self._stokes_nullspace = None
         self._stokes_nullspace_basis = ()
+        self._velocity_rotation_nullspace = None
 
         # Construct strainrate tensor for future usage.
         # Grab gradients, and let's switch out to sympy.Matrix notation
@@ -5338,6 +5341,10 @@ class SNES_Stokes_SaddlePt(SolverBaseClass):
     def _reset_stokes_nullspace(self):
         self._stokes_nullspace = None
         self._stokes_nullspace_basis = ()
+        # the cached velocity rotation nullspace is built from node coordinates,
+        # so it must be invalidated whenever the nullspace config / geometry
+        # changes (e.g. mesh deformation) — see _build_velocity_rotation_nullspace.
+        self._velocity_rotation_nullspace = None
 
     def _pressure_dirichlet_bcs(self):
         """Return essential boundary conditions applied to the pressure field."""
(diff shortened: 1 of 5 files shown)


lmoresi:
Reviewed (high-recall code review of the full diff):

**Approve.** The four parts are mathematically — not just empirically — justified:
1. **Auto pressure gauge** is conservatively guarded (6 preconditions; no-op unless all hold) and physics-neutral (velocity bit-identical).
2. **fgmres + unpreconditioned norm** is a correctness fix, not tuning: a non-flexible Krylov over a variable fieldsplit/Schur preconditioner false-converges on the preconditioned norm.
3. **Knockout audit** docstring proves the interior-multiplier reduction is lossless (h_i = −M_ii⁻¹M_ib h_b, O(ε) feedback), reframing the old "2–5%" as conditioning.
4. **Rotation-gauge projection** removes a *genuine* velocity nullspace from the converged solution → residual-neutral and partition-independent by construction; placement (after both solve branches, before copy-back) is correct. The recorded negative result on the in-solve sub-block route is valuable.

The internal-natural-BC `TODO(BUG)` honestly documents the PETSc `support[0]` limitation with ruled-out workarounds. Serial tests (test_1064/1065/1010) confirmed locally; CI green.

Caveat carried forward to #254: the headline parallel numbers (1.2% → 2.5e-14 tangential at np8) still want the remote np=8 confirmation, but serial correctness and the mathematics stand independently.

Merging via admin override (team-authored; branch protection needs a non-author review).

PR #266, opened 2026-06-22
fix(meshing): mmpde metric — monotone RBF bake from nodal values

lmoresi:
Restores the intended **RBF metric-eval robustness** for the mmpde mover (a design decision that had been undercut by an FE bake).

## Problem
The mover's `metric_eval="rbf"` path baked the metric via an **FE evaluation at the fixed reference cloud `ref`**, then Shepard-interpolated. On a *deformed* mesh that FE evaluation mis-locates points and the P1 density — **strictly positive by construction** — comes back **negative** (measured ≈ −3, even at a field's own DOF coords). A negative density is a **non-SPD metric**, which the mover either NaN-bails on (a hidden stall) or, with the #259 SPD-floor, acts on as "coarsen hard here" → **giant cells / holes / divergence** in the adapted mesh.

## Fix
Bake the metric at the **current mesh NODES** (its own DOF locations) and Shepard-interpolate from there:
- **Monotone:** a Shepard (positive-weight, convex) average of positive nodal values is guaranteed ≥ 0 — the eval can never produce a non-SPD metric.
- **Robust + fast:** nodes are always inside the mesh (no drift/out-of-domain), and Shepard needs no per-step cell location. RBF doesn't need high-precision eval.
- `ref` is kept for the `_edge_mats` reference frame.

## Relationship to #264
Best paired with #264 (deformed-mesh point-location fix), which makes the nodal FE bake exact. With both, the adapted mesh stays clean under **forced every-step adaptation at R=5** (folded=0, cell-area-ratio flat, vrms→18, Nu→1.86) where it previously tore holes (area-ratio spikes 189/96/875).

One-line source change; validated in a sibling worktree. CI runs the full mover suite.

### DIFF
--- a/src/underworld3/meshing/smoothing.py
+++ b/src/underworld3/meshing/smoothing.py
@@ -3145,22 +3145,32 @@ def _winslow_mmpde(mesh, metric, pinned_labels, verbose,
         mesh._mmpde_reference_coords = ref
 
     # --- RBF/Shepard bake of the metric (the production-fast path) ------
-    # Evaluate the analytic metric ONCE on the fixed reference cloud, then
-    # interpolate to the moving centroids each step via k-NN inverse-
-    # distance (Shepard). The reference cloud is fixed in space ⇒ Eulerian.
+    # Bake the metric at the CURRENT mesh NODES (its own DOF locations), then
+    # interpolate to the moving centroids each step via k-NN inverse-distance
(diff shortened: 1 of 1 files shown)


lmoresi:
This is a minor detail - the metric evaluation for the adaptive meshing needs to be 1) fast and 2) monotonic ... rbf is a good choice, especially since it does not require the full mesh interpolation infrastructure to be valid during mesh movement (a difficult task to sequence correctly)

PR #268, opened 2026-06-22
Polish dimensionality demo unit assignment

ss2098:
This follow-up to #261 updates docs/examples/Dimensionality-Demo.py to use the public unit-aware .array interface for dimensional assignment.

The previous fix made the demo run by manually writing non-dimensional values into .data. This polish assigns dimensional quantities through .array, for example:

T.array[...] = uw.quantity(1300.0, "kelvin")

and then shows that .data contains the corresponding non-dimensional internal storage.

This better demonstrates the intended units interface:

- .array for dimensional/unit-aware values
- .data for non-dimensional internal values

Tested with:

pixi run python docs/examples/Dimensionality-Demo.py

### DIFF
--- a/docs/examples/Dimensionality-Demo.py
+++ b/docs/examples/Dimensionality-Demo.py
@@ -55,10 +55,13 @@ print(f"  p: {p.scaling_coefficient:.0e} Pa")
 # %% [markdown]
 # ## 3. Non-Dimensional Conversion
 #
-# Current strict-units mode stores non-dimensional values in `.data`.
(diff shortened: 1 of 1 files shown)


lmoresi (approved):
This is exactly the polish I suggested on #261 — `.array[...] = uw.quantity(value, unit)` dimensional assignment instead of manual scaling, plus the `.array` (dimensional) vs `.data` (ND) explanation. Verified it runs end-to-end (✅ Dimensional analysis complete). The demo now actually showcases the units interface. Thanks @ss2098!

PR #270, opened 2026-06-22
Make Poisson gradient recovery example script-safe

ss2098:
This updates docs/examples/heat_transfer/basic/Ex_Poisson_Gradient_Recovery.py so that the migrated notebook example also runs cleanly as a plain Python script.

pixi run python docs/examples/heat_transfer/basic/Ex_Poisson_Gradient_Recovery.py

the script failed because it used notebook-only display(...) calls before display was defined in the script environment:

NameError: name 'display' is not defined

The example had been migrated from a notebook, where display() is available automatically, but this is not true when executing the .py file directly.

Changes made:

- Replaced notebook-only display(...) calls with terminal-safe print(...) calls.
- Removed the notebook-specific jupyter_backend="html" argument from pl.show(...), because PyVista warns that this argument is ignored outside a Jupyter environment.
- Kept the solver setup, equations, boundary conditions, and numerical workflow unchanged.

This keeps the example usable both as documentation and as a script-style test of the Poisson gradient recovery workflow.

Tested with:

pixi run python docs/examples/heat_transfer/basic/Ex_Poisson_Gradient_Recovery.py

Result: the script proceeds past the previous display() failure and runs without that NameError.

### DIFF
--- a/docs/examples/heat_transfer/basic/Ex_Poisson_Gradient_Recovery.py
+++ b/docs/examples/heat_transfer/basic/Ex_Poisson_Gradient_Recovery.py
@@ -2,8 +2,8 @@
 """
 # 📚 Poisson Gradient Recovery
 
-**PHYSICS:** heat_transfer  
-**DIFFICULTY:** basic  
+**PHYSICS:** heat_transfer
+**DIFFICULTY:** basic
 **MIGRATED:** From underworld3-documentation/Notebooks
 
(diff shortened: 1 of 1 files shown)


lmoresi (approved):
Verified it runs end-to-end (exit 0, no errors) — produces the expected gradient-recovery output. Thanks @ss2098. Merging.

PR #272, opened 2026-06-22
Fix Darcy 1D benchmark for current API

ss2098:
This updates docs/examples/porous_flow/basic/Ex_Darcy_1D_benchmark.py so that the migrated Darcy 1D benchmark runs with the current API.

pixi run python docs/examples/porous_flow/basic/Ex_Darcy_1D_benchmark.py

the script failed at the no-gravity comparison setup because it used an outdated MeshVariable clone pattern:

p_soln.clone("P_no_g", r"{p_\textrm{(no g)}}")

The example only needs the no-gravity pressure profile for comparison, so this update avoids cloning entirely. It solves the no-gravity case first, evaluates and stores the no-gravity pressure profile as a NumPy array, then solves the gravity case and evaluates the final pressure profile.

This also flattens the evaluated pressure arrays with np.asarray(...).reshape(-1), so the Matplotlib comparison uses 1D arrays and avoids shape errors from singleton dimensions returned by uw.function.evaluate.

The PyVista visualisation block is now optional. If PyVista is not installed, the example skips only the visualisation and still runs the analytical comparison and validation.

The solver setup, permeability structure, boundary conditions, analytical solution, and validation logic are unchanged.

Tested with:

pixi run python docs/examples/porous_flow/basic/Ex_Darcy_1D_benchmark.py

Result:

Validation passed: numerical solution matches analytical solution

### DIFF
--- a/docs/examples/porous_flow/basic/Ex_Darcy_1D_benchmark.py
+++ b/docs/examples/porous_flow/basic/Ex_Darcy_1D_benchmark.py
@@ -32,7 +32,7 @@ permeabilities. This example:
 For 1D flow through two layers with permeabilities k1 and k2:
 - Pressure is linear within each layer
 - Flux is continuous across the interface
-- Pressure gradient is discontinuous (inversely proportional to k)
+- Pressure gradient is discontinuous, inversely proportional to k
 
 ## Parameters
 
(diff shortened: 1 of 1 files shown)


lmoresi (approved):
Verified it runs end-to-end (exit 0, no errors). Thanks @ss2098. Merging.

PR #274, opened 2026-06-22
Fix explicit flow grains script run

ss2098:
This updates docs/examples/porous_flow/advanced/Ex_Explicit_Flow_Grains.py so that the explicit grain-flow example runs correctly as a plain Python script.

The example failed during pipemesh.write_timestep(...) with:

AttributeError: 'NoneType' object has no attribute 'name'

The gmsh-created mesh did not have an explicit coordinate system, so self.CoordinateSystemType was None when the mesh writer serialized coordinate-system metadata. This PR sets:

coordinate_system_type=uw.coordinates.CoordinateSystemType.CARTESIAN

in the uw.discretisation.Mesh(...) constructor.

The example also imported PyVista during a terminal run. The visualization blocks are now guarded with a notebook check, so the solve, timestep writing, porosity calculation, and swarm advection path run without optional notebook visualization dependencies.

The gmsh geometry, boundary tags, Stokes setup, pressure projection, swarm advection, and output logic are otherwise unchanged.

Tested with:

pixi run python docs/examples/porous_flow/advanced/Ex_Explicit_Flow_Grains.py

Result:

Explicit grain flow example complete: 556 steps

### DIFF
--- a/docs/examples/porous_flow/advanced/Ex_Explicit_Flow_Grains.py
+++ b/docs/examples/porous_flow/advanced/Ex_Explicit_Flow_Grains.py
@@ -69,6 +69,19 @@ nest_asyncio.apply()
 import numpy as np
 import sympy
 
+
+def in_notebook():
+    try:
+        from IPython import get_ipython
+    except ImportError:
+        return False
+
+    shell = get_ipython()
+    if shell is None:
+        return False
+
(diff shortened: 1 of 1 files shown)


lmoresi (approved):
Verified it runs end-to-end (exit 0; the only flagged line is a benign geometric-multigrid UserWarning, no traceback). Thanks @ss2098. Merging.

PR #275, opened 2026-06-23
Nitsche BC: scale penalty by a local per-cell mesh size

lmoresi:
## Summary

The Nitsche free-slip / constraint penalty term `gamma*mu/h` previously used a single **global** scalar `h = mesh.get_min_radius()` — the smallest cell *anywhere* in the mesh. The viscosity `mu` was already evaluated locally; only `h` was wrong.

On a non-uniform, adaptive, or deforming mesh this mis-scales the boundary penalty: the global-minimum cell size is applied on every facet (over-stiffening coarser boundary cells), and it **drifts** as refinement or distortion changes the global minimum. This was found while debugging spurious one-step surface "mountains" in adaptive free-surface convection: an mmpde-distorted interior cell drags the global minimum far below the surface cells, the held-lid Nitsche penalty over-stiffens, and the surface normal velocity spikes.

## What changed

- **New `Mesh.cell_size()`** — a deformation/adaptation-tracking, cell-constant (degree-0, discontinuous) scalar field holding each cell's characteristic size (`self._radii`). On a boundary integral the kernel sees the adjacent cell's size. Refreshed on `deform()` and `adapt()` (and via a reinit callback in the remesh transaction), mirroring the existing `boundary_normal` machinery, so it is never stale. The fill is purely **rank-local** (no collective) → parallel-safe.
- **`add_nitsche_bc`** on both `SNES_Stokes_SaddlePt` and `SNES_Vector` now scales the penalty with the local size behind **`local_h=True` (default)**. `local_h=False` restores the exact legacy global-h path. On a uniform mesh `local == global`, so existing tuned behaviour is preserved.

## Validation

**Existing Nitsche suite — green serial + np2** (`test_1060`, `test_0056`, `test_1011`, `test_1014`, `test_1061`). The only np2 failure, `test_1061`, segfaults on unmodified `development` too — it exercises the LU/mumps block-constrained multiplier path and uses no Nitsche, so it is unrelated.

**New `test_1065_nitsche_local_h` (serial + np2)** — on a graded mesh: the penalty size is genuinely local (≈6× the global minimum at a coarse free-slip boundary), tracks deformation (not stale after `deform`), and still solves free-slip correctly (matches the essential-BC reference; `v·n` small on both coarse and fine ends).

> Note: the test asserts the *mechanism* (local scaling + deformation tracking + correctness), not a boundary-`v·n` leak comparison. On a *static* graded mesh global-h is the global *minimum*, so it over-stiffens coarse cells and actually gives a *smaller* leak there — a naive leak test would be backwards. The real cost of global-h is conditioning and the spurious spike on a deforming mesh, demonstrated by the integration below.

**Held-lid adaptive free-surface convection** (`fs_convection_goal4`, `--surface freesurface --mesh adaptive --res 24 --rho-g 7.5e5 --held-gamma 10`, 40 steps):

| metric | global-h (baseline) | local-h (fix) |
|---|---|---|
| vhmax peak | 2873 | **342** |
| vhmax median | 135.9 | 132.8 |
| glitches | 2 | **0** |
| h_inf peak (mountain) | 2.69e-2 | 7.54e-3 |

Medians are identical → normal operation and per-step cost are unchanged (same `gamma=10`); local-h purely removes the spurious spikes/mountain. No cost hike (unlike cranking `gamma`→1000).

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -2887,7 +2887,7 @@ class SNES_Vector(SolverBaseClass):
         self.petsc_options["ksp_atol"]  = self._tolerance * 1.0e-6
 
 
-    def add_nitsche_bc(self, boundary, g=None, direction=None, gamma=10.0, theta=1):
+    def add_nitsche_bc(self, boundary, g=None, direction=None, gamma=10.0, theta=1, local_h=True):
         r"""Add Nitsche weak enforcement of a velocity constraint along a direction.
 
         For vector solvers (no pressure field), this constrains
@@ -2906,6 +2906,11 @@ class SNES_Vector(SolverBaseClass):
             Dimensionless stabilisation parameter.
         theta : {-1, 0, 1}, default=1
             Symmetry parameter (1=symmetric, -1=skew-symmetric).
+        local_h : bool, default=True
+            Scale the penalty by a local per-cell mesh size
+            (:meth:`Mesh.cell_size`) rather than the global minimum
+            (:meth:`Mesh.get_min_radius`). See
+            ``SNES_Stokes_SaddlePt.add_nitsche_bc`` for details.
 
         Warnings
         --------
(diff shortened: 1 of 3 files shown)


lmoresi:
Reviewed (high-recall code review of the full diff):

**Approve.** Surgical and correctly scoped — `gamma*mu/h` was using `get_min_radius()` (global min cell anywhere) while `mu` was already local, so the penalty was mis-scaled on non-uniform/deforming meshes.

- New `Mesh.cell_size()` mirrors the `boundary_normal` machinery: degree-0 DG field, rank-local fill from `self._radii`, refreshed on `deform()`/`adapt()`/remesh-reinit. The explicit "no var.coords, no collective" path in `_assemble_cell_size` correctly avoids a rank-divergent deadlock — parallel-safe.
- `local_h=True` default with `local_h=False` restoring the exact legacy expression; uniform mesh ⇒ local==global, so tuned setups are preserved.
- Convection validation is convincing: vhmax peak 2873→342, glitches 2→0, **medians identical** (no per-step cost change).

Confirmed via `git merge-tree` that this merges cleanly with #265 (both touch the .pyx but in different methods). CI green.

Minor non-blocking follow-up: the refresh callbacks use bare `except Exception: pass` (matches existing `boundary_normal` house style) — a one-line verbose warning on fill failure in the deform/reinit paths would be cheap insurance against a silently stale size.

Merging via admin override (team-authored; branch protection needs a non-author review).

PR #277, opened 2026-06-24
fix(ddt): units-active semi-Lagrangian trace-back (#267)

### LINKED ISSUE #267: Semi-Lagrangian trace-back is unit-inconsistent under active units (dt not dimensionalised in has_units path)
## Summary
With the units system active (unit-bearing mesh/variables), `NavierStokesSLCN` / `AdvDiffusionSLCN` fail in the semi-Lagrangian trace-back:
```
ValueError: Cannot subtract arrays with incompatible units: 'meter' and 'meter / second'
```

## Root cause
`SemiLagrangian.update_pre_solve` computes the trace-back midpoint (`src/underworld3/systems/ddt.py:2651`):
```python
mid_pt_coords = coords - v_at_node_pts * (0.5 * dt_for_calc)
```
In the `has_units` branch, `coords` is `UnitAwareArray[meter]` and `v_at_node_pts` is `[meter/second]`, but `dt_for_calc` is left as a plain (dimensionless) float (`ddt.py:2519`):
```python
if has_units:
    if hasattr(dt, "to"):
        dt_for_calc = dt.to("second")
    else:
        dt_for_calc = dt          # comment says 'treat as seconds' but leaves it unitless
```
so `v[m/s] * dt[dimensionless] = [m/s]` and the subtraction `[m] - [m/s]` raises.

Naively attaching `seconds` would fix the *units* but be **numerically wrong**: a CFL `dt` computed in nondimensional model-time is not SI seconds — it must be reconciled with the dimensional coords/velocity via the model time-scale. The sibling `not has_units` branch already does the consistent thing (nondimensionalise coords, velocity **and** dt together).

## Fix direction
Make the units-active trace-back internally consistent — either dimensionalise `dt` to seconds via the model time-scale in the `has_units` path, or (cleaner) always perform the trace-back nondimensionally (strip units from coords/velocity/dt together, as the `not has_units` branch does) regardless of `has_units`.

## Reproduction
`docs/examples/Tutorial_Thermal_Convection_Units.py` (PR #263) reaches the time-stepping loop and fails here. That tutorial is the natural regression test once fixed.

Exposed by PR #263 (held pending this fix).

lmoresi:
## Summary
Fixes #267 — `SemiLagrangian.update_pre_solve` crashed for **any unit-bearing model** in the time loop:
```
ValueError: Cannot subtract arrays with incompatible units: 'meter' and 'meter / second'
```

## Root cause
The SL trace-back is performed in the mesh's **non-dimensional (DM)** coordinate space — `evaluate`/`global_evaluate` treat plain arrays as DM coords, and DM point-location uses DM values (`0..L_model`, not dimensional metres). The `has_units` branch instead kept **dimensional** coords (`mesh.X.coords` ≈ metres, ~1e9) and velocity (m/s) and left `dt` a unitless float, so `coords[m] − v[m/s]·dt[dimensionless]` raised — and even patched, dimensional coords fed to a DM that locates in `0..1000` space would **mislocate**.

## Fix
Route the whole trace-back through ND/DM space regardless of units:
- departure point from `psi_star[i].coords_nd` (== `.coords` for a non-units model; the ND reduction of dimensional coords when units are active),
- velocity non-dimensionalised,
- `dt` non-dimensional.

Both RK2 legs (node-velocity and midpoint-velocity) corrected. **The non-units path is unchanged** — it already used exactly this ND logic; the fix just removes the broken parallel dimensional branch.

## Verification
- Units-active `AdvDiffusionSLCN` now runs and **tracks the equivalent non-dimensional run to ~1e-3** (the small residual is the constitutive *diffusivity* scaling under units — a separate concern, not the trace-back; advection-magnitude-independent).
- Non-units SL regression unchanged: `test_0855` (oldframe SL), `test_0610` (NS SLCN), `test_1054`, `test_1100` pass.
- `test_1110_advDiffAnnulus` fails **identically with and without** this change — a pre-existing `PCSetUp_MG`/`DMCreateInjection` multigrid issue, unrelated.
- New regression: `tests/test_1056_units_slcn_traceback.py` (runs + nondim-equivalence).

Unblocks the time-loop half of **#263** (`Tutorial_Thermal_Convection_Units`).

### DIFF
--- a/src/underworld3/systems/ddt.py
+++ b/src/underworld3/systems/ddt.py
@@ -2508,31 +2508,26 @@ class SemiLagrangian(uw_object):
         coords_template = self.psi_star[0].coords
         has_units = hasattr(coords_template, "magnitude") or hasattr(coords_template, "_magnitude")
 
-        # Maintain unit system consistency: either keep everything with units or convert to non-dimensional
-        if has_units:
-            # Physical coordinate system with units
-            # dt must be converted to base SI seconds so that dt * velocity(m/s) = distance(m)
-            if hasattr(dt, "to"):  # It's a Pint quantity
-                dt_for_calc = dt.to("second")  # Convert to seconds (still a quantity)
+        # The semi-Lagrangian trace-back is performed ENTIRELY in the mesh's
+        # NON-DIMENSIONAL (DM) coordinate space: evaluate()/global_evaluate treat
+        # plain arrays as DM coords and the DM point-location uses DM values
+        # (0..L_model, NOT dimensional metres). So coords, velocity AND dt are all
+        # reduced to non-dimensional values here, whether or not the model carries
+        # units. (Previously the has_units branch kept dimensional coords/velocity
+        # and left dt unitless -> a 'meter' vs 'meter/second' subtraction crash and
+        # mislocation against the ND DM; UW3 issue #267.)
+        if hasattr(dt, "magnitude") or hasattr(dt, "value"):
+            # dt carries units -> non-dimensionalise it
+            dt_nondim = uw.non_dimensionalise(dt, model)
+            if hasattr(dt_nondim, "magnitude"):
+                dt_for_calc = float(dt_nondim.magnitude)
+            elif hasattr(dt_nondim, "value"):
+                dt_for_calc = float(dt_nondim.value)
             else:
-                # If dt is already a dimensionless number, treat it as seconds
-                dt_for_calc = dt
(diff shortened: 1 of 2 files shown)


lmoresi:
Self-review: routes the SL trace-back through ND/DM space (the non-units path's existing logic); removes the broken dimensional branch. Repro fixed, regression test passes (2/2), non-units SL suite unchanged (the one annulus failure is a pre-existing MG issue, identical with/without). Merging for the release.

PR #290, opened 2026-06-28
Generalized FMG hierarchy: custom-P transfers for arbitrary nested/non-nested grids (Layer 1)

lmoresi:
## What

Generalize UW3's geometric FMG so the multigrid hierarchy can use **arbitrary coarse grids**  (can be nested *or* non-nested, uniform *or* locally (SBR) refined) by building the prolongations in house (barycentric / local-RBF) and installing them via `PC.setMGInterpolation`, with Galerkin RAP forming the coarse operators. This decouples geometric multigrid from uniform `refine()` nesting, enabling locally-adapted hierarchies while keeping FMG.

Independent of any adapter ("adapt-on-top" is a separate follow-on). The existing `Mesh(refinement=N)` uniform FMG path is unchanged.

## Status — Layer 1, Phase-1 hardening complete

Home: `utilities/custom_mg.py`. Entry point: `set_custom_fmg(solver, coarse_meshes, builder=, field_id=)`.

**Working + tested** — serial `test_1014/1015/1016/1017` (20) + parallel `tests/parallel/test_1017_custom_mg_parallel_mpi.py` (np=2):

- `CustomMGHierarchy`, `set_custom_fmg`, `sbr_refine` / `sbr_refine_where` (local Skeleton-Based Refinement; no MMG; on-rank; conforming).
- **BC applied at every level** (transfers reduced→reduced) with a zero-column guard — the load-bearing invariant (PETSc native FMG does this via constrained sections; custom-P must replicate it).
- **Scalar / single-field-vector** (top-level PC) **and Stokes velocity block** (`field_id=0`).
- **Serial and parallel** (np>1), nested co-partitioned.

### Hardening steps (all done)

1. **Stokes / saddle-point velocity block.** The velocity sub-PC is unreachable until the monolithic Jacobian is assembled (`PCFieldSplit` forms `A_vv` via `MatCreateSubMatrix`; `snes.setUp` builds structure only). So `_install_velocity_block_transfers` forces a Jacobian assembly (`computeFunction`+`computeJacobian` at the zero guess; `max_it=0` fallback), reaches the velocity sub-PC, `reset`s it and rebuilds a **fresh PCMG** from our P (mirrors the proven standalone recipe; sidesteps the `MatProductReplaceMats` live-swap bug), and re-attaches the coupled Stokes nullspace. Wired into `SNES_Stokes_SaddlePt.solve` as a guarded no-op.

2.  **Dropped the throwaway-solver factory.** Each coarse level's BC-constrained reduced map is derived directly from its DM via `copyFields` + `copyDS` + `createDS` (`_coarse_reduced_map`) — the DS carries UW's exact essential-BC definitions and is topology-independent, so it constrains any coarse mesh sharing the solver's boundary labels. Validated byte-identical to the old factory path; leak-free (no SNES / JIT). `set_custom_fmg` / `build` no longer take a `level_solver_factory`.

3.  **Parallel (np>1).** Nested co-partitioned transfers: each rank builds its block of P rank-locally (ghost-inclusive coarse coords → every owned fine node lands in a local coarse simplex), reduced global numbering rides the DM global section (`_level_dof_layout`), transfers assemble as MPIAIJ (constrained coarse DOFs drop → reduced→reduced), parallel zero-column guard via `Pᵀ·1`. The legacy finest-only `set_custom_mg` path stays serial-only and raises loudly at np>1.

### Validation

| | iters | vs GAMG | notes |
|---|---|---|---|
| Scalar jump-coeff Poisson, 5-level (3 uniform + 2 SBR) | 3 FMG | 46 | |
| SolCx velocity block (η-jump 1e6), 3-level | 6 | ~198 | matches GAMG solution to 1.5e-9 |
| SolCx velocity block vs **native FMG** | 6 | (native 5) | ties native FMG; value = works where native FMG cannot (non-nested / adapted) |
| Parallel np=1/2/4 (Poisson + Stokes) | 4 / 6 | | matches GAMG and each other across rank counts |

## Out of scope (later)

Non-nested custom-P **in parallel** (cross-rank point location) — serial-only. Layer-2 adapt-on-top adapter is a separate follow-on. 

Design: `docs/developer/design/GENERALIZED_FMG_HIERARCHY_AND_ADAPT.md`


lmoresi:
Thanks — Copilot's 4 comments are addressed on the stacked branch `feature/rotated-freeslip-bc` (#293), commit `3dc7b893`, since that branch carries `custom_mg.py` and is where the follow-on work lives:

- **`custom_mg.py:179` (int32 CSR indices)** — fixed: `_to_petsc_aij` now casts `indptr`/`indices` to `PETSc.IntType`, matching the PETSc build's integer width (avoids int32 overflow / mis-addressing on a 64-bit `PetscInt` build with large meshes).
- **`custom_mg.py:97` (explicit inverse)** — fixed: `rbf_prolongation` now uses `np.linalg.solve(M, B.T).T` instead of forming `np.linalg.inv(M)` (faster, more stable; `M` is symmetric so `B M⁻¹ = solve(M, Bᵀ)ᵀ`).
- **`test_1017:91` and `test_1017:110` (absolute iteration bounds)** — kept intentionally. The margins are large (custom MG ~6 iters vs GAMG ~198) and each bound is paired with the "custom MG beats GAMG" assertion, so they are low flake-risk and preserve a useful regression signal. For tests, these tight margin cases are actually better because they will trigger if (say) gamg improves and that's sometimes what we want to know.

Verified after the change: `test_1017` serial 5/5, `test_1017` parallel np=2 (4 passed, 1 skipped = the unrelated #291 constrained-parallel), and the dependent `test_1018` rotated 6/6.

lmoresi:
Follow-up: these two fixes are now applied **directly on this branch** (`feature/custom-mg-prolongation`, commit `b7d1c4fd`), not only via the stacked #293 — so they merge with #290 itself. (The identical patch on #293 will drop as a no-op when #293 rebases onto the merged #290.) Validated identically: test_1017 serial 5/5, parallel np=2 (4 passed, 1 skipped=#291), test_1018 6/6.

PR #293, opened 2026-07-01
Rotated strong free-slip BC + σ_nn dynamic-topography (parallel), free-surface hand-off

lmoresi:
## What

In-solver **rotated strong free-slip** for the Stokes saddle, its **σ_nn / CBF dynamic-topography** recovery, and a **MeshVariable hand-off to the free-surface machinery** — all validated in parallel.

`stokes.add_rotated_freeslip_bc(boundary, normal=None)` imposes `v·n̂ = 0` in a per-node (normal, tangential) frame (SVD constraint frame, dim-general). `stokes.boundary_normal_traction(boundary, mass="lumped")` returns σ_nn from the constraint reaction. `stokes.dynamic_topography(boundary, field)` writes `h = -(σ_nn - mean)/(Δρg)` onto a scalar surface field for the 3-number topography integrator.

## Stacking

⚠️ **Stacked on #290** (custom-mg-prolongation): the rotated FMG path uses `custom_mg.set_custom_fmg`. Based against `feature/custom-mg-prolongation` so the diff shows only the rotated-bc work — **retarget to `development` once #290 merges.**

## Highlights this series

- **Parallel np>1 "hang" fixed** — it was an asymmetric crash (global row index into a local RHS slice), not a hang. Same class fixed in the σ_nn read and the field write.
- **σ_nn recovery**: Cartesian reaction projected onto the boundary normal (`n̂·r_c`) — corner-correct; whole-boundary relL2 vs analytic SolCx σ_yy **0.43 → 0.042** (res48), converging to 0.030 (res96), corr 0.999.
- **Mass lumping** removes the Gibbs overshoot at a stress jump (monotone M-matrix de-smear) — the safe default for driving a free surface; also marginally more accurate and parallel-trivial (local division).
- **Parallel-safe** throughout: reaction via `globalToLocal`, boundary mass via coordinate-keyed allgather → byte-identical np=1/2/4.
- **Full stack verified**: custom geometric FMG × rotated × annulus × parallel converges (reason 2) and is partition-independent.

## Tests

- `tests/test_1018_rotated_freeslip.py` (serial, 6): box↔essential, annulus zero-leakage, FMG velocity block, σ_nn accuracy, lumped no-overshoot (total-variation), dynamic_topography field.
- `tests/parallel/test_1064_rotated_freeslip_parallel.py` (5): box + annulus + FMG-annulus + σ_nn + topography-field partition-independence. Passes at `-n 2` and `-n 4`.

## Not in this PR (queued)

Default the boundary normal to `mesh.boundary_normal`; 3D spherical validation; wiring into the fs4 free-surface driver.

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -120,8 +120,15 @@ class SolverBaseClass(uw_object):
 
         Notes
         -----
-        Single-field (scalar/vector) solvers only for now; the Stokes
-        velocity-block path is a separate increment. See
+        Supported both on single-field (scalar / vector) solvers — where the
+        prolongation is installed directly on the solver's ``PCMG`` — and on the
+        **Stokes velocity block** (the ``fieldsplit_velocity_`` sub-PC), where the
+        velocity sub-PC is only reachable once the monolithic Jacobian has been
+        assembled; the install there assembles the Jacobian, descends the
+        fieldsplit to the velocity sub-PC, and rebuilds it as a fresh ``PCMG``
+        driven by our ``P``. Injection happens at solve time (after
+        ``setFromOptions`` / nullspace attach) via
+        :func:`underworld3.utilities.custom_mg.inject_custom_mg`. See
         :mod:`underworld3.utilities.custom_mg`.
         """
         if kind not in ("barycentric", "rbf"):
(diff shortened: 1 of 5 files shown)


lmoresi:
Re-review addressed in `f4d217f9`:

- **DM vector-pool leaks (`:217, :258, :265, :409, :548`)** — fixed. Vectors that persist beyond the call (the returned `U`, the LU-branch `Uhat`, the null-space `pv`) now use `dm.createGlobalVec()`; the borrowed temporaries (`U0`, `F0`) are returned with `restoreGlobalVec()`; the transient rigid-rotation vector from `_rigid_rotation_global` is restored at every call site, and the transient duplicates in `_rotation_is_nullspace` are `destroy()`ed. No more pool growth over repeated / time-stepping solves.
- **Stale module docstring (`:9`)** — fixed: it now describes the actual default (iterative fieldsplit-Schur, custom-FMG/GAMG, LU opt-in) and the shared `boundary_flux` de-smear, not the original "direct LU, FMG later" note.
- **Q identity built via per-row `setValue` (`:166`)** — **deferred** (noted here). It's a one-time-per-solve setup and only matters at very large scale; I've left it as a follow-up rather than perturb the well-tested Q construction. Happy to do the `setDiagonal` optimisation in a separate change if you'd like it in this PR.

Verified after the fixes: `test_1018` serial 6/6; `test_1064` parallel 5/5 at `-n 2` and `-n 4`.

lmoresi:
I'm going to merge this in so it can be evaluated. Tests pass, but we don't know about speed, memory, 3D reliability, null-space treatment for this and the related rotated BC functionality. The topography part of the story is the final part of the jigsaw.

PR #294, opened 2026-07-01
General boundary_flux primitive: Consistent Boundary Flux for any solver (heat flux / Nusselt, traction)

lmoresi:
## What

This set of functionality goes hand in hand with the rotated bc toolkit. Once we have exact boundary conditions, we can use the consistent boundary flux methods to obtain (wait for it) consistent fluxes at the boundaries. This is the most accurate was to determine topography, heat flux, etc. It is not compatible with the penalty approaches (very low accuracy).

Adds `solver.boundary_flux(boundary, mass="lumped", remove_mean=False, normal=None)` and
`solver.boundary_flux_field(boundary, field, scale=1.0, ...)` to `SolverBaseClass`, so the
Consistent Boundary Flux (CBF, Gresho et al.) recovery is a **general capability of any
solver**, not a Stokes-only / topography-specific one.

The essential-BC reaction — the assembled interior FEM residual at a Dirichlet node — is the
consistent nodal flux; de-smearing it with the boundary mass gives a pointwise surface flux:

- **scalar diffusion / advection-diffusion** → surface heat flux `-k ∂T/∂n` (boundary **mean = Nusselt number**)
- **Stokes** → boundary traction `σ·n̂` (`σ_nn` with `normal`)

## Design

- `SolverBaseClass._assemble_volume_reaction()`: field-structure-agnostic (Stokes `self.fields`,
  else the single `Unknowns.u`) volume FEM residual (PETSc's rock-solid volume integration).
- `utilities/boundary_flux.py`: the equation-independent parts — boundary-node gathering, the
  lumped/consistent boundary-mass de-smear (`remove_mean` flag: keep the mean for a physical flux
  = Nusselt, remove it for a gauge-free field = dynamic topography), and the field write.

`boundary_flux` is the **primitive**; `boundary_normal_traction` / `dynamic_topography` (on the
rotated-free-slip branch #293) become thin readings of it. `remove_mean` defaults **False** — the
mean flux is physical.

## Parallel correctness (the hard part)

A partition that CUTS the flux boundary needs the nodal reaction assembled across ranks. Rather
than hand-roll `localToGlobal(ADD)` (which was wrong at cut nodes), the DM has overlap=0 so PETSc's
volume residual gives each rank its OWNED-cell partial; the complete reaction is assembled by
**summing the partials across ranks by coordinate** — the same rock-solid gather used for the
boundary mass. Verified against the rock-solid volume-integral identity
`total_flux = ∫_Ω ∇T·∇w`, which is byte-identical at np=1/2/4.

## Validation

Harmonic Poisson with a known analytic flux: serial corr **1.0000**, relL2 **2e-4** (converging
8e-4→2e-4→4e-4 at res 24/48/96), mean flux = Nusselt `2/sinh π`; `BdIntegral` of the flux field
byte-identical np=1/2/4.

- `tests/test_1019_boundary_flux.py` (serial)
- `tests/parallel/test_1065_boundary_flux_parallel.py` — passes at `-n 2` and `-n 4` (boundary cut)

## Follow-up

Once this lands, #293 rebases `boundary_normal_traction` / `dynamic_topography` onto `boundary_flux`.

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -1836,6 +1836,10 @@ class SolverBaseClass(uw_object):
         cdef Vec xvec
         cdef Vec fvec
         cdef DM _time_dm_reaction
+        cdef PetscFormKey key
+        cdef IS ccell_is
+        cdef PetscReal residual_time = 0.0
+        cdef PetscReal implicit_form_time = <PetscReal>-1.7976931348623157e308
 
         self._build(verbose, False, None)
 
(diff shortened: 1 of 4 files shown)


lmoresi:
Updated after copilot reviews

Issue #279, opened 2026-06-24
Enforce the ND<->units boundary: evaluate/global_evaluate should reject or nondimensionalise UnitAwareArray input


lmoresi:
## Background
Per the [ND<->units boundary contract](https://github.com/underworldcode/underworld3/blob/development/docs/developer/design/ND_UNITS_BOUNDARY_CONTRACT.md) (added with #267/#278): the DM/PETSc/solver layer is non-dimensional-only; units live above it in the MeshVariable API.

Today that boundary is **convention, not enforced**, and there is a silent trap: `uw.function.evaluate` / `global_evaluate` strip the `UnitAwareArray` subclass with `np.array()` but **keep the numeric value**. So a *dimensional* coordinate array (`.coords` ≈ 1e9 m) passed in is kept as 1e9 and located against a 0..1000 ND DM — **no exception, wrong location, wrong answer**.

This is exactly how #267 slipped in (the SL trace-back fed dimensional `.coords` into the ND DM). #277 fixed that one call site, but the same class of mistake can recur anywhere a developer reaches for `.coords`/`.array` instead of `.coords_nd`/`.data`.

## Proposal
Make the ND entry points **guard** the boundary instead of silently stripping:
- `uw.function.evaluate` / `global_evaluate`: if handed a `UnitAwareArray` for the query coordinates (or a unit-bearing expression where ND is expected), either
  - **reject** it with a clear error ('pass non-dimensional coordinates — use `mesh.X.coords_nd` / `var.coords_nd`, or `uw.non_dimensionalise(...)`'), or
  - **correctly non-dimensionalise** it (not just strip the label) before locating.
- Consider the same guard on the DM coordinate setters / any other ND entry point.

Rejecting is the safer default (forces the caller to be explicit); auto-nondimensionalising is more convenient but hides the boundary. Prefer reject-with-actionable-message unless there's a strong ergonomic case.

## Acceptance
A regression test that passing a dimensional `UnitAwareArray` of coordinates to `evaluate`/`global_evaluate` no longer silently mislocates (errors clearly, or returns the correctly-located result).

Tracked from the #267 boundary-contract discussion.

Underworld development team with AI support from Claude Code

lmoresi:
[...]

PR #295, opened 2026-07-02
fix(units): UWQuantity handles offset temperature units (degC/degF)

lmoresi:
## Summary

`UWQuantity.__init__` built its Pint quantity as `value * unit`, which raises `OffsetUnitCalculusError` for offset (non-multiplicative) units. So `uw.quantity(20, "degC")` **crashed at construction**, and any offset temperature that reached non-dimensionalisation failed **silently** (a bare `except` fell back to the dimensional value — the #279 silent-boundary class).

## Design: offset units live only at the user boundary

- **Input** (`__init__`): construct via the registry `Quantity(value, unit)` constructor, then normalise an offset temperature (degC/degF) to absolute **kelvin**. The stored quantity is always multiplicative, so no internal code (arithmetic, nd, JIT) ever meets a non-multiplicative unit. Offset detection uses Pint's own `_is_multiplicative` flag, which correctly leaves a temperature *difference* (`delta_degC`) untouched.
- **Output** (`.to("degC")`): a display result keeps its offset unit via a non-normalising construction path.

## Verified

| behaviour | before | after |
|---|---|---|
| `uw.quantity(20,'degC')` | ❌ OffsetUnitCalculusError | ✅ stored 293.15 K |
| `.to('degC')` round-trip | — | ✅ 20 °C |
| `q * 2` on a degC input | ❌ raises | ✅ 586.3 K |
| `nd(500 degC)` | ❌ silent-wrong | ✅ 0.77315 (== nd(773.15 K)) |
| `delta_degC` (difference) | — | ✅ preserved, not normalised |

New `tests/test_0757_uwquantity_offset_units.py` (8 cases). Existing units tests (`test_0756`, `test_1011`) unaffected — 14/14 green locally.

Companion to the #263 thermal-convection units tutorial (which previously side-stepped this by staying in Kelvin). Relates to #279 (ND↔units boundary enforcement).

### DIFF
--- a/src/underworld3/function/quantities.py
+++ b/src/underworld3/function/quantities.py
@@ -53,7 +53,8 @@ class UWQuantity:
     def __init__(
         self,
         value: Union[float, int, np.ndarray],
-        units: Optional[str] = None
+        units: Optional[str] = None,
+        _normalise_offset: bool = True,
     ):
         """
         Initialize a UWQuantity.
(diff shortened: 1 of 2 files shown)


lmoresi:
Credited @ss2098 as co-author on this commit — the offset-unit gap surfaced directly from their thermal-convection units-tutorial work (#263, and the #261/#262/#268 series), where they worked around Pint's offset limitation by staying in Kelvin. This PR removes that limitation at the source so degC/degF can be used as inputs directly.

(Co-authored-by trailer added; contribution will attribute to their GitHub profile.)

PR #298, opened 2026-07-02
Rotated strong free-slip inside the nonlinear Stokes solve

lmoresi:
## Rotated strong free-slip inside the nonlinear Stokes solve

Rotated strong free-slip (`add_rotated_freeslip_bc`, per-node `v·n̂ = 0` in a
rotated frame) now works as an ordinary boundary condition **inside** the
nonlinear SNES iteration, so nonlinear rheology, warm-start and time-stepping
"just work" with it — exactly as for native axis-aligned essential BCs. This is
the gate for promoting rotated free-slip to the **default** free-slip BC
(reserving Nitsche / penalty for BCs that must *evolve* in time).

Previously the rotated path was a single LINEAR solve: it assembled `J(0)`,
`F(0)` once, rotated the operator + RHS, imposed `v_n=0` on the rotated normal
rows, and did one KSP solve — bypassing `snes.solve()`. Exact for Newtonian
Stokes, but a nonlinear model solved with defaults silently returned one
linearisation from `u=0`.

### What changed

**Fail-fast probe.** A numerical nonlinearity probe (`_residual_is_nonlinear`)
assembles the Jacobian at two velocity states and compares (a symbolic test on
`F1.sym` cannot see it — the viscosity's strain-rate dependence is a
JIT-substituted placeholder, decoupled from the velocity gradient in the symbolic
flux). Landed first, on its own, as interim safety.

**Nonlinear integration.** A manual outer nonlinear loop
(`solve_rotated_freeslip_nonlinear`): each iteration computes the Cartesian
residual `F(u)` and Jacobian `J(u)` via the DM callbacks, rotates them
(`F̂ = Q F`, `Ĵ = Q J Qᵀ`), zeros the constrained normal rows, and solves the
increment with the already-validated self-contained rotated fieldsplit-Schur KSP
(custom geometric FMG / GAMG velocity block + rotated coupled null space). That
KSP is reused because the DM-coupled fieldsplit cannot precondition a DM-less
`ptap`'d operator (SUBPC_ERROR). A backtracking line search, a step-norm
convergence exit (fast, correct warm start), a divergence warning if it does not
converge, and exact re-imposition of `v_n=0` each iterate complete the loop. `Q`,
the rotated FMG prolongation and the null space depend only on geometry, so they
are built once and reused across iterations.

**Tangent controls (transparent to the rotated constraint).** The driver honours
`solver.consistent_jacobian` exactly as the standard solve path does, so the
rotated constraint does not alter the tangent:

- **`consistent_jacobian=True` — consistent Newton** (recommended for smooth
  nonlinear rheologies): the rotated solve reaches the same solution in the same
  small iteration count as the native *essential* Newton solve on the
  unconstrained problem — the constraint is transparent to convergence.
- **`"continuation"` — staged Picard→Newton** (α=0 phase to `newton_switch_rtol`,
  then α=1), for robustness far from the solution; `solve(picard=N)` sets the
  length of the warm-up phase.
- **default frozen tangent** — the globally-robust defect-correction path.

A `picard>0` warm-up requested with the pure Newton tangent raises a clear error
pointing to `"continuation"` rather than being silently ignored. (This also
surfaced and fixed a pre-existing bug: `SNES_Stokes.solve` dropped `picard` for
plain — non-viscoelastic — Stokes.)

Dispatch: a linear model takes the unchanged one-shot linear solve; a nonlinear
model takes the driver. σ_nn recovery stashes the converged Cartesian residual as
the constraint reaction (identical to `A·u−b` for the affine linear case).

### Validation

Power-law rheology (η = ε̇_II^(1/n−1)) with the consistent Newton tangent, against
the native essential Newton solve:

| case | result |
|---|---|
| box, GAMG velocity block | matches essential Newton to machine precision (`~1e-13`); wall-normal leakage machine-zero on every wall |
| box, custom **geometric FMG** | converges via geometric FMG — **5–6 outer fieldsplit iterations per solve**, angle-independent; matches essential |
| annulus, per-node radial free-slip (genuinely-rotated frame) | radial leakage `~2e-16`, rigid-rotation gauge removed |
| warm-start (2-step time loop) | ≤ 3 iterations |
| parallel np = 1 / 2 / 4 | byte-identical velocity `L2`, partition-independent |

### Tests

- `test_1018` (serial): nonlinear coverage — Newton tangent vs essential,
  continuation (incl. Picard-phase length), Picard-warm-up guard, warm-start,
  geometric FMG, and the genuinely-rotated annulus. The cases share one essential
  reference and use small meshes.
- `test_1064` (parallel np2/4): nonlinear partition-independence.
- The linear SolCx tests were run at a lower resolution — SolCx breakage is
  catastrophic (correlation collapses, error blows up to O(1)) rather than
  gradual, so a coarse mesh still catches it — bringing the suite from ~9 min to
  ~1 min.

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -5208,8 +5208,11 @@ class SNES_Stokes_SaddlePt(SolverBaseClass):
         # partition-independent enough for the norm comparison.
         rs, re = U1.getOwnershipRange()
         idx = np.arange(rs, re, dtype=float)
-        U1.setArray(0.1 * np.sin(0.7 * idx + 0.3))
-        U2.setArray(0.1 * np.sin(1.3 * idx + 1.1))
+        # write IN PLACE into each vec's PETSc-owned array (getArray returns a writable
+        # view). setArray with a fresh numpy temporary on a POOLED getGlobalVec is risky:
+        # its storage lifetime/pool reuse is not guaranteed for the later computeJacobian.
+        a1 = U1.getArray(); a1[:] = 0.1 * np.sin(0.7 * idx + 0.3)
+        a2 = U2.getArray(); a2[:] = 0.1 * np.sin(1.3 * idx + 1.1)
         J1 = J.copy(); J2 = J.copy()
         try:
             snes.computeJacobian(U1, J1)
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -7908,25 +7908,26 @@ class SNES_Stokes_SaddlePt(SolverBaseClass):
             self.dm.setAuxiliaryVec(self.mesh.lvec, None)
             self._update_constants()
 
-            # Two paths:
-            #  * LINEAR model, zero initial guess, no Picard warmup → the validated
-            #    one-shot solve (assemble J(0),F(0); rotate; single self-contained
-            #    fieldsplit KSP). Fast, and the linear solution is exact so warm-start
-            #    / Picard would add nothing — hence the extra guards on this branch.
-            #  * otherwise (nonlinear rheology, or an explicitly-requested warm-start
-            #    / Picard) → the manual outer Newton/Picard loop that rotates F(u),
-            #    J(u) and the v_n=0 constraint every iteration. A nonlinear model is
-            #    detected by probing the assembled Jacobian for solution-dependence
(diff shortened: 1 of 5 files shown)


lmoresi:
This PR unifies the rotated bcs with the generalised FMG in stokes solvers. These both use the same interjections in the SNES pathway - building restrictions / prolongations AND rotations and interspersing them where they are needed. This should allow full MG, on hierarchically adapted meshes with non-coordinate aligned boundary constraints. Among things that have not been tested but would be nice to know if they work: free surface evolution with all that stuff; non-nested grids with rotated bcs. What has been demonstrated: you can now run an adapted fault model with free slip (rotated) conditions using FMG. Some progress today.

PR #313, opened 2026-07-04
fix(swarm): Track-0 stale-cache and migration-semantics fixes (BF-02/05/06/07/08 + #286)

### LINKED ISSUE #286: AMR: swarm migration after mesh.adapt() raises IndexError -- k-d tree rebuilt before nav_coords refresh
tests/test_0810_amr_swarm_migration_regression.py::test_swarm_migration_after_adapt_does_not_raise fails with:

IndexError: index 91 is out of bounds for axis 0 with size 80

Cause: in nuke_coords_and_rebuild() (discretisation_mesh.py), the k-d tree is rebuilt at line 2346, but self._nav_coords isn't refreshed until line ~2401 -- later in the same function. So the rebuild runs against a stale nav_coords, sized for the old mesh, while nav_dm already reflects the new one. Confirmed by instrumenting a live run: at rebuild time, _nav_coords.shape=(80, 2) vs. nav_dm already at 127 points.

Checked the other two caches that also read nav_coords (faces_outer/inner_control_points from #135, and boundary_face_control_points_kdtree/_sign) -- both are invalidated but rebuilt lazily on next access, after this function has already finished, so they are not affected by this ordering issue.

### LINKED ISSUE #289: swarm.advection() leaves proxy mesh variables stale (frozen material) — auto-update hook orphaned in deprecated access path
## Summary

`swarm.advection()` moves the particles but **does not mark the swarm-variable proxy
mesh variables stale**, so the lazy `.sym` accessor never rebuilds them. The Stokes/
energy solver on the *next* step therefore reads a **frozen** proxy (composition,
density, viscosity) projected from the *previous* particle positions. The material is
dynamically stationary even though the particles are advecting — silently producing
wrong physics (e.g. a buoyant plume that never rises; a Rayleigh–Taylor interface that
never evolves).

This affects the **standard advect-then-use-`.sym`** pattern used by the shipped
examples `docs/examples/fluid_mechanics/intermediate/Ex_Stokes_Swarm_RT_Cartesian.py`
and `docs/examples/convection/intermediate/Ex_Convection_Cartesian_ThermoChem.py`,
which call `swarm.advection(...)` and rely on `material.sym` being current on the next
solve, with **no** explicit `_update()`.

## Minimal reproducer

```python
import underworld3 as uw, numpy as np, sympy
mesh = uw.meshing.UnstructuredSimplexBox(minCoords=(0,0), maxCoords=(1,0.25), cellSize=1/48, qdegree=3)
v = uw.discretisation.MeshVariable("V", mesh, vtype=uw.VarType.VECTOR, degree=2)
swarm = uw.swarm.Swarm(mesh=mesh)
mat = uw.swarm.IndexSwarmVariable("M", swarm, indices=2, proxy_degree=1, proxy_continuous=True)
swarm.populate(fill_param=4)
mat.data[...] = 0
pc = swarm.data
mat.data[(pc[:,0]-0.5)**2 + (pc[:,1]-0.107)**2 < 0.0179**2, 0] = 1
v.data[:,1] = 0.01                                   # uniform upward velocity

def particle_cy(): m = mat.data[:,0] > 0.5; return swarm.data[m,1].mean()
def proxy_cy():
    pv = np.asarray(uw.function.evaluate(mat.sym[1], v.coords)).reshape(-1)
    w = np.clip(pv,0,None); return (w*v.coords[:,1]).sum()/w.sum()

print("before:", particle_cy(), proxy_cy(), "stale=", mat._proxy_stale)
swarm.advection(v.sym, 2.0, order=2, corrector=True)
print("after :", particle_cy(), "_proxy_stale=", mat._proxy_stale)
print("proxy_cy (lazy):", proxy_cy())
mat._update()
print("proxy_cy after _update():", proxy_cy())
```

Output:
```
before: 0.1071 0.1070 stale= False
after : 0.1271 _proxy_stale= False     # particles moved 0.02 up; proxy NOT marked stale
proxy_cy (lazy): 0.1070                 # FROZEN at the initial position
proxy_cy after _update(): 0.1271        # explicit _update() fixes it
```

The particle centroid moves (0.107 → 0.127); the proxy stays at 0.107; `_proxy_stale`
is `False` after advection, so `.sym` returns the stale projection.

## Root cause

The proxy auto-refresh contract is wired into the **deprecated** `swarm.access()`
context manager: `access.__exit__` calls `var._update()` for all vars when the particle
coordinates were writeable —
`src/underworld3/swarms/pic_swarm.py:994–1001` ("if swarm migrated, update all").
But after the **`c52201b8` remesh field-transfer redesign**, `_proxy_stale=True` is set
only by the **remesh/adapt** path (`_remesh_reinit_callback`,
`src/underworld3/swarm.py:1057`), not by ordinary advection. The lazy `.sym` gate
(`if self._proxy_stale: _update_proxy_variables()`, `swarm.py:2197`) is therefore never
triggered by `advection()`. Net: the staleness hook that *advection* depends on lives in
the now-deprecated `access` machinery and is effectively orphaned for the advection code
path (proven above: `_proxy_stale` stays `False`).

## Impact

- **Silent correctness failure** in any swarm-advection model whose buoyancy/viscosity/
  composition is read from a proxy (`material.sym[i]`) — the dominant UW3 idiom.
- The looks-plausible failure mode is a model that reaches a spurious "steady state"
  (the material is frozen) — easy to misread as physics.
- Affects shipped examples (RT, ThermoChem) and the documented pattern.

## Workaround

Call `material._update()` (the `SwarmVariable`/`IndexSwarmVariable`, not the swarm)
immediately after every `swarm.advection(...)`:
```python
swarm.advection(v.sym, dt, order=2, corrector=True)
material._update()        # refresh proxy so the next solve sees the moved material
```

## Suggested fix

Mark all proxies stale (or eagerly `_update()` them) **inside `swarm.advection()`**
after the particles are migrated, independent of the deprecated `access` context
manager — e.g. at the end of `advection()`:
```python
for var in self.vars.values():
    var._update()      # or: if var._proxy: var._proxy_stale = True
```
This re-establishes the auto-update contract the examples rely on, and decouples it from
the deprecated `access().__exit__` path. A regression test that advects a uniform
velocity and asserts the proxy centroid tracks the particle centroid would guard it.

---
Found while reproducing the Crameri (2012) free-surface plume benchmark; UW3 build on
`feature/integrate-surface-submesh` (commit ~`a01e79f3`).

lmoresi:
Track-0 remediation (2026-07 quality audit, `docs/reviews/2026-07/REMEDIATION-WORKLIST.md`):
BF-02 (SWARM-01/02/17), BF-05 (SWARM-04), BF-06 (SWARM-07), BF-07 (SWARM-03),
BF-08 (LE-03 = SWARM-05, issue #215 Bug 3), plus the freshly diagnosed #286.

Closes #289
Closes #286

### BF-02 — stale caches after no-move migrate / particle addition / populate (SWARM-01/02/17)
`Swarm.migrate()` early-returned when no particle was globally unclaimed without
invalidating the canonical `.data` caches, the particle kd-tree, or the proxy
staleness flags. Reproduced (serial): after `add_particles_with_global_coordinates`
the DMSwarm held 974 particles while every cached `.data` still had 972 rows; after
an in-place coordinate mirror + `migrate()`, `rbf_interpolate` returned garbage from
the poisoned no-copy kd-tree; after `populate()` a pre-existing cache stayed at 0
rows. **This is also the root cause of issue #289**: `swarm.advection()` ends with a
no-move `migrate()` in serial, so `_proxy_stale` stayed `False` and the material
proxy froze at the launch positions.

Fix: invalidate before the early return; explicit invalidation in
`add_particles_with_global_coordinates` (mirroring the #216 fix in
`add_particles_with_coordinates`) and in `populate()` (SWARM-17 verified at
runtime, folded in).

**Collateral defect exposed by this fix** (same #216 stale-cache class):
`Swarm.apply_snapshot_payload()` wrote restored variable data into a detached
`np.asarray` view of the canonical cache — the DMSwarm field never received the
restored values, and the restore only "worked" because the SWARM-01 bug preserved
the stale canonical copy. With the invalidation fixed, `test_0007_snapshot_inmemory`
continuation tests caught it (restored material became zeros after the first
advection). Fixed by writing through the canonical array so the PETSc pack fires.

### #286 — AMR: kd-tree rebuilt against stale `_nav_coords`
In `nuke_coords_and_rebuild()` the point-location kd-tree was rebuilt before
`self._nav_coords` was refreshed, so the rebuild indexed an old-sized coordinate
array with new-mesh point ranges: `IndexError: index 91 is out of bounds for axis
0 with size 80` (reproduced via `tests/test_0810_amr_swarm_migration_regression.py`).
Fix: refresh `_nav_coords` (volume and manifold branches) before
`_build_kd_tree_index()`.

### BF-05 — writes inside `migration_disabled()` silently discarded (SWARM-04)
Both PETSc-sync callbacks early-returned while `_migration_disabled` was set and
nothing ever re-packed; `migrate()` then read the stale PETSc coordinates and its
trailing invalidation destroyed the only copy of the writes. Reproduced (serial):
after writing 42.0 inside the context, the PETSc field still held 0.0.
Fix: separate "suppress migration" from "suppress PETSc sync" — writes are recorded
and flushed to the DMSwarm by `Swarm._flush_pending_petsc_sync()` at context exit
(and defensively at `migrate()` entry); only the migration itself is deferred
(default) or skipped (`disable=True`). The vacuous assertions in
`tests/parallel/test_0755_swarm_global_stats.py` (the perturbation never reached the
DMSwarm) now assert that it does.

### BF-06 — empty/starved ranks: silent zeros or hard crash (SWARM-07)
`rbf_interpolate` silently returned zeros for ranks with <= 1 particles and
`_rbf_to_meshVar` wrote them into the proxy; `IndexSwarmVariable`'s proxy update
reached unguarded `KDTree` construction on a 0-particle rank —
`IndexError: Out of bounds on buffer access` (reproduced) inside a collective
update, i.e. an MPI abort/hang.
Fix: starved ranks leave their proxy nodal values untouched and warn (never silent
zeros); guard the KDTree/nearest-neighbour machinery on `local_size <= 1` while
keeping every rank inside the same sequence of collective reads/writes/contexts
(MeshVariable syncs are collective — a naive early-return would deadlock).
`IndexSwarmVariable` gains a proper `_update_proxy_if_stale()` override (its
proxies live in `_meshLevelSetVars`, so the base implementation was a no-op for it).

The `update_type=0` level-set projection was also restructured to compute into a
local buffer and issue exactly ONE MeshVariable write per level set per rank
(serially bit-identical): the previous formulation issued a data-dependent number
of writes whose deferred collective syncs mismatched across ranks — the np4
starved-rank test deadlocked on it (each write's ghost sync is collective).

### BF-07 — modern coordinate writes never migrate (SWARM-03)
The class docstring promised automatic migration, but `swarm.coords` /
`swarm._particle_coordinates.data` writes only packed to PETSc; only the
*deprecated* `points` path migrated. Fix (maintainer-approved shape): coordinate
writes mark `swarm._needs_migration`; the collective `migrate()` runs DEFERRED at
the next collective point — migration-control context exit or solve entry — never
per-write (uneven writes would deadlock). The flag is combined across ranks with a
MAX reduction before acting, so uneven writes are safe. Docstrings updated to state
the actual contract.

### BF-08 — stale proxy consumed by solvers (LE-03 = SWARM-05, issue #215 Bug 3)
Proxy refresh only fired via the lazy `.sym` accessor, but solvers pull the proxy
DM directly through `mesh.update_lvec()`. Reproduced (serial): write
`material.data`, re-solve a captured `Projection` — the solve consumed the old
values. Fix (maintainer-approved shape): a single eager, collective refresh at
solve entry — `Swarm._sync_before_assembly()` invoked from `Mesh.update_lvec()`
before its staleness check (the refresh itself is what sets `_stale_lvec`). It
also performs BF-07's deferred migration. Flag-guarded: repeated calls are no-ops,
so nothing is re-interpolated per access (the `test_0006` memory-leak constraint).
The `TODO(BUG)` at the old `swarm.py:1075` is resolved and removed.
Together with BF-02 this closes #289 (verified with the issue reproducer: the
proxy centroid now tracks the advecting particle centroid).

## Tests

- `tests/test_0113_swarm_stale_cache_regression.py` (new, level_1/tier_a):
  SWARM-01 (both `migrate=` variants), SWARM-17, SWARM-02 (kd-tree invalidation
  contract + behavioral mirror probe), the #289 reproducer as a regression test,
  and the #215-Bug-3 solve-freshness test.
- `tests/parallel/test_0756_swarm_migration_semantics.py` (new): np2
  migration-suppressed write survival, np2 deferred migration at context exit and
  at solve entry, np4 starved-rank proxy test (all particles on one rank).
- `tests/parallel/test_0755_swarm_global_stats.py`: repaired the vacuous
  perturbation block.
- `tests/test_0810_amr_swarm_migration_regression.py` (#286) now passes.

## Gate results

All runs from the fix worktree, env `amr-dev` (AMR-enabled PETSc build, needed
for the #286 test), after `./uw build`.

- **Serial `level_1 and tier_a`**: BEFORE = 295 passed, 6 failed -- the 6 failures
  are exactly the new regression tests, each demonstrating its bug on the pristine
  baseline. AFTER = **301 passed, 0 failed** (1 skipped, 4 xfailed, 1 xpassed --
  unchanged from baseline).
- **np2** (`mpirun -np 2`, `test_0755` + `test_0756` + `test_0760` + `test_0765`):
  **19 passed**, 1 skipped (the np4-only test).
- **np4** (`mpirun -np 4`, `test_0755` + `test_0756` + `test_0765`): **17 passed**.
- **`test_0006_memory_leak`**: passed (the BF-08 gate -- the solve-entry refresh is
  flag-guarded, nothing re-interpolates per access).
- **`test_0810_amr_swarm_migration_regression`** (#286): failed with the reported
  IndexError before the fix; passes after.
- **Issue #289 reproducer**: particle centroid 0.1068 -> 0.1268, proxy centroid
  0.1071 -> **0.1270** (was frozen at 0.1071); also enshrined as a regression test.

**Pre-existing np4 defect found while gating (NOT addressed here):**
`tests/parallel/test_0760_swarm_cache_migration.py` hangs at `mpirun -np 4` on the
**pristine baseline** (verified by stashing this branch's src changes and
rebuilding): with rank-biased coordinates, `global_evaluate` leaves some ranks with
zero interior points, and those ranks diverge inside the `DMInterpolation`
machinery in `petsc_interpolate` while the others block in its collectives. It
passes at np2 (CI runs `--p 2`). Per-rank stack traces confirm all of this branch's
collective hooks enter and exit symmetrically before the hang. This should be
tracked as its own issue (parallel-evaluation empty-rank asymmetry, related to the
audit's SWARM-15/16 provisional findings).

### DIFF
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -2930,11 +2930,38 @@ class Mesh(Stateful, uw_object):
         return project(numpy.asarray(coords, dtype=float))
 
     @timing.routine_timer_decorator
-    def update_lvec(self):
+    def update_lvec(self, swarm_sync=True):
         """
         This method creates and/or updates the mesh variable local vector.
         If the local vector is already up to date, this method will do nothing.
+
+        ``swarm_sync=False`` skips the swarm-dependency hook below. It MUST
+        be passed at call sites that only a SUBSET of ranks reach (e.g. the
+        refresh calls inside ``petsc_interpolate`` — ranks with zero interior
+        points skip that function entirely): the hook performs collective
+        reductions, so running it on a subset of ranks deadlocks. Such sites
+        rely on an earlier all-ranks ``update_lvec()`` for freshness, exactly
+        as they already do for the collective ``globalToLocal`` below.
         """
+
+        # Swarm dependencies first (issues #215 Bug 3 / #289): run any
+        # deferred particle migration and refresh stale swarm-variable
+        # proxies BEFORE the staleness check below — the refresh writes
+        # proxy MeshVariable data, which is what sets _stale_lvec. Solvers
+        # read the proxy DM directly, so the lazy `.sym` refresh alone
+        # cannot guarantee freshness at assembly. Collective; flag-guarded
+        # no-op when nothing changed. Ordered deterministically so all
+        # ranks act on swarms in the same sequence.
+        if swarm_sync and not getattr(self, "_swarm_sync_in_progress", False):
+            self._swarm_sync_in_progress = True
+            try:
+                for swarm in sorted(
+                    self._registered_swarms,
+                    key=lambda s: s._instance_number,
+                ):
+                    swarm._sync_before_assembly()
+            finally:
+                self._swarm_sync_in_progress = False
 
         if self._stale_lvec:
             if not self._lvec:
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -2345,6 +2345,29 @@ class Mesh(Stateful, uw_object):
                 flush=True,
             )
 
+        # The navigation coords (used to build the kd-tree control points and
+        # for point location) were captured as a reference to the ORIGINAL
+        # coords in __init__ and are not updated by the DM rebuild above, so
+        # on an adapted/deformed mesh they still describe the old geometry.
+        # They MUST be refreshed BEFORE _build_kd_tree_index() runs: that
+        # rebuild indexes _nav_coords with the NEW nav-DM point range, and a
+        # stale (old-sized) array raises IndexError when the mesh grew — and
+        # silently mislocates points otherwise (issue #286).
+        if getattr(self, "_nav_dm", None) is None:
+            self._nav_coords = numpy.asarray(
+                self.dm.getCoordinatesLocal().array
+            ).reshape(-1, self.cdim)
+        else:
+            # manifold mesh: the nav clone carries its own (ghosted) coords;
+            # refresh them from the rebuilt main DM where possible.
+            try:
+                self._nav_dm.setCoordinatesLocal(self.dm.getCoordinatesLocal())
+                self._nav_coords = numpy.asarray(
+                    self._nav_dm.getCoordinatesLocal().array
+                ).reshape(-1, self.cdim)
+            except Exception:
+                pass
+
         self._index = None
         self._build_kd_tree_index()
 
@@ -2395,25 +2418,8 @@ class Mesh(Stateful, uw_object):
         self.boundary_face_control_points_sign = None
         self._domain_radius_squared = float("inf")
 
-        # The navigation coords (used to build those control points and for
-        # point location) were captured as a reference to the ORIGINAL coords
-        # in __init__ and never refreshed here, so on a volume mesh they stayed
-        # at the undeformed geometry — the real reason a bulged-out region read
-        # as exterior. Re-point them at the current (deformed) DM coordinates.
-        if getattr(self, "_nav_dm", None) is None:
-            self._nav_coords = numpy.asarray(
-                self.dm.getCoordinatesLocal().array
-            ).reshape(-1, self.cdim)
-        else:
-            # manifold mesh: the nav clone carries its own (ghosted) coords;
-            # refresh them from the rebuilt main DM where possible.
-            try:
-                self._nav_dm.setCoordinatesLocal(self.dm.getCoordinatesLocal())
-                self._nav_coords = numpy.asarray(
-                    self._nav_dm.getCoordinatesLocal().array
-                ).reshape(-1, self.cdim)
-            except Exception:
-                pass
+        # NB: the _nav_coords refresh for the deformed/adapted geometry happens
+        # ABOVE, before _build_kd_tree_index() — see the issue #286 note there.
 
         # BUGFIX(#130): recompute the DOF coordinate cache for every
         # already-registered variable. Variables created before this
--- a/src/underworld3/function/_function.pyx
+++ b/src/underworld3/function/_function.pyx
@@ -1279,7 +1279,11 @@ def petsc_interpolate(   expr,
 
         if cached_info is not None:
             # CACHE HIT - Fast path. Evaluate using cached structure
-            mesh.update_lvec()  # Ensure fresh values
+            # swarm_sync=False: petsc_interpolate is reached by only the
+            # ranks that hold interior points — the swarm-dependency hook
+            # does collective reductions and must not run on a subset.
+            # Freshness comes from the all-ranks update_lvec() in evaluate().
+            mesh.update_lvec(swarm_sync=False)  # Ensure fresh values
             cached_info.evaluate(mesh, outarray)
 
         else:
@@ -1320,7 +1324,9 @@ def petsc_interpolate(   expr,
             mesh._dminterpolation_cache.store_structure(coords, dofcount, cached_info)
 
             # Evaluate
-            mesh.update_lvec()
+            # swarm_sync=False: see the cache-hit branch above — only a
+            # subset of ranks reaches petsc_interpolate.
+            mesh.update_lvec(swarm_sync=False)
             cached_info.evaluate(mesh, outarray)
         # === END CACHING ===
 
--- a/src/underworld3/swarm.py
+++ b/src/underworld3/swarm.py
@@ -509,6 +509,15 @@ class SwarmVariable(DimensionalityMixin, MathematicalMixin, Stateful, uw_object)
             # STEP 1: Sync to PETSc using established method with correct shape
             self.pack_raw_data_to_petsc(canonical_array, sync=True)
 
+            # Coordinate writes may strand particles on the wrong rank. Mark
+            # the swarm for DEFERRED migration — migrate() itself is
+            # collective and must not run from a per-write callback (ranks
+            # write unevenly → deadlock). The migration happens at the next
+            # collective point: migration-control context exit or solve entry
+            # (SWARM-03; the class docstring's automatic-migration promise).
+            if getattr(self.swarm, "_coord_var", None) is self:
+                self.swarm._needs_migration = True
+
             # STEP 2: Handle variable-specific updates (like IndexSwarmVariable proxy marking)
             if hasattr(self, "_on_data_changed"):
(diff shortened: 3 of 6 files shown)


lmoresi:
CI failure in `test_0726_bare_variable_composition_137.py` (5 tests, "Swarm for variable 'C_137' has been garbage collected") is fixed in 34437577.

**Root cause**: the test's module fixture drops the parent swarm and keeps only the SwarmVariable. On development this silently worked because `populate()` never marked the proxy stale (the SWARM-17 hole this PR fixes) — `.sym` returned the last projection without dereferencing the dead parent. With the invalidation fixed, the lazy refresh hit the (pre-existing, unchanged) weakref lifetime guard.

**Fix by contract, not band-aid**: variables keep holding their swarm weakly — a strong back-reference would cycle with the swarm's own strong `_coord_var`/`_X0` members and defer DMSwarm destruction from refcount-immediate to gc time, re-opening the transient-evaluation-swarm leak (guarded by `test_0006_memory_leak`). A variable that outlives its swarm is now formally a symbolic *fossil*: the stale-refresh path warns that the proxy retains its last projection and cannot be refreshed; `.sym` stays usable for composition; particle-data access still raises through the untouched guard. Contract pinned by a new test in `test_0113`.

Gates re-run after the fix: test_0726 5/5, test_0113 7/7, test_0810 1/1, CI-equivalent unmarked `test_05* + test_07*` batch 429 passed / 0 failed, np2 swarm parallel set 19 passed.

Issue #244, opened 2026-06-16
Stokes_Constrained is slow in parallel despite low KSP iterations


gthyagi:
## Summary

`Stokes_Constrained` works in parallel, but the 8-rank runtime is much slower than the equivalent Nitsche free-slip solve even though the constrained solve only takes two KSP iterations. This suggests the bottleneck is setup/assembly/field handling for the block-constrained system, not Krylov convergence.

This came up while validating the Zhong et al. spherical-shell internal-boundary benchmark path using `SphericalShellInternalBoundary()` and exterior free-slip boundaries.

## Reproducer Context

External benchmark script:

```bash
cd /Users/tgol0006/uw_folder/uw3_git_gthyagi_latest/underworld3/.claude/worktrees/mantle-convection-benchmarks
pixi run -e amr-dev mpirun -np 8 python /Users/tgol0006/uw_folder/uw3-mantle-convection-benchmarks/benchmarks/005_internal_boundary_delta_probe.py
```

The script uses:

- `uw.meshing.SphericalShellInternalBoundary()`
- `ri = 0.55`, `rint = 0.775`, `ro = 1.0`
- `cellSize = 0.125` under 8 MPI ranks
- velocity `P2`, pressure `P1`
- internal radial natural load on the `Internal` boundary
- exterior free slip on `Upper` and `Lower`

The Nitsche comparison is:

```bash
pixi run -e amr-dev mpirun -np 8 python /Users/tgol0006/uw_folder/uw3-mantle-convection-benchmarks/benchmarks/005_internal_boundary_delta_probe.py -uw_freeslip_type nitsche
```

## Timing Evidence

Measured with `/usr/bin/time -p` on the same machine and mesh resolution:

| Method | Wall time | KSP iterations | Result |
| --- | ---: | ---: | --- |
| Nitsche free slip | `27.93s` | `1` | pass |
| `Stokes_Constrained` | `434.92s` | `2` | pass |
| constrained with temporary degree-1 multiplier test | `409.62s` | `2` | pass |

The temporary degree-1 multiplier test did not materially improve runtime, so multiplier polynomial degree is unlikely to be the main bottleneck.

## Observed Metrics

8-rank constrained run:

```json
{
  "cellsize": 0.125,
  "ksp_iterations": 2,
  "ksp_reason": 2,
  "l": 2,
  "max_boundary_area_relative_error": 0.009919085347124204,
  "max_y_l0_norm_error": 0.009849659575170255,
[...]

PR #318, opened 2026-07-05
fix(constrained): guard empty stratum IS in interior-multiplier reduction (#291)

### LINKED ISSUE #291: Stokes_Constrained segfaults at np>1 in the interior-multiplier section reduction
## Summary

`Stokes_Constrained` (block-constrained Stokes with in-saddle Lagrange multipliers) **segfaults (signal 11) at np>1** during `solve()`. It solves correctly in serial. The crash is localized to the interior-multiplier section reduction (`_constrain_interior_multipliers_in_section`, gated by `_reduce_interior_multiplier`, default `True`).

This is **independent of any preconditioner / custom-MG work** — it reproduces with plain GAMG and reproduces on a clean `development` checkout. Surfaced while extending the generalized-FMG custom-P feature (PR #290) to all solver families; custom-P on the constrained velocity block works in **serial**, but the parallel test had to be skipped because the underlying solver crashes.

## Reproducer (no custom-P, StructuredQuadBox, free-slip SolCx)

```python
import sympy, underworld3 as uw
from underworld3.function import analytic as A

mesh = uw.meshing.StructuredQuadBox(elementRes=(32, 32), minCoords=(0,0), maxCoords=(1,1), qdegree=3)
sol = A.SolCx(mesh, eta_A=1.0, eta_B=1.0e6, x_c=0.5, n=1)
s = uw.systems.Stokes_Constrained(mesh)
s.constitutive_model.Parameters.shear_viscosity_0 = sol.fn_viscosity
s.bodyforce = sol.fn_bodyforce
s.add_constraint_bc("Left",   g=0.0, normal=sympy.Matrix([[-1.0, 0.0]]))
s.add_constraint_bc("Right",  g=0.0, normal=sympy.Matrix([[ 1.0, 0.0]]))
s.add_constraint_bc("Bottom", g=0.0, normal=sympy.Matrix([[ 0.0,-1.0]]))
s.add_constraint_bc("Top",    g=0.0, normal=sympy.Matrix([[ 0.0, 1.0]]))
s.petsc_use_pressure_nullspace = True
s.tolerance = 1.0e-9
s.solve()
```

- `python repro.py` → solves, velocity error 8.8e-6.
- `mpirun -n 2 python repro.py` → **segfault during `solve()`** (both ranks build, then crash).

The canonical test `tests/test_1062_constrained_solcx.py` also segfaults under `mpirun -n 2 python -m pytest --with-mpi`.

## Diagnosis — `_reduce_interior_multiplier`

Toggling the interior-multiplier reduction isolates it (all at np=2):

| config | result |
|---|---|
| default (`_reduce_interior_multiplier = True`) | **segfault** |
| `s._reduce_interior_multiplier = False` | **solves**, verr ~1.7e-5 |
| `petsc_use_pressure_nullspace = False` | segfault (unchanged) |
| `snes_type = ksponly` | segfault (unchanged) |

So the crash is in **`_constrain_interior_multipliers_in_section`** (`src/underworld3/cython/petsc_generic_snes_solvers.pyx:6403`), the hand-built PetscSection that constrains interior multiplier DOFs. It iterates the local chart and uses `getTransitiveClosure` / boundary `getStratumIS` to build a new constrained section — likely not handling ghost / off-rank points correctly in parallel. Setting `_reduce_interior_multiplier = False` (line 6462 early-returns) is a working **workaround** but re-admits the interior multiplier DOFs (larger, more ill-conditioned `[p,h]` Schur block — see the method's own docstring).

## Notes

- Serial is unaffected; this is parallel-only.
- Reproduces on `development` (commit a7f0b11f) and the PR #290 branch — pre-existing, not introduced by either.
- If the constrained solver was exercised in parallel previously (e.g. around PR #265), it may predate the section-based `_constrain_interior_multipliers_in_section` rewrite, or used a configuration that didn't trigger it.

## Suggested fix direction

Make `_constrain_interior_multipliers_in_section` parallel-safe: restrict the chart walk / closure to owned points and respect the global section's ghost handling, or rebuild the constraint set from the global (not local) section. Add an np=2 regression (e.g. a parallel variant of `test_1062`).

lmoresi:
Closes #291.

## Summary

\`Stokes_Constrained\` segfaulted at np>1 in
\`_constrain_interior_multipliers_in_section\`
(\`src/underworld3/cython/petsc_generic_snes_solvers.pyx:7001\`).

The root cause was the boundary-IS guard. On a rank that owns zero points
with a given boundary label value,
\`dm.getLabel(\"UW_Boundaries\").getStratumIS(bvalue)\` returns a **valid PETSc
IS with \`getSize() == 0\`** — not None, non-zero handle, and
\`bool(IS) == True\`. So the existing guard \`if bd_is is not None:\` passed
through, and the subsequent \`bd_is.getIndices()\` crashed with SIGSEGV.

## Change

\`petsc_generic_snes_solvers.pyx:7001\` — replace

\`\`\`python
if bd_is is not None:
\`\`\`

with

\`\`\`python
if bd_is and bd_is.getSize() > 0:
\`\`\`

\`bool(bd_is)\` short-circuits the null-handle case (the same idiom
\`utilities/boundary_flux.py:70\` uses); \`.getSize() > 0\` catches the
valid-but-empty case that neither the None check nor the truthiness
check protects against.

## Verification

Lightweight unit-viscosity repro (16×16 quad, sinusoidal body force,
free-slip on all four walls; \`_reduce_interior_multiplier = True\` default):

| np | \`|u|_inf\`  | wall  | pre-fix       |
|----|-------------|-------|---------------|
| 1  | 2.533e-02   | 6.6 s | (unchanged)   |
| 2  | 2.533e-02   | 8.5 s | **SIGSEGV**   |
| 4  | 2.533e-02   | 5.0 s | **SIGSEGV**   |

Bit-identical parallel-vs-serial in this regime.

\`tests/test_1062_constrained_solcx.py\` (canonical SolCx \`eta_B=1e6\` test):
serial passes (198 s); the np=2 with-mpi rerun exceeded the 300 s timeout
under this session's compute (\`Stokes_Constrained\` is 34× slower than
Nitsche per #244), so a longer-timeout parallel variant of this test
(probably \`tests/parallel/test_1062_constrained_solcx_parallel.py\`) is
warranted as follow-up. Not blocking — the guard is a strict guard-widening
that cannot produce wrong physics; it only stops a previously-crashing rank
from calling \`.getIndices()\` on an empty IS.

## Related

- Issue #291 — this fix.
- #244 — the \`Stokes_Constrained\` parallel-slow issue that stops
  \`test_1062\` np=2 fitting a 300 s timeout on this machine.

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -6997,8 +6997,13 @@ class SNES_Stokes_SaddlePt(SolverBaseClass):
             fid_h = cbc.lam._solver_field_id
             bvalue = self.mesh.boundaries[cbc.boundary].value
             bd_is = dm.getLabel("UW_Boundaries").getStratumIS(bvalue)
+            # Guard for parallel: a rank owning no points with this label value
+            # gets back a valid non-None PETSc.IS with size 0 (bool(bd_is) = True,
+            # bd_is.handle != 0). Calling .getIndices() on that IS segfaults.
+            # `if bd_is` catches the null-handle case; `.getSize() > 0` catches
+            # the valid-but-empty case that None / handle checks miss (issue #291).
             keep = set()
-            if bd_is is not None:
+            if bd_is and bd_is.getSize() > 0:
                 for bp in bd_is.getIndices().tolist():
                     keep.update(dm.getTransitiveClosure(bp)[0].tolist())
             iset = interior_pts.setdefault(fid_h, set())
--- a/src/underworld3/utilities/rotated_bc.py
+++ b/src/underworld3/utilities/rotated_bc.py
@@ -31,6 +31,27 @@ from underworld3.utilities.boundary_flux import (
 _ROT_SOLVE_COUNT = 0
 
 
+def _warn_if_ksp_diverged(ksp, kind):
+    """Emit a rank-0 warning if a KSP finished on a KSP_DIVERGED_* reason
+    (negative converged-reason). ``ksp.solve`` never raises on divergence, so
+    without this the linear rotated-freeslip solvers here would return a
+    partial answer to the caller silently — which is exactly what let the 3D
+    rotation-nullspace bug (#306) go unnoticed until it produced clearly-wrong
+    physics in a downstream test."""
+    reason = int(ksp.getConvergedReason())
+    if reason >= 0:
+        return
+    its = int(ksp.getIterationNumber())
+    try:
+        rnorm = float(ksp.getResidualNorm())
+    except Exception:
+        rnorm = float("nan")
+    from underworld3 import mpi
+    mpi.pprint(f"[rotated_bc] WARNING: {kind} KSP did NOT converge "
+               f"(reason={reason}, iterations={its}, |r|={rnorm:.3e}); "
+               f"proceeding with the last iterate.")
+
+
 # --------------------------------------------------------------------------- #
 #  Rotation construction
 # --------------------------------------------------------------------------- #
@@ -291,6 +312,7 @@ def solve_rotated_freeslip(solver, boundaries, remove_rotation_gauge=True, verbo
         pc = ksp.getPC(); pc.setType("lu"); pc.setFactorSolverType("mumps")
         Uhat = dm.createGlobalVec(); ksp.solve(bhat, Uhat)   # returned in info → own it
         ksp_reason = ksp.getConvergedReason(); ksp_its = ksp.getIterationNumber()
+        _warn_if_ksp_diverged(ksp, kind="rotated direct-LU")
     else:
         Mp = _pressure_mass_schur_pmat(solver)
         Uhat, ksp_reason, ctx = _solve_rotated_iterative(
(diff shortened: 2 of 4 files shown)


lmoresi:
Thanks @copilot-pull-request-reviewer — added in 958f6f41:

1. **New lightweight parallel regression** at `tests/parallel/test_1062_constrained_stratum_guard_parallel.py` — unit-viscosity 16×16 quad box, free-slip on all four walls at np≥2. On a 2-way axis-aligned partition this puts one full boundary on each rank and zero on the other, triggering the empty-stratum path from the guard. Runs in **~10 s at np=2** and asserts `|u|_inf` matches the serial GOLDEN to `rtol=1e-3`. Pre-fix this test SIGSEGVs; post-fix it passes.
2. **Un-skipped `test_parallel_custom_fmg_stokes_constrained`** (`tests/parallel/test_1017_custom_mg_parallel_mpi.py`) — its skip reason explicitly said "auto-enables once #291 is fixed." Now marked `slow` (the SolCx eta_B=1e6 benchmark it uses is slow, per #244) so CI can select via `-m "not slow"` if needed.

Issue #327, opened 2026-07-05
Internal-facet normal orientation is rank-dependent at partition seams (vector surface integrals wrong at np>1)


lmoresi:
## Summary

Found while unskipping the test_0502 MPI tests for #326 (BF-13): on a BoxInternalBoundary mesh at np2, |∫ n_y dS| over the internal surface = 0.9375 instead of 1.0 — exactly one seam facet's normal is flipped (1 − 2/32). Scalar integrands are exact; only VECTOR integrands involving the internal-facet normal are affected, and only at partition seams.

## Status

The two affected tests (internal_normal_ny, internal_normal_weighted in tests/test_0502) are re-skipped with an accurate reason and a TODO(BUG) marker at the site. The other four MPI tests in that file are now unskipped and green at np2 after the BF-13 fix (#326).

## Impact

Any np>1 surface integral of a vector quantity over an internal boundary (fluxes across internal interfaces, internal traction integrals) can be silently wrong by O(seam-facet count / total facets).

Related: 2026-07 audit loose-ends/parallel family; discovered during PR #326 gating.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
# #327 — clean minimal repro: `BoxInternalBoundary(zintCoord=0.55, cellSize=1/32)` at np=2 gives 0.9375 exactly

## What I found

Attempted the fix at the /remote-control 48-core Linux box. The reported
signature does **not** fire at the test's default `zintCoord=0.5` on
Linux/MPICH — I get exact `∫ n_y = 1.0` and `∫ x·n_y = 0.5` at np=2 and
np=3 with `zint=0.5`, both on current `development` HEAD and at
`4680121a` (the commit that added the TODO(BUG) marker). The bug is
avoided because gmsh's partition seam at `zint=0.5` routes cleanly
around the internal boundary (the mesh is grid-aligned with the seam).

A sweep across `cellSize ∈ {1/32, 0.05, 0.0625, 0.08, 0.1, 0.125, 0.16,
0.2}` × `zintCoord ∈ {0.3, 0.4, 0.5, 0.55, 0.7}` at np=2 turned up
reliable reproductions at **off-grid** `zint` values:

```
[np=2 cell=1/32 zint=0.55] length=+1.000000(OK)  |ny_sum|=+0.937500(X)  |x_ny_sum|=+0.469727(X)
[...]

PR #359, opened 2026-07-09
fix(mesh): add mesh.canonical_normal accessor for partition-safe internal-boundary normals (#327)

lmoresi:
Partial fix for #327. Wires up the mesh factories' declared analytic normals as a partition-safe alternative to \`Gamma[i]\` on internal boundaries.

## Background

\`mesh.Gamma[i]\` JIT-substitutes \`petsc_n[i]\`, PETSc's per-quadrature
outward-normal-from-\`support[0]\` in \`DMPlexComputeBdIntegral\`. On an internal
boundary at a partition seam, different ranks disagree on which cell is
\`support[0]\` for the one seam facet, so the outward normal flips sign there and
a signed integral like \`∫ n_y dS\` is silently off by O(seam facets / total facets)
in parallel.

Reproducer (from the /remote-control session's #327 investigation):

\`\`\`python
mesh = BoxInternalBoundary(minCoords=(0,0), maxCoords=(1,1),
                           cellSize=1/32, zintCoord=0.55, simplex=True)
# np=1: ∫ Gamma[1] dS = 1.000000
# np=2: ∫ Gamma[1] dS = 0.937500   ← wrong, one seam facet flipped
\`\`\`

At \`zintCoord=0.5\` (grid-aligned) gmsh happens to route the np=2 seam around
the internal boundary and the bug is invisible; move \`zint\` off-grid and it
reappears reliably.

## Change

The mesh factory Enums (\`BoxInternalBoundary.boundary_normals_{2,3}D\`,
\`AnnulusInternalBoundary.boundary_normals\`, \`SphericalShellInternalBoundary\`,
etc.) already declare the analytic outward-pointing normal for each boundary,
but nothing consumed them. This PR wires them up:

**\`Mesh.canonical_normal(boundary_name)\`** — returns the declared sympy Matrix
outward normal for a boundary, or \`None\` if this mesh factory did not declare
one. Because it is a symbolic expression built from mesh coordinates it does
not touch \`petsc_n[]\`, so any signed-normal integral evaluated with it is
partition-independent for the meshes whose internal-boundary geometry the
factory knows (Box, Annulus, SphericalShell, ...).

## Tests

- \`test_bd_integral_internal_normal_ny\` and \`test_bd_integral_internal_normal_weighted\`
  in \`tests/test_0502_boundary_integrals.py\` were serial-only with a
  \`TODO(BUG)\` marker citing #327. Un-skipped and switched to use
  \`mesh.canonical_normal(\"Internal\")\`.
- New \`test_bd_integral_internal_canonical_normal_off_grid_zint\` builds the
  \`zint=0.55\` seam-crossing configuration (which is the one that reliably
  reproduces the bug on Linux gmsh) and asserts the canonical normal gives
  exactly \`|·| = 1.0\` at np≥1.
- \`test_0502_boundary_integrals.py\`: **23/23 pass** at both np=1 and np=2.

## What this does *not* fix

\`Gamma[i]\` itself remains partition-dependent on internal boundaries. Any
user code that reads \`Gamma[i]\` on an internal boundary at np>1 still gets
the wrong answer unless they switch to \`canonical_normal\`. The full fix is
larger — either canonicalising \`support[0]\` at construction (Cython for
\`dm.setCone\`) or exposing \`DMPlexOrientLabel\` with a proper depth-encoded
label constructor. Both routes explored in the [#327 investigation notes](https://github.com/underworldcode/underworld3/issues/327#issuecomment-4929238776);
neither is a session slice.

### DIFF
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -2709,6 +2709,61 @@ class Mesh(Stateful, uw_object):
             self._update_projected_normals()
         return self._projected_normals.sym
 
+    def canonical_normal(self, boundary_name):
+        r"""Analytic outward-pointing normal for a boundary, or ``None``
+        if no analytic normal was declared for that boundary.
+
+        Sourced from the mesh factory's ``boundary_normals`` Enum: for
+        axis-aligned box boundaries this is a constant sympy Matrix, for
+        annulus / spherical-shell radial boundaries it is the analytic
+        radial unit vector, and so on.
+
+        The primary caller is code that needs a **partition-safe** normal
+        on an *internal* boundary — see :issue:`327`. On an internal
+        boundary at a partition seam, PETSc's per-quadrature ``petsc_n[]``
+        (surfacing as :attr:`Gamma` / :attr:`Gamma_N`) is derived from
+        ``support[0]`` of the DMPlex facet closure, which is
+        partition-dependent; different ranks disagree on which cell is
+        "support[0]" for the one seam facet, and the outward normal of
+        that facet flips sign. A signed integral of ``Gamma[k]`` is then
+        wrong by O(seam-facets / total-facets). The analytic normal
+        returned here is partition-independent and does not touch
+        ``petsc_n``, so it sidesteps the defect entirely for the mesh
+        classes that know their internal-boundary geometry
+        (:class:`BoxInternalBoundary`, :class:`AnnulusInternalBoundary`,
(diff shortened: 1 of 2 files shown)


lmoresi:
Superseded by #361, which contains this branch's commit (the canonical_normal accessor) and builds the full unified interface on top: mesh.Gamma resolves per boundary (external → exact petsc_n[]; internal → the declared analytic normal), with deformation invalidation and an np≥4 label-completion deadlock fix. See #361 for details.

Issue #360, opened 2026-07-13
Moving mesh (mesh.deform / free surface / r-adapt) crashes at np>1: _build_kd_tree_index nav_coords[points - pStart] out of bounds


lmoresi:
## Summary

**Every moving-mesh workflow (free surface, `mesh.deform`, r-adaptation) crashes at `np > 1`.** The Stokes solve completes correctly on all ranks, then `mesh.deform()` crashes at the end of the step inside `_build_kd_tree_index`.

This is **parallel-only** — the identical script runs fine serially. It is **not** particle/swarm related: reproduced with both a compositional swarm AND a pure mesh-level-set / field configuration (no particles). Both call `mesh.deform()` every step and hit the same line.

## Reproduce

Any free-surface / `mesh.deform` driver under `mpirun -np 4` (verified with the Crameri Case-2 driver, both swarm and mesh-LS plume). Step 1 solves on all ranks, then:

```
File ".../discretisation/discretisation_mesh.py", line 3129, in deform
File ".../discretisation/discretisation_mesh.py", line 3188, in _deform_mesh
File ".../discretisation/discretisation_mesh.py", line 2349, in nuke_coords_and_rebuild
File ".../discretisation/discretisation_mesh.py", line 4547, in _build_kd_tree_index
    cell_point_coords = nav_coords[points - pStart]
IndexError: index -11 is out of bounds for axis 0 with size 10
IndexError: index 14 is out of bounds for axis 0 with size 14
```

Indices go **both negative and past the end** of the local array (different value per rank).

## Root cause

`_build_kd_tree_index` (`discretisation_mesh.py:4547`, and the identical `_build_kd_tree_index_PIC` at :4723) maps vertex **plex-point numbers** to **coordinate rows** with an affine shift:

```python
points = nav_dm.getTransitiveClosure(cell_id)[0][-cell_num_points:]  # vertex plex-points
cell_point_coords = nav_coords[points - pStart]                      # <-- wrong in parallel
```

`points - pStart` is only a valid row index when vertex numbering is serial-contiguous. After `nuke_coords_and_rebuild` re-points `_nav_coords` at `self.dm.getCoordinatesLocal()` (:2404), the local **overlapped** coordinate array is ordered by the coordinate **PetscSection**, not by `plex_point − pStart`. In parallel, ghost/halo vertices make `points − pStart` run negative or overflow → `IndexError`.

[...]

PR #361, opened 2026-07-14
Unify boundary-normal interface behind mesh.Gamma (#327) + np=4 label deadlock fix

lmoresi:
## Summary

**One user-facing normal symbol.** `mesh.Gamma` now works on any boundary; the four alternative interfaces (`Gamma_N`, `Gamma_P1`, `boundary_normal(b)`, `canonical_normal(b)`) are no longer needed in user code (all retained for back-compat; the first two are documented as deprecated).

Consumers that know their boundary — `uw.maths.BdIntegral` and natural BCs (`add_condition` / `add_natural_bc`) — resolve `Gamma` per boundary via the new `Mesh._resolve_boundary_normals`:

- **External boundary** — unchanged: `Gamma` compiles to PETSc's exact per-quadrature outward unit normal `petsc_n[]`. This path is byte-identical to before, so zero solver risk.
- **Internal boundary** — `petsc_n[]` orientation follows the partition-dependent DMPlex `support[0]` (#327: `mpirun -n 2` gave 0.9375 instead of 1.0 for a unit signed-normal integral). `Gamma` components are substituted with the mesh factory's declared analytic normal (`canonical_normal`), which is partition-independent by construction.
- **Internal boundary, no declared normal** — loud `RuntimeError`. There is no orientation convention for an arbitrary internal surface until `DMPlexOrientLabel` handles open surfaces (pending upstream with PETSc); failing loudly beats integrating a partition-dependent sign. (Note `boundary_normal()` silently returns zeros there — it skips support-size-2 facets — which is why it is not a fallback.)

Internal-boundary detection: any labeled facet with support size 2, MAX-allreduced across ranks and cached per boundary.

## Orientation convention (behaviour change)

Declared internal-boundary normals point **from region `Inner` to region `Outer`**:
- 2D `BoxInternalBoundary`: `Internal` flipped `[0,-1]` → `[0,1]` (now consistent with 3D `[0,0,1]`)
- `AnnulusInternalBoundary` / `SphericalShellInternalBoundary`: radially outward `unit_e_0` (the annulus Enum existed but was never attached to the mesh; the spherical factory had none — both now attached)

Code that consumed the old 2D `[0,-1]` declaration via `canonical_normal` will see a sign flip.

## Second commit: np≥4 deadlock in `_dm_unstack_bcs` (pre-existing)

While validating at np=4, every `BoxInternalBoundary` construction deadlocked — **on `development` too**, i.e. pre-existing and independent of this change. `_dm_unstack_bcs` iterated the rank-local `getNonEmptyStratumValuesIS()` while making collective `labelComplete` calls per value; ranks whose partition misses a boundary make fewer collective calls → deadlock (np=2 escaped because both halves touch all boundaries). Fixed by iterating the allgathered union of values, with local stratum access guarded. Kept as its own commit for cherry-pick ease.

## Validation

- `test_0502_boundary_integrals.py`: 24/24 at np=1, np=2, np=4 (np=4 previously hung indefinitely; now 5s)
- Regression trio: `test_1000_poissonNaturalBC` (external natural BCs through the new resolver), `test_1060_nitsche_freeslip` (`mesh.Gamma` in `add_natural_bc`), `test_1061_constrained_freeslip` (new `boundary_normal(boundary)` default) — 23 passed
- New/updated tests: plain-`mesh.Gamma` internal signed-normal integrals with 1e-6 assertions and the Inner→Outer sign; `canonical_normal` accessor back-compat; the #327 off-grid `zintCoord=0.55` partition-seam regression

## Relationship to PR #359

This branch **contains** #359's commit (`canonical_normal` accessor) and supersedes it — the accessor alone still required users to pick the right interface. Suggest closing #359 in favour of this PR (leaving that to the maintainers).

## Deferred (flagged, not in this PR)

- Doc sweep: 16 user-doc references to `Gamma_N`/`Gamma_P1` (not mechanical — curved-boundary guides should point at `boundary_normal`)
- `TODO(BUG)` in `cartesian.py`: Box **external** `boundary_normals` entries are inward-pointing while `canonical_normal` documents outward — needs a consumer audit before flipping
- `_assemble_boundary_normal` parallel seam ADD-reduce TODO
- Report back on #327 that the UW3-side fix landed via declared analytic normals, with `DMPlexOrientLabel` open-surface support still pending upstream

### DIFF
--- a/src/underworld3/adaptivity.py
+++ b/src/underworld3/adaptivity.py
@@ -502,6 +502,17 @@ def _dm_unstack_bcs(dm, boundaries, stacked_bc_label_name):
     stacked_bc_label = dm.getLabel(stacked_bc_label_name)
     vals = stacked_bc_label.getNonEmptyStratumValuesIS().getIndices()
 
+    # BUGFIX: ``vals`` is rank-local (a rank only sees stratum values that
+    # are non-empty on its own partition), but the loop below makes
+    # collective calls (``labelComplete``) once per value. If ranks
+    # iterate different value sets — routine at np>=4, where a partition
+    # quadrant may touch no facet of some boundary — the collective call
+    # counts diverge and the run deadlocks. Iterate the global union so
+    # every rank makes the same sequence of collective calls; guard the
+    # local stratum access to values that are live on this rank.
+    local_vals = set(int(v) for v in vals)
+    vals = sorted(set().union(*uw.mpi.comm.allgather(local_vals)))
+
     # Clear labels just in case
     for b in boundaries:
         dm.removeLabel(b.name)
@@ -520,8 +531,9 @@ def _dm_unstack_bcs(dm, boundaries, stacked_bc_label_name):
             continue
 
         b_dmlabel = dm.getLabel(b.name)
-        lab_is = stacked_bc_label.getStratumIS(v)
-        b_dmlabel.setStratumIS(v, lab_is)
+        if v in local_vals:
+            lab_is = stacked_bc_label.getStratumIS(v)
+            b_dmlabel.setStratumIS(v, lab_is)
 
         # BUGFIX(#162): expand the boundary label to include its closure
         # (the vertices and edges bordering the labeled facets). The
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -1600,6 +1600,13 @@ class SolverBaseClass(uw_object):
 
         from collections import namedtuple
         if c_type == 'neumann':
+            # mesh.Gamma in a natural-BC expression resolves per boundary:
+            # external boundaries keep the exact per-quadrature petsc_n[];
+            # internal boundaries substitute the declared analytic normal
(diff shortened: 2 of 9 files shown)


lmoresi:
Added 2f98dec7: **declared analytic normals are invalidated by mesh deformation.**

The declaration describes the factory geometry, so after `mesh.deform()` / `adapt()` resolving `mesh.Gamma` on an internal boundary (or calling `canonical_normal`) now raises a clear `RuntimeError` instead of silently integrating a stale normal. Re-assigning `mesh.boundary_normals` re-declares against the current geometry (e.g. radial normals remain valid after a radius-preserving remesh: `mesh.boundary_normals = mesh.boundary_normals`).

Mechanism: `nuke_coords_and_rebuild` bumps a `_geometry_version` counter (all deformation paths funnel through it); the `boundary_normals` setter stamps the declaration with the current version. External boundaries unaffected — `petsc_n[]` is exact on deformed geometry.

New regression test (serial-only until #360: `deform` itself crashes at np>1). Re-validated: 0502 serial 25 passed, np=2/np=4 24 passed + 1 skip; natural-BC/Nitsche/constrained trio 23 passed.

Issue #346, opened 2026-07-07
Latent MPI deadlock: spring smoother returns early on rank-local n_free==0 before collectives


lmoresi:
Found (marker-only, Charter §9) during the Wave D smoothing readability pass, 2026-07-08. A `TODO(BUG)` marker lands with PR #345 in `src/underworld3/meshing/smoothing/spring.py`.

The spring-equilibrium mover takes a rank-local early return when a rank has no free (unpinned) nodes, BEFORE reaching the collective reduction calls later in the routine. On a partition where some rank owns only pinned/boundary nodes (small partitions, heavily graded meshes at high rank counts), that rank exits while the others enter the allreduce — a classic collective mismatch deadlock.

Not observed in the campaign's np2 gates (both ranks own free nodes in the test meshes); the hazard is latent and partition-dependent.

Suggested fix: replace the early return with a zero-work path that still participates in every collective (empty contribution), or hoist the collectives above the return. Needs an np>2 test with a deliberately starved rank.

Related: the Wave D smoothing PR #345 (marker placed, behaviour untouched per readability-wave scope discipline).

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
Maintainer context (2026-07-08): the spring smoother is one of the retired mover methods — only mmpde really survived testing. This deadlock is therefore in a low-priority legacy path: do not invest in hardening it. The right resolution is likely deletion of the retired movers (spring / Monge-Ampère / OT-step / anisotropic-Winslow) in a future prune, keeping mmpde (and the Taubin surface smoother, which is separate). If a prune is ruled, close this issue with that PR instead of fixing the collective ordering.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
Additional maintainer context: mmpde *replaced* the Monge-Ampère/OT family — it provides the equivalent equidistribution functionality under an isotropic metric (and handles anisotropic metrics besides). The retirement is supersession, not just test attrition, so a prune of the legacy movers loses no capability.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
Resolved by #371 (4bcee78a): the spring mover — and with it this rank-local early-return deadlock — is deleted; mmpde (which superseded the spring/MA/OT family) and the Taubin surface smoother survive.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

PR #368, opened 2026-07-15
fix(mmpde): finite SPD fallback for fully-degenerate metric evaluations (#352)

### LINKED ISSUE #352: _spd_sanitise: all-NaN eigenvalues on a rank propagate NaN instead of the SPD fallback (mmpde live path)
Found by the Copilot pass over Wave D PR #345 (2026-07-08); `TODO(BUG)` marker at `src/underworld3/meshing/smoothing/mmpde.py` (commit 300bba41). Confirmed by numpy probe; pre-existing (bit-identical at base, code from the #259 SPD-floor era).

When every eigenvalue on a rank is NaN/inf, `np.nanmax(w)` is non-finite, making the floor non-finite and the projection emit NaNs instead of a usable SPD fallback — defeating the sanitiser's purpose exactly when it is most needed. **This is the surviving mmpde mover's path** (unlike the retired-mover findings), so it deserves a real numerics fix: guard `wmax` to a finite value (e.g. fall back to identity metric scaled by a mesh-derived length) when no finite eigenvalue exists.

Related: #259 (original SPD-floor fix), #346 (retired-mover collective hazard — different path).

lmoresi:
Closes #352.

## The bug

`_spd_sanitise` — the guard that keeps the MMPDE energy's fractional powers finite — failed exactly when it was most needed: if EVERY eigenvalue on a rank was NaN/inf (fully-degenerate metric evaluation, e.g. a starved or badly-deformed partition under live convection), `np.nanmax` returned NaN, `max(nan, 1.0)` propagated it (Python `max` compares with `>`, false for NaN), and the "sanitised" output was NaN instead of the intended benign SPD fallback. Confirmed by direct probe on the installed build before any change.

## The fix (two guards — the issue's sketch alone was insufficient)

1. **Floor anchor**: computed from the finite eigenvalues only; unit scale when none exist.
2. **Basis fallback**: the eigenvector basis of a fully-degenerate tensor is itself NaN, so even a finite floor rebuilt NaN through the `einsum`; such tensors now rebuild in the identity basis — the isotropic floor metric ("no information here, stay isotropic").

The helper is hoisted to module level (it captured nothing) for unit-testability; the mover's behaviour for any input with at least one finite eigenvalue is unchanged, and a genuine SPD batch passes through **bit-identical** (asserted).

## Verification

- New `TestSPDSanitise` in `test_0850` (level_1/tier_a): all-NaN and all-inf batches → finite SPD; valid SPD → bit-identical passthrough; one degenerate tensor cannot perturb neighbours. Shown failing against the unmodified build first.
- Gates: `level_1 and tier_a` 380 passed (376 baseline + the 4 new tests); np2 `test_0855_mesh_smoothing_parallel` 4 passed; style gate green (86 allowlisted).
- No dependency/lockfile changes.

This is the live-mmpde robustness path (the surviving mover; see #346 discussion) — relevant to the planned MMPDE+NVB capstone.

### DIFF
--- a/src/underworld3/meshing/smoothing/mmpde.py
+++ b/src/underworld3/meshing/smoothing/mmpde.py
@@ -14,6 +14,54 @@ from .graph import (_tri_cells, _signed_areas, _owned_cell_mask,
                     _global_sum, _global_min, _global_max, _global_mean)
 
 
+def _spd_sanitise(M):
+    """Project a batch of metric tensors ``(N, d, d)`` onto SPD.
+
+    The MMPDE functional uses fractional powers (``sqrt(detM)``,
+    ``detM**((1-p)/2)``, ``S**q``) that are defined only for a
+    symmetric-positive-definite metric. FE extrapolation outside the current
+    mesh can hand back a non-SPD tensor; this projection floors its
+    eigenvalues at a small RELATIVE level so a genuine SPD metric passes
+    through unchanged while garbage becomes a benign "coarsen here" tensor
+    (tiny positive eigenvalues) instead of a NaN in the energy.
+    """
+    # Symmetrise: the metric is symmetric by construction, so for a valid
+    # tensor this is an exact no-op (M_ij == M_ji bit-for-bit).
+    Ms = 0.5 * (M + np.swapaxes(M, -1, -2))
+    if Ms.shape[0] == 0:
+        return Ms                                   # rank owns no cells
+    w, Vc = np.linalg.eigh(Ms)
+    wmax = float(np.nanmax(w[np.isfinite(w)], initial=-np.inf))
+    if not np.isfinite(wmax):
+        # Every eigenvalue on this rank is NaN/inf: the metric carries no
+        # information here, so anchor the floor at the unit scale — the
+        # projection below then rebuilds every tensor as the same benign
+        # isotropic fallback instead of propagating NaN (#352).
+        wmax = 1.0
+    floor = max(wmax, 1.0) * 1.0e-8
+    # Per-tensor SPD test: a cell is "bad" only if one of its OWN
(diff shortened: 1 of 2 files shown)


lmoresi:
An independent adversarial review pass (standing in for the quota-exhausted Copilot) probed this fix and produced two accepted findings, now addressed in the second commit:

1. **Latent 3D gap (confirmed by probe)**: degenerate-input `eigh` is LAPACK-path dependent — the 2×2 kernel returns quiet NaNs but the general kernel raises `LinAlgError` from the *batched* call before any guard runs, and one bad tensor takes down the whole batch. Unreachable via the 2D-only mover today, but the 3D MMPDE capstone lands exactly on this path. Fixed with a per-tensor retry; two new 3×3 tests (all-NaN, and mixed-batch-survives-batched-failure with neighbours preserved bit-identical).
2. **Docstring honesty**: the floor has an absolute 1e-8 minimum (pre-existing semantics), so the pass-through guarantee is scoped to the O(1)-normalised metrics the mover uses — now stated.

One PR-body correction from the review: the invariance claim holds **per tensor** (any tensor with finite eigenvalues is unchanged); a fully-degenerate tensor in a *mixed* batch previously produced NaN and now gets the finite fallback — a strict improvement, and exactly what the tests assert.

Review verdicts on the remaining attack surfaces: the feared `nanmax`-inf divergence is unreachable (symmetrisation bounds eigenvalues below inf); finite-eigenvalue/NaN-basis never occurs (10k-tensor hunt); hoist is behaviour-identical elsewhere; both metric-eval paths remain sanitised; no new cross-rank dependence.

Gates after hardening: `level_1 and tier_a` 382 passed (376 baseline + 6 tests), `test_0850` 15/15.

PR #380, opened 2026-07-22
feat(constitutive): power-mean yield soft-min smoother + rampable delta constant

lmoresi:
## What

Adds the **power-mean soft-min yield smoother** and makes the soft-min softness δ a **`constants[]` atom** (runtime-rampable with no JIT recompile) on the `ViscousFlowModel` base class, so the visco-plastic subclasses inherit one implementation. This is the generalisable, scalable substrate for a yield homotopy toward the sharp `Min` surface (companion to the solver difficulty/bounded-solve tooling in #377).

**Base class (`ViscousFlowModel`):**
- `_combine_yield(eta_ve, eta_pl)` -- the shared viscous/plastic combination, keyed on `yield_mode`: `"min"` (exact hard `Min`), `"harmonic"`, or `"softmin"` (the δ soft-min in the family chosen by `yield_smoother`). Development's tri-modal `yield_mode` semantics are **preserved exactly** (the source branch had collapsed `"softmin"` into `"min"`; that is deliberately NOT adopted).
- `_get_yield_softness` / `_get_yield_offset` -- δ and the onset offset held as `constants[]` UWexpression atoms (the offset defined symbolically in terms of δ = one symbol in the stress tensor, no tensor blow-up; one δ update repacks both). Ramping δ via the atom + `solver._update_constants()` forces no recompile.
- `yield_smoother` property/setter -- `"sqrt"` (default; overshoots τ_y in the transition) or `"powermean"` (undershoots τ_y -- `eta_eff <= Min` always -- overflow-safe on geodynamic ranges).

**Models (Stage 1 -- isotropic, test-covered):**
- **`ViscoPlasticFlowModel` (Drucker-Prager):** defaults to `yield_mode="min"` == today's exact hard `Min` (zero behaviour change), and **gains** `yield_mode` / `yield_softness` knobs so it can opt into the softmin / power-mean homotopy -- the model the hard-case DP homotopy needs.
- **`ViscoElasticPlasticFlowModel` (VEP):** its inline sqrt soft-min routed through `_combine_yield`; δ moves float -> atom (value-identical); `yield_softness` setter syncs the atom.

## Why

The power-mean smoother is the preferred, physically-admissible (undershoot) homotopy family; holding δ as a rampable constant is what makes a scalable δ-continuation cheap (no recompile per step).

## Safety -- zero behaviour change at stock settings

`_combine_yield` at default settings is **bit-identical** to the previous inline law (verified numerically, diff 0.00e+00) -- δ merely moves from a baked float to a `constants[]` symbol evaluated to the same number. Power-mean and runtime-δ are entirely opt-in. The distrusted in-solve `enable_yield_homotopy` ramp driver is intentionally **not** part of this PR.

## Scope

Stage 1 (this PR): base helper + δ-atom + power-mean, wired to the DP and VEP models. **Stage 2 (follow-up):** the TI-VEP / TI-VEP-Split anisotropic sites (4 sites feeding rank-4 split/lagged tensors), gated by the TI solver tests.

## Tests

`tests/test_1055_yield_smoother.py`: δ=0 exact Min; power-mean undershoot + overflow-safety + ->Min as δ->0; smoother validation/default; runtime δ ramp no-recompile; DP opt-in.

`level_1 and tier_a` gate green: **402 passed, 0 failed, 0 errors** (1 pre-existing unrelated xfail).

Note: 3 `test_1052_VEP_stability_regression` (level_2) tests are red, but they fail **identically on the development baseline without this change** (bisected) -- pre-existing, unrelated.

### DIFF
--- a/src/underworld3/constitutive_models.py
+++ b/src/underworld3/constitutive_models.py
@@ -899,6 +899,125 @@ class ViscousFlowModel(Constitutive_Model):
             )
         )
 
+    # --- Yield soft-min smoother (shared by the visco-plastic subclasses) -----------
+    # The δ soft-min regularisation and the smooth-min FAMILY selection live on the
+    # base class so every yielding model inherits one implementation. δ is held as a
+    # constants[] UWexpression atom (not a baked float) so a homotopy can ramp it at
+    # runtime via PetscDSSetConstants with no JIT recompile.
+
+    def _get_yield_softness(self):
+        r"""The soft-min regularisation δ as a ``constants[]`` UWexpression atom.
+
+        Created lazily and kept in sync with ``self._yield_softness`` (the
+        configured numeric value).  Storing δ as a UWexpression — rather than
+        baking the float into the compiled flux — lets a yield homotopy ramp
+        δ at runtime via ``PetscDSSetConstants`` with no JIT recompile.
+        ``δ = 0`` makes the sqrt law identically ``Min``.
+        """
+        delta_value = getattr(self, "_yield_softness", 0.0)
+        if getattr(self, "_yield_softness_expr", None) is None:
+            self._yield_softness_expr = expression(
+                R"{\updelta_{y}}",
+                sympy.Float(delta_value),
+                "Yield soft-min regularisation δ (rampable constant; δ=0 ⇒ exact Min)",
+            )
+            # Onset offset (-1+√(1+δ²))/2 keeps g(0)=1 (no spurious yield below
+            # onset).  Held as its OWN constant atom — a single symbol in the
+            # stress tensor — so it does not blow the tensor up, while still
+            # tracking δ symbolically (one δ update repacks both constants).
+            self._yield_offset_expr = expression(
+                R"{\updelta_{y,0}}",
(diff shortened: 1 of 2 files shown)


lmoresi:
Closed as fully superseded — audit against current development (698d89a4):

- The power-mean soft-min smoother, δ and onset-offset as rampable constants[] atoms, `_combine_yield` tri-modal semantics, the `yield_smoother` property, and `tests/test_1055_yield_smoother.py` all landed on development (via the #441 line of work), so everything this PR introduced is already merged.
- Development's power-mean is the **inf-safe harmonic-normalised form** (constitutive_models.py ~1010-1024, with the inf/inf-at-rigid-points rationale documented inline); this branch still carries the older product-over-sum form that NaNs on a cold start — merging would regress that fix.
- The remaining lines unique to this branch are pre-refactor remnants development deliberately replaced (old `-oo` defaults, old `Max(yield_stress_min, …)` guards superseded by the smooth_max floor, old `flux_jacobian` docstring superseded by the settable version from #362).

Nothing of value is lost by closing.

Issue #379, opened 2026-07-22
Callback-array parallel safety: remaining exposures after #376 (swarm.points callback, delayed-callback replay, central guard)


lmoresi:
Follow-up to #376 (fixed in #378). The root cause there — `NDArray_With_Callback.__array_finalize__` hands the sync callbacks to every array derived by indexing, so callbacks written for the canonical storage fire on views and temporary copies — is more general than the two canonical callbacks that were fixed. The historical `disable_inplace_operators=True` blocking on `swarm.points` ("disabled for parallel safety", af308d86, Sept 2025) was a symptom-level mitigation of the same disease. Four remaining items, none blocking #378:

## 1. `swarm.points` / `swarm_update_callback` still has the disease, plus a worse one

- The callback calls collective `migrate()` from inside a per-write callback — the exact pattern the swarm *variable* canonical callback was rewritten to avoid (SWARM-03 deferred migration) — and raises `RuntimeError` rank-locally on a size mismatch. Ranks that write unevenly diverge on collectives.
- The in-place blocking only covers the `i`-operators; `__setitem__` on a derived fancy-indexed copy still fires the callback.
- **The documented example in `migration_control()`'s own docstring raises today** (verified):

```python
with swarm.migration_control():
    swarm.points[mask1] += delta1
# RuntimeError: In-place addition (+=) is disabled for parallel safety.
```

The `disable_inplace_operators` flag propagates to the fancy copy via `__array_finalize__`, so the advertised pattern is simply broken. `.points` is already deprecated in favour of swarm variables — either port it to the guarded-callback + deferred-migration pattern (after which the blocking flag can be dropped and the docstring becomes true), or accelerate its retirement.

## 2. Delayed-callback replay is rank-asymmetric by construction

`delay_callback` / `uw.delay_callbacks_global` queue `(callback, array, context)` per rank and replay them between blanket barriers at context exit. Queue lengths and contents are rank-dependent (each rank queues what it locally triggered) while the callbacks contain collectives — the barriers around the batch cannot fix mismatched collectives inside it. Per-item exceptions are swallowed with `logger.warning` (the same swallow that hid #376 for ten months). The robust shape already exists in the codebase: the swarm canonical callback's `_pending_petsc_sync` marks variables dirty by NAME and flushes collectively at context exit. The delay machinery should do the same — flush dirty variables canonically, not replay rank-local event queues.

## 3. The #376 guard lives in two closures, not the class

[...]

Issue #376, opened 2026-07-21
Canonical PETSc-sync callback fires on fancy-indexed array copies — serial warning, parallel hang/crash


lmoresi:
## Symptom

Serial: `mesh.redistribute_nodes(...)` (and anything that touches `mesh.Gamma_P1`) emits swallowed log warnings:

```
Callback error in ... canonical_data_callback ...: could not broadcast input array from shape (64,2) into shape (65,2)
```

Results are unaffected in serial. **In parallel the same defect is fatal**: at np=2 the MMPDE mover crashes/hangs with `General MPI error / MPI_ERR_BUFFER` inside `coord_dm.getGlobalVec()`, with one rank spinning at 100% CPU (mismatched collectives).

## Minimal repro

```python
import sympy, underworld3 as uw
mesh = uw.meshing.UnstructuredSimplexBox(minCoords=(0, 0), maxCoords=(1, 1),
                                         cellSize=0.4, refinement=1, qdegree=2)
x, y = mesh.X
d = sympy.Abs((x - 0.35) * 0.866 + (y - 1.0) * 0.5)
h = sympy.sqrt(0.08**2 + (1.6 * d) ** 2)
mesh.redistribute_nodes(1.0 / h**2, slip_surfaces=True)   # 25 swallowed callback errors
```

Run the same under `mpirun -n 2` and the mover dies with `MPI_ERR_BUFFER` / hangs.

## Root cause

`NDArray_With_Callback.__array_finalize__` (`utilities/nd_array_callback.py:185`) copies `_callbacks` and `_owner` onto **every** derived array — including fancy-indexed **copies**. A masked in-place update such as

```python
n.data[nonzero] /= mag[nonzero, None]     # discretisation_mesh.py:_update_projected_normals
```

expands to `__getitem__` (a copy), an in-place divide **on the copy**, then `__setitem__` back into the parent. The in-place divide on the copy triggers the canonical PETSc-sync callback, which tries to pack the subset (64 rows — the box corner node has a zero-magnitude fallback normal and is masked out) into the full 65-row vec.

- **Serial**: the pack raises `ValueError`, is caught and logged (`nd_array_callback.py:525`); the later `__setitem__` on the parent syncs correctly, so values survive. Cosmetic only.
[...]

PR #381, opened 2026-07-22
Callback-array parallel safety: central view/copy guard, honest exceptions, no dead snapshots (#379 items 3+4b)

lmoresi:
First of the #379 series (see the issue for the full map). **Draft until #378 merges** — the two canonical callbacks this PR migrates are the same closures #378 patches inline, so this PR merges second and absorbs a small, mechanical conflict (the inline guard block is deleted here because the guard moves into the class).

## What this does

**A central guard for canonical-storage callbacks.** Callbacks written for a variable's canonical storage are inherited by every array derived from it, so they also fire on views and on temporary fancy-index copies. A copy's contents differ from rank to rank, so a PETSc sync run from inside the callback executes its collective operations on some ranks only — that is the #376 parallel hang. The new `add_canonical_callback()` applies the fix once, on the class: views hand the full canonical array to the callback, copies are skipped (numpy's write-back re-fires through the parent), and view-vs-copy is decided by identity in numpy's base chain — never `np.may_share_memory`, which is false for any zero-size array and would re-create the rank asymmetry on ranks with no local data. Any future callback registered this way gets the protection for free instead of re-inheriting the trap.

The mesh-variable and swarm-variable canonical callbacks now register through it. The mesh callback body moves to `MeshVariable._flush_canonical_to_petsc()`, which becomes the single write-back path (and is reused by the planned `synchronised_array_update` rework, the next PR in the series).

**Callback exceptions now propagate.** The immediate callback path used to swallow every exception into a `logger.warning`. A swallowed failure leaves PETSc out of sync with the canonical array on one rank only — the silent desynchronisation that hid #376. Failures now surface where they happen.

**Dead weight removed.** `_create_variable_array` (mesh and swarm) and `_create_flat_data_array` had no callers and embodied the unguarded pattern — deleted. Every write also used to snapshot prior values into `change_info['old_value']` (a full-array copy per in-place operation); nothing ever read it, so the snapshot is gone and the key stays as `None` for compatibility.

**Not touched here:** the `swarm.points` legacy callback (marked `TODO(BUG)`; removed by the #379 item-1 PR) and the delayed-replay machinery (rebuilt in the item-2 PR).

## Verification

- New `tests/test_0059_canonical_callback_guard.py` (level_1 / tier_a, written first and shown to fail): view/copy classification including the zero-size-view empty-rank case, exactly-once firing for masked `+=` writes, and a no-reference-cycle check.
- Quick gate `level_1 and tier_a`: 405 passed, 1 pre-existing xfail.
- Parallel smoke at np=2: swarm write/timestep and mesh-deform kd-tree tests pass.

### DIFF
--- a/docs/developer/UW3_Developers_NDArrays.md
+++ b/docs/developer/UW3_Developers_NDArrays.md
@@ -438,21 +438,13 @@ The NDArray_With_Callback system intercepts NumPy operations:
 ```python
 
 def __setitem__(self, key, value):
-    # Capture old value for callback info
-    old_value = self[key].copy() if self._track_changes else None
-    
     # Perform the actual assignment
     super().__setitem__(key, value)
-    
-    # Trigger callbacks with operation info
-    if self._callbacks_enabled:
-        change_info = {
-            'operation': 'setitem',
-            'indices': key,
-            'old_value': old_value,
-            'new_value': self[key],
-        }
-        self._trigger_callback(change_info)
+
+    # Notify callbacks. old_value is always None: no registered callback
+    # ever read prior values, and snapshotting them cost a full-array
+    # copy per write (the key is retained for dict-shape compatibility).
+    self._trigger_callback("setitem", indices=key, new_value=value)
 ```
 
 ## PETSc Synchronization Details
--- a/docs/developer/UW3_Style_and_Patterns_Guide.md
+++ b/docs/developer/UW3_Style_and_Patterns_Guide.md
@@ -159,8 +159,14 @@ arr = NDArray_With_Callback(data, owner=self, callback=func)  # With callback
 
 # Callback signature
 def callback(array: NDArray_With_Callback, change_info: dict) -> None:
-    # change_info contains: operation, indices, old_value, new_value, array_shape, array_dtype
+    # change_info contains: operation, indices, old_value (always None,
+    # retained for compatibility), new_value, array_shape, array_dtype
     pass
+
+# For callbacks that synchronise a variable's CANONICAL storage, register
+# with add_canonical_callback(func): it guards against firing on derived
+# views/copies (rank-asymmetric collectives, #376) and hands the callback
+# the canonical array every time.
 ```
 
 ## Array vs Data Property Shapes
--- a/docs/developer/subsystems/data-access.md
+++ b/docs/developer/subsystems/data-access.md
@@ -181,7 +181,8 @@ Key implementation locations:
 - **MeshVariable**: `discretisation_mesh_variables.py`
   - `array` property (lines 700-720)
   - `data` property for compatibility (lines 680-700)
-  - Callback registration in `_create_variable_array()`
+  - Callback registration in `_create_canonical_data_array()` via
+    `NDArray_With_Callback.add_canonical_callback()` (central view/copy guard)
 
 - **SwarmVariable**: `swarm.py`
   - `array` property with caching (lines 857-871)
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -244,7 +244,18 @@ def _mesh_coords_update_callback(array, change_context):
         uw.pprint(f"Mesh update callback - mesh deform")
 
     coords = array.reshape(-1, mesh.cdim)
-    mesh._deform_mesh(coords, verbose=verbose)
+    try:
+        mesh._deform_mesh(coords, verbose=verbose)
+    except Exception:
+        # The user's write landed in the canonical array BEFORE this
+        # callback ran. A rejected/failed deform must not leave
+        # mesh.X.coords disagreeing with the DM — the guard's message
+        # says the write "is rejected", so make that true by restoring
+        # the canonical from the DM, then let the error surface.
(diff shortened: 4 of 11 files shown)


lmoresi:

Two independent reviewers (numpy-subclass semantics; codebase regressions) attacked this branch. The core guard survived everything thrown at it, with a structural reason worth recording: **numpy's view base-collapse only walks past bases of matching type, and the canonical array's own base is always a plain ndarray — so the base chain always stops exactly at the canonical.** Verified empirically for slice/nested/reshape/transpose/newaxis/dtype-changed views, zero-size slices on empty selections, masked and integer fancy writes (exactly one firing via parent write-back, including when the parent is itself a view), and detached copies (correctly silent — previously these packed the copy's wrong-sized contents, so skipping is strictly better). The deleted factories are confirmed dead (no caller in src/ or tests/), and `old_value` is confirmed unread everywhere.

### Findings to fix (this branch)

1. **Raise-after-mutation poisons `mesh.X.coords` (should-fix).** `__setitem__` writes before notifying, so now that callback exceptions propagate, the coordinate-lock rejection ("an in-place write to mesh.X.coords is rejected") raises AFTER the values landed in `mesh._coords` — the error claims rejection while the mesh's coordinate cache now disagrees with the DM. `tests/test_0540_coordinate_change_locking.py` does exactly the catch-and-continue this invites. Fix: the rejection path restores `_coords` from the DM before raising, making the message honest.

2. **The delayed path still swallows exceptions (should-fix).** The commit message says "exceptions propagate", but only the immediate path was fixed — the delayed executors (which `mesh.access()` still routes through) keep the `logger.warning` swallow. The stacked #383 rewrites this machinery and removes the swallow, but this branch should not ship a claim it doesn't hold on its own. Fix: remove the two delayed-path swallows here too.

3. **`sync_data` (different-size branch) orphans canonical callbacks (should-fix, latent).** The new object copies the callback list, but the copied dispatch weakrefs the OLD canonical — every future write to the returned array is classified as a copy and silently skipped. No in-tree caller pairs the two today (the only `sync_data` caller is the deprecated `swarm.points` wrapper, removed by #382), but `sync_data` is exactly the resize-after-migration idiom canonical arrays would reach for. Fix: re-register canonical callbacks against the new object (and make `remove_callback` able to remove a canonical callback by its original function, closing the related API wart).

4. **Governing docs drifted (should-fix, docs).** `docs/developer/subsystems/data-access.md` (the authoritative document) still points at the deleted `_create_variable_array()`; `UW3_Developers_NDArrays.md` shows the removed `old_value` snapshot as current internals; `UW3_Style_and_Patterns_Guide.md` documents `old_value` as carrying previous values.

5. **Swarm canonical callback contradicts the new no-cycle comment (note→fix).** It closes over the variable strongly (pre-existing, gc-collectable), while the mesh side uses `array.owner`. Aligning it on `array.owner` makes the comment's guarantee true.

6. **Test gap (should-fix).** No test pins the new propagate behaviour or the `old_value is None` contract — the fix half of the third commit ships without the test that would catch its reversion.

### Recorded, no action

- `_is_canonical`/`_canonical_ref` tags are unused within this branch alone (they become load-bearing in #383) — dissolves when the stack lands together.
- Pickling of arrays carrying closures was equally impossible before this branch.

Fixes follow on this branch; the stacked branches will be rebased over them.

lmoresi:
## Round 1 — response (fixed in 1f421284)

All six findings addressed:

1. **Poisoned coords on rejection** — the coordinate callback now restores the canonical from the DM before the rejection surfaces, so `mesh.X.coords` always agrees with the DM after a refused write. Pinned by a new asserting test in `test_0540` (that module previously only printed).
2. **Delayed-path swallow** — removed; both executors now propagate, so the exceptions-propagate contract holds on this branch alone, including via `mesh.access()`.
3. **`sync_data` orphaning** — the different-size branch re-registers canonical callbacks against the new object (the dispatch wrapper records its original function as `_wrapped`). Pinned by `test_sync_data_resize_rehomes_canonical_callbacks`.
4. **`remove_callback` wart** — now removes a canonical callback by its original function.
5. **Swarm callback cycle** — resolves the variable via `array.owner` like the mesh side; the no-cycle rationale in the guard's docstring is now honoured by all three registration sites.
6. **Docs** — `data-access.md`, `UW3_Developers_NDArrays.md`, and the patterns guide updated; new tests pin propagation and the `old_value = None` contract.

Quick gate: 409 passed, 1 pre-existing xfail. The stacked branches (#382, #383, #385) will be rebased over this; their own round-1 fixes follow on their threads.

lmoresi:
## Round 2 — all round-1 fixes verified closed (plus one diagnostics guard, ca406f14)

The verifier confirmed: the coord-rejection restore is genuine (the cache never aliases the live DM buffer, so the rollback is real, shapes verified on bare and variable-carrying meshes, test run on two builds); the delayed-path swallow removal, `sync_data` re-homing (order-preserving), `remove_callback`-by-original, and the swarm `array.owner` rewrite all hold. One small addition from a round-2 note: the restore now runs only when cache and DM shapes agree, so a failure that replaced the coordinate Vec at a different size (mid-rebuild) cannot mask the original error with a broadcast exception.

PR #391, opened 2026-07-22
fix(evaluate): recover boundary-face query points dropped by DMLocatePoints on quad/hex meshes (#390)

### LINKED ISSUE #390: test_1052 VEP stability tests fail under PETSc 3.25 (variable-dt pure-VE blow-up), pass in level_1 gate
## Summary

The three `level_2, tier_a` tests in `tests/test_1052_VEP_stability_regression.py` fail
in a local `amr-dev` worktree running **PETSc 3.25.0** (numpy 2.4.6). They are **not** in
the `level_1 and tier_a` quick gate, so CI on `development` stays green — but they are
`tier_a` (hardened) and should pass.

Failing tests:
- `test_pure_ve_variable_dt_accuracy` — **`max|err| = 240.24` vs threshold `0.10`**
  (pure viscoelastic, variable timestep; a blow-up, not drift)
- `test_vep_yield_lock_fixed_dt`
- `test_vep_snes_no_divergence_loading_through_yield`

## This is environment/PETSc, not code

- The failure is **byte-identical** (`max|err| = 240.24221587183746`) on both the current
  `development` `constitutive_models.py` and a feature branch — swapping the source does
  not change it.
- It **persists through a clean `build/` wipe** (`rm -rf build/lib.* build/bdist.*` +
  rebuild), so it is not a stale-build artifact.
- The pure-VE test returns *before* any yield / `_combine_yield` path, so recent yield-law
  work cannot be the cause.

Taken together: the failing behaviour is intrinsic to this worktree's environment
(**PETSc 3.25.0**), not to any code change on `development`.

## Reproduce

```bash
pixi run -e amr-dev python -m pytest \
  "tests/test_1052_VEP_stability_regression.py::test_pure_ve_variable_dt_accuracy" -q
# -> AssertionError: max|err|=240.2422 (threshold 0.10; current head ~0.06)
```

## Hypothesis / suggested investigation

- Most likely a **PETSc 3.24 → 3.25** behaviour change reaching the variable-dt VEP path
  (a SNES/line-search/TS default, or a `constants[]` / DS interaction), rather than a UW3
  regression. `reference_env_numpy25_petsc4py_gmsh` notes a `PETSc3.24→3.25 uw-rebuild`.
- Confirm by running the same test against a PETSc 3.24 build; if it passes there, pin the
  changed PETSc default.
- If it is a genuine variable-dt VEP accuracy regression, that is worth fixing on its own
  merits; if it is an env/threshold sensitivity, the test may need an env guard or a
  documented tolerance.

## Notes

Discovered while root-causing an unrelated consistent-Newton Drucker–Prager Jacobian
recursion (fixed separately). These VEP-stability failures are pre-existing and orthogonal
to that work; filing for a dedicated session.

lmoresi:
## What this fixes

Closes #390 (the test_1052 VEP stability failures) — but the root cause is not PETSc 3.25.

`uw.function.evaluate` was returning **silent zeros** for query points sitting exactly on the domain's closed upper boundary faces of quad/hex meshes (e.g. y = y_max on a `StructuredQuadBox`). PETSc's point location uses a half-open cell convention, so a point exactly on the top face belongs to the nonexistent cell "above" and gets dropped; the dropped point was then zero-filled by the evaluator.

Semi-Lagrangian stress-history trace-backs slide departure points along the top boundary of the VEP shear box, so every step read ψ* = 0 there instead of the true history. The BDF history update amplified that boundary corruption exponentially: max|err| = 240 against a threshold of 0.10. With this fix the same run gives max|err| = 0.063 — the value recorded in the test docstring when it was written.

## Why this is a UW3 regression, not a PETSc one

- Before 17a5a8d3 (May 24) the kd-tree cell hint was always passed into `DMInterpolationSetUp_UW`, and it prefilled the dropped-point recovery map — points PETSc dropped were evaluated in the adjacent hinted cell (with reference-coordinate clamping). When the hint pointer became the DMLocatePoints-bypass switch (simplex/manifold only), quad/hex volume meshes lost the recovery.
- PETSc's `plexgeometry.c` (point location, grid hash) has only cosmetic changes between v3.24.2 and v3.25.0.
- The failure reproduces byte-identically with conda-forge PETSc 3.25.3 and the custom 3.25.0 build, on any source branch that postdates May 24.

## The fix

`DMInterpolationSetUp_UW` gains an explicit `hintAuthoritative` flag, separating the hint's two roles:

- **simplex / manifold meshes**: hint is authoritative, DMLocatePoints is bypassed — unchanged behaviour;
- **quad / hex volume meshes**: DMLocatePoints remains authoritative (the kd-tree nearest hint can be wrong on deformed cells), but the hint again rescues points it drops — restored behaviour.

For located points nothing changes: DMLocatePoints results always overwrite the hint. The evaluator's wrapper (`_dminterp_wrapper.pyx`) now always passes the hint and only varies the flag.

## Testing

- New level_1/tier_a regression test: linear-field evaluation at points on every closed face of quad and simplex boxes (would have caught this immediately — the top face returned zeros).
- `tests/test_1052_VEP_stability_regression.py`: 5/5 pass (was 3 failing).
- level_1 + tier_a gate: 437 passed.
- Parallel suite (`mpirun -np 2`): 92 passed; the single failure (`test_0765` deformed-shell) is pre-existing and carries its own `TODO(BUG)` note about the PR #326 `_deform_mesh` guard.

### DIFF
--- a/src/underworld3/function/_dminterp_wrapper.pyx
+++ b/src/underworld3/function/_dminterp_wrapper.pyx
@@ -34,7 +34,7 @@ cdef extern from "petsc.h" nogil:
 
 # Import custom UW routines
 cdef extern from "petsc_tools.h" nogil:
-    PetscErrorCode DMInterpolationSetUp_UW(DMInterpolationInfo ipInfo, PetscDM dm, int petscbool, int petscbool, size_t* owning_cell)
+    PetscErrorCode DMInterpolationSetUp_UW(DMInterpolationInfo ipInfo, PetscDM dm, int petscbool, int petscbool, size_t* owning_cell, int petscbool)
     PetscErrorCode DMInterpolationEvaluate_UW(DMInterpolationInfo ipInfo, PetscDM dm, PetscVec x, PetscVec v)
 
 cdef class CachedDMInterpolationInfo:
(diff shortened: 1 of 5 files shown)


lmoresi:
Design follow-up agreed with Louis: #391 stays scoped to restoring the hint recovery (exact for regular/near-regular quad geometry). The general resolution — NaN sentinel instead of zero-fill, plus an RBF fallback rung for deformed quad/hex meshes where the kd-tree hint is not trustworthy — is tracked as #392.

Issue #392, opened 2026-07-22
Point-location fallback ladder: NaN sentinel + RBF rung for deformed quad/hex meshes


lmoresi:
Follow-up to #390 / #391, from design discussion with Louis (2026-07-23).

#391 restored the hint-cell recovery for points that DMLocatePoints drops on quad/hex meshes. That is exact for regular and near-regular geometry (the adjacent cell's FE basis with clamped reference coordinates reproduces the boundary trace to machine precision), but on heavily deformed quad meshes the kd-tree nearest-centroid hint can select a wrong cell and return a plausible-looking wrong value. Quads need a fallback that is appropriate for them and unnecessary for simplices.

## Agreed resolution: a fallback ladder with no silent values

1. **Simplex / manifold** — barycentric hint is authoritative, DMLocatePoints bypassed. No fallback needed: every in-domain point locates by construction.
2. **Quad/hex, point located by PETSc** — normal FE evaluation (the common case, unchanged).
3. **Quad/hex, point dropped, near-affine geometry** (rectilinear boxes, cubed-sphere-class distorted hexes) — hint cell + reference-coordinate clamp, as restored in #391. Exact for boundary-trace queries; keep it above RBF because semi-Lagrangian history reads (the #390 loop) are sensitive to systematic per-step smoothing error.
4. **Quad/hex, point dropped, deformed geometry** (hint not trustworthy) — route through the existing RBF evaluation path: bounded, mesh-topology-free, honest about being approximate. This rung is currently missing.
5. **Genuinely out-of-domain** — existing RBF/extrapolation path (unchanged).

## Implementation sketch

- Replace the explicit zero-fill for unlocated points in `DMInterpolationEvaluate_UW` with a **NaN sentinel**, and surface a not-located mask from the wrapper. Any future gap in the ladder is then loud instead of silently plausible.
- `evaluate_nd` routes masked points through the RBF machinery it already uses for exterior points (per-rank local, consistent with the COMM_SELF evaluator).
- Rung 3 vs 4 selection belongs on a mesh capability flag ("hint-authoritative geometry": affine / near-affine reference maps) rather than another `isSimplex()` fork — the same regular/near-regular boundary as the simplex-first design direction recorded in the planning file. `StructuredQuadBox` (undeformed) and cubed-sphere-class meshes qualify; a mesh that has been through `deform()` with non-affine distortion does not.

## Acceptance

[...]

PR #394, opened 2026-07-22
feat(evaluate): measured point-location capability + NaN/RBF fallback ladder

lmoresi:
Implements #392 (stacked on #391 — retarget to `development` once that merges; the diff shown here is the capability-ladder work only).

## What this does

Makes the cell locator's authority (i.e. is it authoritative or an estimator)  a **measured mesh property** instead of an `isSimplex()` type flag, and completes the fallback ladder agreed on #392 so quad/hex meshes keep working with a fallback that simplex meshes don't need.

**`Mesh._location_capability()`** audits face planarity (sagitta) and cell convexity, cached against `(_mesh_version, _topology_version)` so `deform()` and adaptation refresh it automatically:

- **`exact`** — simplex, manifold, and any quad/hex mesh whose faces are planar and cells convex at machine level. This graduates **all valid 2-D quad meshes, deformed included** (straight edges are always planar; convexity is equivalent to an untangled mesh — verified 4000/4000 against exact polygon containment on a deformed box) and rectilinear/affine hex boxes to simplex-grade location: the hint bypasses `DMLocatePoints`.
- **`continuous`** — warped hexes within a 5% sagitta tolerance (cubed-sphere class: 64% of faces exactly planar, spherical faces ≤ 1e-2). Authoritative for continuous fields only — a misclassified point lands in the face-adjacent neighbour where continuous FE interpolants agree to O(sagitta × gradient); a face-aligned jump (layered viscosity on a spherical shell) would see O(jump) errors.
- **`none`** — badly warped / non-convex cells: no authority.

**The ladder**: non-authoritative evaluations (capability `none`, or `continuous` with a discontinuous variable in the mix) run `DMLocatePoints`; points it drops (the #390 class — queries exactly on the domain's closed upper faces) are **NaN-filled — never zeros** — reported via a new unlocated mask, and filled per-variable through `rbf_interpolate`: bounded, topology-free, honest. NaN survives only if the plumbing is bypassed, which is exactly when it should be visible.

The policy lives in one place (`Mesh._hint_is_authoritative`) and joins the DMInterpolation cache key (same coords with a different field-continuity mix legitimately need different structures).

## Testing

- Three new tier_a tests in test_0503: capability grading of regular/deformed quad and hex meshes; deformed-quad boundary evaluation exact to 1e-8; warped-hex RBF rung (no NaN escapes, no silent zeros, values bounded by data range).
- `test_1052` VEP stability: 5/5 (the shear box now takes the authoritative bypass).
- level_1 + tier_a gate: 440 passed.
- Parallel np2 suite: identical to baseline (92 passed; the single failure is the pre-existing `test_0765` `_deform_mesh`-guard issue with its own TODO).

Design record: issue #392 and `docs/developer/design/point-location-capability.md`.

### DIFF
--- a/.github/workflows/build_uw3_and_test.yaml
+++ b/.github/workflows/build_uw3_and_test.yaml
@@ -44,5 +44,12 @@ jobs:
         # editable installs are project policy violation, see CLAUDE.md).
         run: pixi run -e dev build
 
+      # TEMPORARY (PR #394 debug): the test step was dying with no pytest
+      # output — a hard crash during `import underworld3` before the first
+      # stdout flush. Surface the crash point with faulthandler; remove this
+      # step once the cause is fixed.
+      - name: Import smoke test (faulthandler)
+        run: pixi run -e dev python -u -X faulthandler -c "import faulthandler; faulthandler.enable(); import underworld3; print('import OK', underworld3.__version__)"
+
       - name: Run tests
         run: pixi run -e dev ./scripts/test.sh --p 2
--- a/.github/workflows/build_uw3_and_test.yaml
+++ b/.github/workflows/build_uw3_and_test.yaml
@@ -44,10 +44,14 @@ jobs:
         # editable installs are project policy violation, see CLAUDE.md).
         run: pixi run -e dev build
 
-      # TEMPORARY (PR #394 debug): the test step was dying with no pytest
-      # output — a hard crash during `import underworld3` before the first
-      # stdout flush. Surface the crash point with faulthandler; remove this
-      # step once the cause is fixed.
+      # Import smoke test: a hard crash during `import underworld3` (segfault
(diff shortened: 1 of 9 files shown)


lmoresi:
CI resolved — the failure was a runner-side flake, not the code. The failing run died during `import underworld3` before pytest's first stdout flush (every batch ~2s, zero output, undiagnosable from the log). A faulthandler import smoke step was added to surface the crash point; on the very next run the identical code imported cleanly and the full test job passed (25 min, green). The smoke step is kept as a permanent 5-second guard so any future import-time crash is loud and localised. Local reproduction attempts on both custom PETSc 3.25.0 and conda-forge PETSc 3.25.3 (full CI batch selection via scripts/test.sh) were green throughout.

Issue #400, opened 2026-07-23
_tensor_stats crashes for genuine (d,d) tensor variables: flat component index used on the structured array axis


lmoresi:
Found by the public adversarial review of the #330/#397/#384 batch (2026-07-23).

`MeshVariable._tensor_stats` (discretisation_mesh_variables.py) reads components as `self.array[:, 0, i]` for `i in range(self.num_components)` — but for a rank-2 tensor variable `num_components = d*d` (4 in 2D) while `self.array` has structured shape `(N, d, d)`, so the loop indexes 4 against an axis of size 2:

```
IndexError: index 2 is out of bounds for axis 2 with size 2
```

Demonstrated live on a (2,2) tensor variable. The Frobenius accumulation should read the flat layout (`self.data[:, i]`) or iterate the structured axes. Until then, `stats()` on any true TENSOR/SYM_TENSOR variable raises — the method is effectively dead code for its own case (vector variables are unaffected; they take `_vector_stats`).

Pre-existing on development — the #330-batch diff only renamed the temporary variable inside it.

Underworld development team with AI support from Claude Code

lmoresi:
Fixed by PR #406, merged to development on 2026-07-24 after a public adversarial review round. (Manual close — PRs target development.)

Underworld development team with AI support from Claude Code

Issue #399, opened 2026-07-23
Mesh construction crashes on empty ranks: navigation kd-tree build passes a 1-D empty array (StructuredQuadBox(elementRes=(2,2)) at np4)


lmoresi:
Found during the #330 batch (2026-07-23): fixing the Swarm.save race unmasked this in `tests/test_0003_save_load.py` at np4 — the file previously hung before reaching it.

Any mesh construction where a rank ends up with **zero local cells** crashes:

```
mpirun -np 4 python -c "import underworld3 as uw; uw.meshing.StructuredQuadBox(elementRes=(2,2))"
# rank with no cells:
#   ValueError: Buffer has wrong number of dimensions (expected 2, got 1)
#   at src/underworld3/ckdtree.pyx:99, via _build_kd_tree_index
#   (discretisation_mesh.py:5154); other ranks then die on MPI_ERR_BUFFER
#   from the resulting collective divergence.
```

Two layers to it:

1. `_build_kd_tree_index` builds `numpy.array(control_points_list)` / `numpy.array(centroids_list)` — empty lists give 1-D `(0,)` arrays, which the KDTree memoryview (`double[:,::1]`) rejects.
2. Shaping them `(0, dim)` is not enough on its own: `KDTree.__cinit__` takes `&points[0][0]` (ckdtree.pyx:104), which is invalid for zero rows — the constructor needs an explicit empty-index path (related to the #313 empty-rank KDTree *query* guards, which protected callers but not construction).

Verified identical on unmodified development (`54a4fc58`). Reproduces with any partition that starves a rank of cells — tiny meshes are just the easy trigger; extreme partition imbalance on real meshes would hit the same wall.

Also worth noting: `tests/test_0003_save_load.py` runs cleanly under MPI up to this point once #330 is fixed; this is the remaining blocker to making that file part of a parallel gate.

Underworld development team with AI support from Claude Code

lmoresi:
Fixed by PR #406, merged to development on 2026-07-24 after a public adversarial review round. (Manual close — PRs target development.)

Underworld development team with AI support from Claude Code

PR #401, opened 2026-07-23
Swarm/mesh checkpoint quiescence, .msh coordinate-system default, rank-symmetric stats temporaries

### LINKED ISSUE #330: Swarm.save: non-zero ranks return while rank 0 still appends metadata — immediate reopen hits HDF5 file locking (BlockingIOError)
## Summary

Found during #329's gating: `Swarm.save` returns on non-zero ranks while rank 0 is still appending metadata to the HDF5 file. A caller that immediately reopens the file (e.g. read_timestep straight after write_timestep) hits HDF5 file locking — BlockingIOError errno 35.

A TODO(BUG) marker sits at the metadata append in swarm.py (added in #329); the new parallel read_timestep tests barrier after write_timestep to stay clear of it.

## Suggested fix

Barrier before returning from save (or move the metadata append before the collective close). Regression: write-then-immediately-read at np2 without an explicit user barrier.

### LINKED ISSUE #384: Rank-asymmetric temporary variable names in _vector_stats/_tensor_stats (_temp_mag_{id(self)})
Found while building the #379 synchronised_array_update rework (PR #383).

`_vector_stats` and `_tensor_stats` (discretisation_mesh_variables.py) create temporary mesh variables named `_temp_mag_{id(self)}` / `_temp_frob_{id(self)}`. `id()` is a rank-local memory address, so these variable names differ across ranks. Variable creation and deletion perform collective DM operations keyed by field name, and the variables are written inside `uw.synchronised_array_update` — any code path that uses the NAME as a cross-rank key breaks on them. This is also why the #379 collective flush uses creation-order registration ids rather than names.

Fix direction: derive the suffix from something rank-symmetric (a per-mesh monotonically increasing counter, or the parent variable's clean_name).

### LINKED ISSUE #397: mesh.write() crashes for a mesh loaded directly from a .msh file (CoordinateSystemType is None)
Found while building the #269 reproducer matrix (2026-07-23).

`uw.discretisation.Mesh("file.msh")` — loading a gmsh file directly, without passing `coordinate_system_type=` — leaves `mesh.CoordinateSystemType = None`. Any subsequent `mesh.write()` / `mesh.write_timestep()` then crashes writing the metadata block in `Mesh.write`:

```python
coordinates_type_dict = {
    "name": self.CoordinateSystemType.name,   # AttributeError: 'NoneType' object has no attribute 'name'
    "value": self.CoordinateSystemType.value,
}
```

Reproducer:

```python
import underworld3 as uw
tmp = uw.meshing.UnstructuredSimplexBox(minCoords=(0,0), maxCoords=(1,1), cellSize=0.2, filename="/tmp/box.msh")
del tmp
mesh = uw.discretisation.Mesh("/tmp/box.msh")
mesh.write("/tmp/box.h5")   # AttributeError
```

Workaround: load via `_from_gmsh` + `Mesh(plex, coordinate_system_type=uw.coordinates.CoordinateSystemType.CARTESIAN, ...)` (the Spiegelman-benchmark pattern), which sets the type explicitly.

Suggested fix: either default the coordinate system to CARTESIAN when loading a plain gmsh file (matching what the `uw.meshing.*` constructors do), or make `Mesh.write` tolerate/refuse-legibly a missing coordinate system rather than tracebacking in the metadata writer.

lmoresi:
Three fixes from the triage backlog, plus an adversarial-review round and its responses. Closes #330, closes #397, closes #384.

## The fixes

**#330 — save() races its own metadata append** (`fix(swarm)` + review follow-up)
`Swarm.save` and `SwarmVariable.save` appended rank-0 metadata after the data write and returned immediately on the other ranks; an immediate reopen (read_timestep after write_timestep) hit HDF5 file locking (BlockingIOError errno 35) and the rank divergence hung the next collective. Both methods now end with a barrier so the file is quiescent on return. The review found the identical pattern in `MeshVariable.save` and `mesh.write` — both now carry the same barrier, and both metadata blocks use context managers so an exception cannot leak the HDF5 handle.

`tests/test_0003_save_load.py` — previously 1 failure + a hard hang at np2 (the swarm tests were never run under MPI by any gate) — now passes 10/10 at np2 and np4 in ~5 s. This unmasked a further pre-existing crash (empty-rank navigation kd-tree, tiny meshes at np4), filed as #399 with a live reproducer; the affected test now uses a 4x4 mesh so the whole file stays MPI-clean.

**#397 — mesh loaded from a .msh file couldn't be written**
`CoordinateSystemType` stayed `None` and `mesh.write()` crashed in its metadata block. The construction now defaults to CARTESIAN — matching every `uw.meshing` constructor, and matching what `CoordinateSystem` already did at runtime (a None-typed mesh was always functionally Cartesian; the review verified no consumer anywhere treats None specially). Per the review: loading an h5 checkpoint that has *lost* its coordinate metadata now warns before defaulting, so a wrong label cannot launder itself silently into re-written files (pre-fix, the crash was the tripwire).

**#384 — rank-asymmetric temporary names in vector/tensor stats**
The stats temporaries were named with `id(self)` — a rank-local address — while variable creation/destruction runs collective DM operations keyed by field name. The suffix is now the parent's `clean_name` (identical on every rank). Per the review: both sites now refuse a pre-existing temp name outright — a collision would silently alias, overwrite, and deregister a user's variable (demonstrated live in the review round).

## Adversarial review (public, round 1)

Posted in full as the first comment below. Summary: all four primary fixes **held** against rank-asymmetric control-flow tracing, barrier/file-close ordering, zero-particle ranks, and live np4 probes of both I/O branches with back-to-back write/read cycles. It caught three things that are fixed in the response commit (temp-name aliasing, the wider #330 pattern, the silent-default loudness), and one pre-existing defect outside the diff, filed as #400 (`_tensor_stats` indexes the flat component count against the structured array axis — crashes for any true tensor variable).

## Verification

- `test_0003_save_load.py`: 10/10 at serial, np2, np4 (first time this file is MPI-clean end to end).
- `tests/parallel/test_0750_global_statistics.py`: 12 passed at np2 and np4.
- Level_1 + tier_a gate: **442 passed**, 0 failures (run after the review-response commit).
- New issues filed on the way: #399 (empty-rank kd-tree construction crash, with baseline repro on unmodified development), #400 (tensor-stats indexing).

Reminder for the merge: manual close needed for #330, #397, #384 (PRs target development; auto-close does not fire).

### DIFF
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -1223,6 +1223,14 @@ class Mesh(Stateful, uw_object):
         # A unique set of vectors / names for each mesh instance
         #
 
+        # A mesh constructed without an explicit coordinate system (e.g.
+        # loaded directly from a .msh file, or from an h5 checkpoint with no
+        # coordinate metadata) is Cartesian — the default every uw.meshing
+        # constructor passes. Leaving None here crashed mesh.write()'s
+        # metadata block (issue #397).
+        if coordinate_system_type is None:
+            coordinate_system_type = CoordinateSystemType.CARTESIAN
+
         self.CoordinateSystemType = coordinate_system_type
 
         from sympy.vector import CoordSys3D
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -795,7 +795,22 @@ class Mesh(Stateful, uw_object):
                     coord_type_dict = json.loads(json_str)
                     coordinate_system_type = CoordinateSystemType(coord_type_dict["value"])
                 except KeyError:
-                    pass
+                    # A checkpointed mesh normally carries its coordinate
+                    # system; a missing entry means the metadata was lost
+                    # (or the file predates it). The construction defaults
+                    # to CARTESIAN (#397) — say so, because for an annulus
+                    # or spherical checkpoint that label is wrong and would
+                    # otherwise propagate silently into re-written files.
+                    if coordinate_system_type is None and uw.mpi.rank == 0:
+                        import warnings
+
+                        warnings.warn(
+                            f"{plex_or_meshfile} has no coordinate_system_type "
+                            "metadata; defaulting to CARTESIAN. Pass "
+                            "coordinate_system_type= explicitly if this mesh "
+                            "is not Cartesian.",
+                            stacklevel=2,
+                        )
 
                 regions = None
                 try:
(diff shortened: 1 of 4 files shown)


lmoresi:

Semi-independent skeptical pass over the code changes, instructed to refute/rebut each fix. Live probes ran at np4 in the branch worktree environment (both the mpio and force_sequential I/O branches, zero-particle ranks, and back-to-back write/read cycles).

### Attack 1 — #330 barriers: can any rank miss them (new deadlock)?
**HELD.** Every exit path in both save methods was traced: one return each, after the barrier; the only raise (the .h5 suffix check) is argument-driven and rank-symmetric for every in-repo caller. Branch selection (mpio vs sequential) depends only on a build constant and a same-valued argument, so no rank can diverge. Both paths were already unconditionally collective before this change (allgather / collective create / barrier loop), so the added barrier introduces no new collectivity requirement. Verified live at np4 with zero-particle ranks and three consecutive write/read cycles on both branches: no hang, no BlockingIOError.
*Documented residual*: if rank 0 raises inside its solo metadata block (disk full, external lock), other ranks now hang at the barrier rather than desynchronising later — both outcomes are fatal under MPI; noted, accepted.

### Attack 2 — barrier vs h5py close ordering
**HELD.** The rank-0 metadata append is inside a context manager, closed strictly before the barrier; non-zero ranks cannot pass until rank 0 arrives, i.e. after close. `uw.selective_ranks` contains no collectives and does not interact.

### Attack 3 — #384 clean_name collisions / cleanup lifecycle
**HELD for the fix — but the review DEMONSTRATED a new corruption surface**: a user variable literally named `_temp_mag_<X>` would be silently aliased by variable creation, its data overwritten by the magnitude (probe: 42.0 → 1.414), and then deregistered. Unreachable under the old id(self) names. **Fixed in the response commit**: both stats sites refuse a pre-existing temp name. Repeated stats calls, exception-path cleanup, and cross-variable uniqueness (mesh.vars keyed by clean_name) all verified live.

### Attack 4 — #397 None→CARTESIAN semantic drift
**HELD.** All construction sites grep-verified: every uw.meshing constructor passes an explicit enum; adapt/submesh children inherit the parent's. Pre-fix, `CoordinateSystem(mesh, None)` already fell to the Cartesian-by-default branch — None-typed meshes were functionally Cartesian everywhere at runtime, and no consumer special-cases None. The one real behaviour change found: an annulus/spherical h5 that LOST its metadata used to crash write() (a loud tripwire) and would now silently re-write a CARTESIAN label into descendant files. **Fixed in the response commit**: a rank-0 warning fires when the default engages on a metadata-less h5.

### Attack 5 — the 2x2→4x4 test mesh change
**HELD.** The guarded behaviour (missing-directory raise, before any collective) is resolution-independent. The #399 justification verified live: elementRes=(2,2) at np4 deadlocks mesh construction on unmodified development.

### Findings outside the diff
1. `_tensor_stats` crashes for genuine (d,d) tensor variables — flat component count indexed against the structured array axis (pre-existing; the diff only renamed its temporary). Filed as #400.
2. The #330 pattern also existed in `MeshVariable.save` and `mesh.write` (no trailing barrier; metadata handles not exception-safe — mesh.write never closed its handle on the exception path). **Fixed in the response commit.**
3. Cosmetic, pre-existing: both swarm save() signatures annotate `filename: int`; left untouched (out of scope).

### Response commit
a559e477 carries all three fixes; the level_1+tier_a gate (442 passed) and the np2/np4 test batteries were re-run after it.

Issue #405, opened 2026-07-24
Empty-rank support, layer 2: evaluate/points_in_domain, radii/centroid reductions, gather_data NaN-stripping


lmoresi:
Umbrella follow-up from the public adversarial review of the #399/#400/#351/#386 batch (2026-07-24). That batch makes starved-rank meshes constructible and solvable (PETSc handles empty ranks fine), fixes the nearest-centroid particle mis-route, and makes the kd-tree sentinel guards honest. The next layer breaks at the first point evaluation:

1. **`uw.function.evaluate` on a starved rank**: `points_in_domain` calls `get_max_radius`, which does `self._radii.max()` on an empty array → ValueError on the starved rank while populated ranks sit in `gather_data` → deadlock. `get_min_radius` fails identically and feeds every solver's `estimate_dt` and penalty scaling. Needs empty-safe local reductions with defined semantics (e.g. `points_in_domain` short-circuits to all-False with no radii access; min/max radius return identity elements and callers reduce globally).
2. **Consequence for field-valued metrics**: `mesh_metric_mismatch` skips `evaluate` on starved ranks (it must — see 1), so a metric containing MeshVariable data deadlocks at np>1 with a starved rank. Documented as a known limit in the function; lifts automatically once (1) is fixed.
3. **`gather_data` silently strips NaN rows** (`utilities/_utils.py`): any consumer whose local reduction produces NaN on an empty rank gets a compacted table whose row index no longer equals rank — this was the mechanism behind the particle mis-route. Either stop stripping (return the full table and let callers mask) or make the stripping opt-in and loudly documented.
4. **Observed hazard class**: when a rank-asymmetric raise occurs and user code catches it, the survivors' pending collective pairs with the *next* collective the failed ranks enter — both 'succeed' with garbage (observed live: a 1-row centroid table with no error). Worth a note in the parallel-patterns doc; the cure is the same as everywhere — no rank-asymmetric raises around collectives.
5. Smaller members of the family: `MeshVariable`'s rbf path and the monotone-clamp kNN (`functions_unit_system.py`) index empty source arrays without a guard (SwarmVariable's rbf path IS guarded).

Related: #314 (global_evaluate np4 hang on empty-interior ranks) is likely the same family seen from the DMInterpolation side — a fix for (1) should be validated against #314's reproducer.

Underworld development team with AI support from Claude Code

lmoresi:
[...]

Issue #314, opened 2026-07-04
petsc_interpolate: the DMLocatePoints call count is rank-local (cache key + unreduced location capability) — hangs at np>=4


lmoresi:
## Summary

`tests/parallel/test_0760_swarm_cache_migration.py` hangs at `mpirun -np 4` on current `development` (pre-existing — verified on a pristine baseline during the gating of #313, by stashing that branch's src changes and rebuilding).

## Mechanism

With rank-biased particle coordinates, `global_evaluate` leaves some ranks with **zero interior points**. Those ranks diverge inside the `DMInterpolation` machinery in `petsc_interpolate` while the remaining ranks block in its collectives — an MPI deadlock. It passes at np2, which is why CI (`test.sh --p 2`) never sees it.

Per-rank stack traces were captured during #313's gating and confirm the hang is inside the DMInterpolation path; the swarm-cache fixes in #313 enter and exit their collective hooks symmetrically before the hang (i.e. #313 neither causes nor masks this).

## Relation to the 2026-07 quality audit

Corresponds to the provisional findings SWARM-15/16 in `docs/reviews/2026-07/SWARM-SUBSYSTEM-REVIEW.md` (parallel-evaluation empty-rank asymmetry family; see also BF-16 in the remediation worklist — substep advection evaluating on the wrong rank).

## Reproduction

`mpirun -np 4 python -m pytest tests/parallel/test_0760_swarm_cache_migration.py` on development — hangs. Same at np2 passes.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
Root cause located, and #513 consolidated in here — one defect, two triggers.

**The number of `DMLocatePoints` calls on the world mesh DM inside `petsc_interpolate` is a rank-local function of that rank's coordinate array.** Two rank-local switches produce it:

- the `DMInterpolationCache` is keyed on a per-rank coords hash (`function/_function.pyx:1293`, `function/dminterpolation_cache.py:124-132`) — a hit skips `create_structure` → `DMLocatePoints` entirely, and hit-vs-miss is decided per rank;
- `_location_capability()` is deliberately not reduced (`discretisation/discretisation_mesh.py:6334-6339`), and the policy it yields is itself part of the cache key, so a capability split forces a hit on some ranks and a miss on others.

[...]

PR #406, opened 2026-07-24
Empty-rank mesh construction, tensor stats, global metric moments, swarm.points label

### LINKED ISSUE #351: mesh_metric_mismatch: rank-local A_target/A_mean sums make the returned δ moments partition-dependent
Found by the Copilot pass over Wave D PR #345 (2026-07-08); `TODO(BUG)` marker at `src/underworld3/meshing/smoothing/metrics.py` (commit 300bba41) with a fix sketch.

`mesh_metric_mismatch` builds `A_target` (and the clamp's `A_mean`) from rank-local sums, so under MPI the returned δ moments (rms/max/median_abs) are partition-dependent and inconsistent with the docstring's global equidistribution target. **The live skip signal (alignment/misalignment used by follow_metric/mmpde) IS globally reduced and unaffected** — this only mis-reports the diagnostic moments.

Fix: global reductions for the sums/counts feeding A_target/A_mean.

### LINKED ISSUE #386: swarm.points physical-coordinate scaling mislabels units (meter magnitudes labelled with mesh.units)
Found during the #379/#382 adversarial review (PR #382 discussion). The deprecated `swarm.points` / `swarm.data` snapshot scales model coordinates by `CoordinateSystem._length_scale` and wraps the result in `UnitAwareArray(..., units=self.mesh.units)`. When the mesh units are kilometres, the magnitudes produced by the scale factor are metres — a 1000x label/magnitude mismatch. Verified in the review's regime battery (max|diff| ~2.7e6 against the correct mesh-variable path). Pre-existing behaviour, faithfully preserved by the #382 read-only snapshot; the sanctioned `swarm.coords` path dimensionalises correctly and is unaffected.

### LINKED ISSUE #399: Mesh construction crashes on empty ranks: navigation kd-tree build passes a 1-D empty array (StructuredQuadBox(elementRes=(2,2)) at np4)

### LINKED ISSUE #400: _tensor_stats crashes for genuine (d,d) tensor variables: flat component index used on the structured array axis

lmoresi:
Four fixes from the triage backlog plus the public adversarial-review round and its responses. Closes #399, closes #400, closes #351, closes #386.

## The fixes

**#399 — mesh construction crashed on ranks owning zero cells.** Two layers: the navigation kd-tree build passed a 1-D `numpy.array([])` into the KDTree memoryview, and the KDTree constructor itself could not represent an empty cloud (`&points[0][0]`). Empty point sets are now first-class: `(0, dim)` clouds build no C++ index and answer queries honestly (found=False, infinite distances); 1-D input gets a legible error. `StructuredQuadBox(elementRes=(2,2))` now constructs, integrates and **solves** at np3/np4 (PETSc itself is fine with empty ranks). Follow-ups from the review, also in this PR: the three sentinel sanity guards were off-by-one (`>` vs `>=` — the empty-tree sentinel masqueraded as valid index 0), and `_get_domain_centroids` produced NaN for starved ranks, which `gather_data` silently stripped — de-aligning the rank/row table and making `_route_by_nearest_centroid` **silently migrate particles to a rank owning zero cells** (demonstrated live in the review; now a huge-finite sentinel keeps the table aligned and the starved rank unreachable).

**#400 — `stats()` crashed for every genuine tensor variable.** Two stacked defects: the Frobenius loop indexed the flat component count against the structured array axis, and the temp variable was built through the enhanced factory whose `__getattr__` blocks underscore delegation. The review then caught my first fix under-measuring SYM_TENSOR norms (off-diagonals must count twice) — the final form reads the structured `(N, d, d)` view, correct for both tensor kinds, validated against numpy.

**#351 — `mesh_metric_mismatch` moments were partition-dependent.** The equidistribution target and the returned moments are now globally reduced (median via gather-to-root); starved ranks participate in the reductions with empty contributions; non-simplex refusal is rank-symmetric via `mesh.isSimplex`. Known limit, documented in place: a metric containing MeshVariable data still cannot run with a starved rank — that needs the evaluate layer (#405).

**#386 — deprecated `swarm.points` labelled metre magnitudes with `mesh.units`.** Scaled values are now labelled `meter`, the `mesh.X.coords` convention; verified all three coordinate views agree physically under a km-scale model.

## Adversarial review (public, round 1)

Posted in full below. All four primary fixes held their core claims; the review found and demonstrated live: the SYM_TENSOR under-measurement, the silent particle mis-route, the sentinel off-by-ones, a deadlock in #351's empty path with field metrics, and — most valuable — the honest map of where empty-rank support *ends*: construction, solve, integrals, swarm populate and stats all work; the first `evaluate`/`estimate_dt` still fails (rank-asymmetric raise in `get_min/max_radius`). Everything in that next layer is scoped in **#405** (which also ties in #314).

## Verification

- Serial battery (kdtree, stats, swarm-units, follow-metric, smoothing): 56 passed.
- np4: test_0700 (incl. new starved-rank construction test) + test_0750_global_statistics (incl. new #351 partition-independence test): 27 passed; np2: 13 passed.
- Direct repro: `StructuredQuadBox(elementRes=(2,2))` constructs at np3 and np4 (crashed on development).
- Level_1 + tier_a gate after the response commit: **443 passed**, 0 failures.

Reminder for the merge: manual close needed for #399, #400, #351, #386.

### DIFF
--- a/src/underworld3/ckdtree.pyx
+++ b/src/underworld3/ckdtree.pyx
@@ -624,7 +624,9 @@ cdef class KDTree:
         # Note: query() returns sqr_dists=True by default, and we use the converted coords
         distance_n, closest_n = self.query(coords, k=nnn)
 
-        if np.any(closest_n > self.n):
+        # valid indices are 0..n-1; the empty-tree sentinel (0 with n=0)
+        # must trip this guard, so the comparison is >= (issue #399).
+        if np.any(closest_n >= self.n):
             raise RuntimeError(
                 "Error in rbf_interpolator_local_from_kdtree - a nearest neighbour wasn't found"
             )
--- a/src/underworld3/ckdtree.pyx
+++ b/src/underworld3/ckdtree.pyx
@@ -91,6 +91,16 @@ cdef class KDTree:
         import underworld3.function.unit_conversion as unit_conv
         self.coord_units = unit_conv.get_units(points_input) if unit_conv.has_units(points_input) else None
 
+        # A legible error beats the memoryview's "Buffer has wrong number
+        # of dimensions" — the common trigger is numpy.array([]) from an
+        # empty point list, which is 1-D (issue #399).
+        if points_input.ndim != 2:
+            raise RuntimeError(
+                f"KDTree points must be a 2-D (n_points, dim) array, "
+                f"got shape {points_input.shape}. (An empty point set must "
+                f"still be shaped (0, dim).)"
+            )
+
         # Extract raw numpy array for C++ interface
         cdef const double[:,::1] points
         if unit_conv.has_units(points_input):
@@ -101,7 +111,13 @@ cdef class KDTree:
         if points.shape[1] not in (2,3):
             raise RuntimeError(f"Provided points array dimensionality must be 2 or 3, not {points.shape[1]}.")
         self.points = points
-        self.index = new KDTree_Interface(<const double *> &points[0][0], points.shape[0], points.shape[1])
+        # An empty point cloud is legitimate — a parallel rank can own zero
+        # cells (issue #399). No C++ index is built (taking &points[0][0]
+        # would be invalid); the query methods return honest empties.
+        if points.shape[0] > 0:
+            self.index = new KDTree_Interface(<const double *> &points[0][0], points.shape[0], points.shape[1])
+        else:
+            self.index = NULL
 
         global _live_instances, _total_constructed
         _live_instances += 1
@@ -190,6 +206,8 @@ cdef class KDTree:
         """
         Build the kd-tree index.
         """
+        if self.index is NULL:
+            return
         self.index.build_index()
 
     def kdtree_points(self):
@@ -234,6 +252,14 @@ cdef class KDTree:
         indices  = np.empty(count, dtype=np.uint64,  order='C')
         dist_sqr = np.empty(count, dtype=np.float64, order='C')
         found    = np.empty(count, dtype=np.bool_,   order='C')
+
+        # Empty index (a rank owning zero points, issue #399): nothing can
+        # be found — report that honestly for every query.
+        if self.index is NULL or count == 0:
+            indices[...] = 0
+            dist_sqr[...] = np.inf
+            found[...] = False
+            return indices, dist_sqr, found
 
         cdef long unsigned int[::1]  c_indices = indices
         cdef            double[::1] c_dist_sqr = dist_sqr
@@ -280,6 +306,13 @@ cdef class KDTree:
 
         n_indices  = np.empty((coords.shape[0], nCount), dtype=np.uint64,  order='C')
         n_dist_sqr = np.empty((coords.shape[0], nCount), dtype=np.float64,  order='C')
+
+        # Empty index (a rank owning zero points, issue #399): no
+        # neighbours exist — infinite distances, sentinel indices.
+        if self.index is NULL or nInput == 0:
+            n_indices[...] = 0
+            n_dist_sqr[...] = np.inf
+            return n_indices, n_dist_sqr
 
         indices  = np.empty(nCount, dtype=np.uint64,  order='C')
         dist_sqr = np.empty(nCount, dtype=np.float64, order='C')
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -5803,7 +5803,9 @@ class Mesh(Stateful, uw_object):
 
         if len(model_coords) > 0:
             dist, closest_points = self._index.query(model_coords, k=1, sqr_dists=False)
-            if np.any(closest_points > self._index.n):
+            # >= : valid indices are 0..n-1, and the empty-tree sentinel
+            # (0 with n=0) must trip this guard, not index _indexMap (#399).
+            if np.any(closest_points >= self._index.n):
                 raise RuntimeError(
                     "An error was encountered attempting to find the closest cells to the provided coordinates."
                 )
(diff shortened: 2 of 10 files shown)


lmoresi:

Independent skeptic pass over the branch diff, instructed to REFUTE each fix; live probes at np2/np4 including starved-rank configurations.

### #399 — KDTree empty clouds
- **Class internals HELD**: every dereference of the C++ index audited (build_index, both query paths, dealloc-on-NULL verified via live-instance count); rbf interpolators reach the index only through the guarded queries.
- **Consumers BROKEN (fixed in response commit)**: `get_closest_cells` indexed `_indexMap` with the empty-tree sentinel — the sanity guard was off-by-one (`>` where `>=`), letting sentinel 0 pass as a valid index (demonstrated at np4: two ranks raise IndexError, two return cells). Same off-by-one at three sites; all now `>=`.
- **Silent mis-route BROKEN (fixed in response commit)**: `_route_by_nearest_centroid` at np4 on a 2x2 box silently migrated 18 particles to a rank owning zero cells — NaN centroid from the empty mean, silently stripped by `gather_data`, de-aligning the rank/row table. Now a huge-finite sentinel keeps rows aligned and starved ranks unreachable (finite, not inf: infinities poison kd-tree bounding boxes).
- **Honest map of what breaks NEXT on a starved rank**: construction OK, Integral OK, **Poisson solve OK**, swarm populate OK — the wall is the first `uw.function.evaluate` / `estimate_dt` (`get_min/max_radius` does an empty-array reduction, rank-asymmetric raise → deadlock). Scoped as #405.

### #351 — global metric moments
- Reductions, rms normalisation, median gather, alignment sums: **HELD** — all five moments bit-identical across ranks for analytic metrics.
- **BROKEN (documented limit + partial fix)**: a metric containing MeshVariable data deadlocks with a starved rank — evaluate is collective and the empty rank must skip it (calling it instead hits the #405 radii failure; broken either way until that layer lands). Documented in place; lifts with #405.
- **BROKEN (fixed in response commit)**: the non-simplex NotImplementedError had become rank-asymmetric (starved ranks sailed into the allreduce while populated ranks raised — hang demonstrated at np4). Now refused rank-symmetrically via `mesh.isSimplex` before any collective.

### #400 — tensor stats
- Non-symmetric TENSOR vs numpy: **HELD** to 1e-15, including at np4 with starved ranks; temp lifecycle clean.
- **SYM_TENSOR BROKEN (fixed in response commit)**: the flat read counted each Voigt component once — measured 'Frobenius' 2.536 vs true 3.232. The structured (N,d,d) read counts mirrored off-diagonals twice; regression test pins the exact value.

### #386 — swarm.points label
**HELD**: all three coordinate views (points / coords / mesh.X.coords) agree physically under a 1000-km-scale model; the save path is confirmed migrated off self.points; no consumer depends on the old label. Two test-hygiene nits (deprecated kwarg, guarded assertion) fixed in the response commit.

### Outside the diff
`gather_data` NaN-stripping as a general landmine, the collective-crosstalk failure mode (a caught rank-asymmetric raise pairs the survivors' pending collective with the next one the failed ranks enter — observed producing garbage with no error), and the unguarded empty-source rbf/kNN paths — all recorded in **#405**, which also links #314 as the same family seen from the DMInterpolation side.

### Response commit
ad63bc95; the level_1+tier_a gate (443 passed) and the np2/np4 batteries were re-run after it.

ss2098:
Professor @lmoresi, this adversarial review is very helpful. Since the remaining failures are in the empty-rank / collective-evaluation family tracked in #405, I will not try to modify the core evaluation path directly.

One place where I may be able to help is with small validation/regression work once the fix direction is clear: running the starved-rank reproducers at np2/np4, checking macOS behaviour, and helping add or update focused tests around the empty-rank cases without touching the solver internals.

I'll keep my contributions limited to small, measurable test/API/documentation fixes unless there is a specific low-risk piece you would like me to try.

lmoresi:
Don’t worry about the naming. “Adversarial” just refers to the fact that we try to play devil’s advocate when reviewing to look for holes in the logic and clashes with other fixes. I used to run these for myself, but now I publish them back to the PR so we have a record of the back and forth.

Thanks for your help with bug fixes, please keep going !

L

Issue #411, opened 2026-07-24
compute_volume_residual_fields / boundary sibling still strip Dirichlet values (the #407 fix was applied to one of three duplicated variants)


lmoresi:
Follow-up from the #407 fix review (2026-07-24): `_assemble_volume_reaction` got the `DMPlexInsertBoundaryValues` insert, but the duplicated round-trip in `compute_volume_residual_fields` and its boundary sibling (petsc_generic_snes_solvers.pyx ~7557) did not — their docstrings advertise the same CBF use-case and they still return garbage on g != 0 walls. No in-repo callers today.

This trio is already flagged as duplication in the readability review (READ-27): the right fix is probably to de-duplicate onto one insert-correct core rather than patch three copies.

Underworld development team with AI support from Claude Code

lmoresi:
Fixed by PR #456, merged to development on 2026-07-28. (Manual close — PRs target development, so auto-close doesn't fire.)

Issue #410, opened 2026-07-24
Time-dependent Dirichlet BCs are broken in solve(): essential values inserted at PETSC_MIN_REAL, mesh.t kernels see -1.8e308


lmoresi:
Found by the adversarial review of the #407 fix (2026-07-24) — pre-existing, upstream of that change.

A Dirichlet BC that references `mesh.t` NaNs the solve:

```python
poisson.add_dirichlet_bc(mesh.t, "Bottom")
poisson.solve(time=2.0)   # DIVERGED_FNORM_NAN; wall values -inf
```

Mechanism (read in PETSc source): the solve-side boundary insertion goes through `DMPlexSNESComputeBoundaryFEM`, which hard-codes `time=PETSC_MIN_REAL` into `DMPlexInsertBoundaryValues`; that value lands as `petsc_t` (-1.8e308) in the JIT'd essential-BC kernel, so any BC expression containing `mesh.t` evaluates to garbage. `solve()`'s docstring claim that 'expressions using mesh.t evaluate at this time' is empirically false for Dirichlet BCs.

Note the working counter-evidence: `_assemble_volume_reaction(time=2.0)` (post-#407) inserts g(t) correctly via a direct `DMPlexInsertBoundaryValues` call with the real time — so the kernels and the insertion machinery are fine; only the solve path's hard-coded PETSC_MIN_REAL is the problem. Fix direction: insert essential values explicitly with the solver's non-dimensional time before SNESSolve (or set the DM time in a way the PETSc path picks up — `UW_DMSetTime` exists but the BoundaryFEM path ignores it).

Related latent detail: `boundary_flux()` never forwards a solve time to `_assemble_volume_reaction`, so once this is fixed, CBF on a time-dependent wall would read g(0) — forward the time in the same change.

Underworld development team with AI support from Claude Code

lmoresi:
## Design ruling (Louis, 2026-07-24): retire the PETSc-time plumbing; time becomes a model-owned rampable constant

This failure is not news — the PETSc time-value route (`UW_DMSetTime` → `petsc_t` in kernels) was tried before and the trick failed; the PETSC_MIN_REAL insertion documented above is just where it visibly dies for essential BCs. The agreed direction instead:

[...]

PR #412, opened 2026-07-24
CBF flux on driven boundaries, kdtree module identity, global velocity_error

### LINKED ISSUE #316: Level-1 fast tests: memprobe failures appear order-dependent
## Summary

`tests/test_0780_memprobe.py` fails during the default level-1 fast test run, but the same file passes when rerun directly. This looks like order-dependent global-state/test pollution in the fast-mode test path rather than a deterministic memprobe failure.

## Environment

- Repo: `underworldcode/underworld3`
- Branch: `development`
- Commit: `6451896a` (`Rotated free-slip: Schur preconditioning parity + 3D rotation nullspace (#306)`)
- Local env: `amr-dev`
- Platform observed: macOS arm64

## Reproduction

After syncing `development` and rebuilding UW3 in `amr-dev`:

```bash
./uw build
./uw test
```

The build passed, then the default level-1 fast-mode test run failed with the memprobe failures below.

## Observed full level-1 failures

```text
FAILED tests/test_0780_memprobe.py::test_kdtree_live_count_tracks_construction_destruction - AttributeError: module 'underworld3.kdtree' has no attribute 'live_count'
FAILED tests/test_0780_memprobe.py::test_diff_reports_kdtree_growth - KeyError: 'kdtree'
FAILED tests/test_0780_memprobe.py::test_probe_context_emits_when_enabled - AssertionError: assert 'kdtree' in '[memprobe] test-block: no change\n'
FAILED tests/test_0780_memprobe.py::test_mesh_variable_kdtree_caching - AttributeError: module 'underworld3.kdtree' has no attribute 'total_constructed'
FAILED tests/test_0780_memprobe.py::test_swarm_kdtree_caching - AttributeError: module 'underworld3.kdtree' has no attribute 'total_constructed'
```

The full run ended with:

```text
6 failed, 660 passed, 14 skipped, 782 deselected, 7 xfailed, 1 xpassed
```

One other failure in that run was unrelated/reproducible in AMR adapt:

```text
FAILED tests/test_0810_amr_swarm_migration_regression.py::test_swarm_migration_after_adapt_does_not_raise - IndexError: index 91 is out of bounds for axis 0 with size 80
```

## Isolation check

Running the memprobe file directly together with the reproducible AMR failure file showed that the memprobe tests pass in isolation:

```bash
pytest --config-file=tests/pytest.ini --timeout=120 \
  tests/test_0780_memprobe.py \
  tests/test_0810_amr_swarm_migration_regression.py -q
```

Result:

```text
...........F
1 failed, 11 passed
```

Only `test_0810_amr_swarm_migration_regression.py` failed in this focused run. The memprobe tests passed.

## Notes

A fresh import has the expected counters available:

```python
import underworld3 as uw
hasattr(uw.kdtree, "live_count")          # True
hasattr(uw.kdtree, "total_constructed")   # True
```

The failure pattern suggests that an earlier test in the level-1 fast run may mutate or replace the module object exposed as `uw.kdtree`, leaving `tests/test_0780_memprobe.py` to see `underworld3.kdtree` without the `ckdtree` counter functions. One suspect is code/tests that import the compatibility wrapper module `underworld3.kdtree`, but this needs confirmation.

## Expected behavior

`tests/test_0780_memprobe.py` should pass both when run directly and when included in the default level-1 fast test run.

### LINKED ISSUE #370: np4: test_parallel_custom_fmg_stokes_constrained rel-L2 0.00562 vs 0.005 tolerance (np2 passes)
Found (pre-existing) during the #365 landing gates (PR #369, 2026-07-16): `tests/parallel/test_1017_custom_mg_parallel_mpi.py::test_parallel_custom_fmg_stokes_constrained` fails at np4 with rel-L2 error 0.00562 against the 0.005 tolerance — **identical on unmerged development**, so it is not caused by the landing. Passes at np2.

Either the tolerance is marginally too tight for the np4 partition (a 12% overshoot smells like partition-dependent iteration count rather than a wrong answer), or there is a genuine small np4-specific accuracy loss in the custom-FMG constrained path. Related territory: #291 / FO-06 (Stokes_Constrained 0.4% 3D velocity, release-blocking list).

### LINKED ISSUE #407: CBF boundary_flux: _assemble_volume_reaction is wrong on inhomogeneous Dirichlet boundaries (2D and 3D)
Found by the adversarial review of PR #404 (2026-07-24) — pre-existing, dimension-independent, NOT introduced by that PR.

`_assemble_volume_reaction` (the residual-based reaction used by `boundary_flux`) is exact on g=0 Dirichlet boundaries but produces garbage when the boundary value is non-zero:

- P1 tet box, unit conduction: Top (g=0) boundary reaction sums to −1.000000 (exact); Bottom (g=1) sums to **−4.41** (P1) / **−11.59** (P2) against a true flux of ±1.
- 2D box, Bottom g=1: sumR = −27.27, recovered 'flux' ≈ −29.5 for a true value of 1.
- Shell case: interior residual is O(1) even though the KSP converged to rnorm 1e-13 — the Dirichlet boundary VALUES are evidently not inserted into the residual evaluation state before `DMPlexSNESComputeResidualFEM`.

Existing regressions (test_1019 etc.) pass only because they measure homogeneous boundaries. Every published CBF heat-flux / Nusselt recipe that reads the flux on a non-zero-temperature boundary is affected.

Likely fix: scatter the constrained boundary values into the local state vector (boundary-FEM-correct scatter, the same class as the SNES-callback scatter fix in #? — see `test_1016`'s driven-boundary machinery) before computing the residual-based reaction.

lmoresi:
Three fixes from the small-issues sweep, with the public adversarial-review round. Closes #407, closes #316, closes #370.

## The fixes

**#407 — CBF boundary flux was garbage on inhomogeneous Dirichlet walls (2D and 3D).** Constrained DOFs don't exist in the global vector, so `_assemble_volume_reaction`'s gather round-trip zero-filled every boundary value before integrating the reaction residual — a g=1 wall read −17.7 for a true unit flux, and every existing test was blind because it measured g=0 walls. `DMPlexInsertBoundaryValues` now fills the essential values first. Regression tests: exact unit flux through a driven wall (end nodes included), and a corner-semantics test (T=x+y, all four walls driven) that pins the method's inherent corner behaviour — a corner reaction integrates BOTH adjacent walls' fluxes. The review's Couette probe confirms the Stokes case: inhomogeneous velocity walls now give exact (±1, 0) tractions where the old code garbled them; all SolCx/rotated-BC goldens are unchanged (g=0 inserts zeros over zeros), and the rotated-reaction path is untouched by construction.

**#316 — order-dependent memprobe failures.** A shim `kdtree.py` shadowed the package attribute on the first `import underworld3.kdtree` (which `test_0000_imports` performs), silently dropping the memprobe counters for the rest of the run. Now one module identity via a `sys.modules` alias; shim deleted; identity regression test. Verified in both import orders.

**#370 — the np4 FMG 'tolerance miss' was never a solver problem.** `velocity_error` computed rank-local norms: the same converged solution measured 0.00027 on one rank and 0.00562 on the rank owning the SolCx viscosity jump, and the test asserted per-rank. Now a global reduction (identical on every rank; the seam-DOF weighting approximation is documented). Global np4 error: 0.0006 against the unchanged 5e-3 tolerance; the full test_1017 file passes at np2 and np4. Second instance of the rank-local-diagnostic disease after #351 — worth watching for as a class.

## Adversarial review (public, round 1 — posted below)

All three held. The review's probes: Couette all-wall Stokes exactness, mixed natural+Dirichlet bit-identity, golden batteries (test_1018 serial 14 passed, test_1064 np2, test_1065 np2/np4, SolCx goldens), corner-test flake repeats (serial/np2x3/np4x3), both kdtree import orders, pickle/module-name attacks, and an exhaustive collective-safety sweep of every `velocity_error` call site (the global norm makes it collective — no rank-guarded caller exists). Serial equivalence of the new norm: 1 ulp.

Found on the way, filed separately: **#410** — time-dependent Dirichlet BCs NaN out in `solve()` (upstream `PETSC_MIN_REAL` insertion reaches `mesh.t` kernels as −1.8e308; pre-existing, and the reason `boundary_flux`'s time handling is currently moot); **#411** — the duplicated `compute_*_residual_fields` diagnostics still carry the stripped-Dirichlet round trip (READ-27 de-duplication is the right fix).

## Verification

- test_1019 (with new tests): serial, np2, np4 — green, repeats clean.
- test_0000+test_0780+test_0101 in both orders: green.
- test_1017 full file: np2 and np4 green.
- Level_1 + tier_a gate: **444 passed**, 0 failures.

Manual close needed on merge: #407, #316, #370.


lmoresi:

Independent skeptic pass over the branch diff; live probes serial/np2/np4 in the branch's own build.

### #407 — essential-value insert before the reaction residual
- **Stokes / mixed-BC corruption attack: HELD.** All-wall inhomogeneous Couette (u=(y,0), P2-P1, pressure nullspace): solution exact to 2e-12, recovered tractions exactly (±1, 0) — precisely the case the old code garbled. Adding a natural wall alongside: bit-identical results. Rotated-BC path untouched by construction (reads the stashed solve reaction).
- **Golden regression attack: HELD.** test_1018 serial (14 passed, incl. SolCx sigma_nn + dynamic-topography goldens), test_1064 np2, test_1065 np2/np4, SolCx goldens — nothing shifted (for g=0 the insert writes zeros over zeros).
- **Time-dependent BC attack: could not be realized — and found a pre-existing solver bug instead.** No caller passes time= today, and no solve can reach boundary_flux with a live mesh.t Dirichlet wall because solve() itself inserts essential values at PETSC_MIN_REAL — mesh.t kernels see −1.8e308 and the solve NaNs. Filed as #410 (with the note that boundary_flux must forward the solve time once that is fixed). The time= plumb in this diff was verified to insert g(t) correctly.
- **Scope gap found**: the duplicated compute_volume/boundary_residual_fields diagnostics did NOT get the insert (no in-repo callers; READ-27 de-duplication territory). Filed as #411.
- Corner-semantics test: no flake across serial + np2 x3 + np4 x3.

### #316 — kdtree module identity
**HELD.** All import spellings resolve to the one module, both import orders green, pickle/module-name attacks dead on arrival (KDTree was never picklable; __module__ was always ckdtree). One packaging note, fixed before push: the stale setuptools build cache re-shipped the deleted shim into the local wheel — build/lib.* cleaned; clean checkouts/CI unaffected, and the sys.modules alias preempts a stale file anyway (verified: the identity test passed even with the stale shim present).

### #370 — global velocity_error
**HELD — the collective-deadlock attack found no victim.** The global norm makes the function collective; an exhaustive sweep of every call site (10 test files, src, docs, scripts) found none inside rank guards or rank subsets. Serial equivalence to the old formula: ~1 ulp. Seam-DOF duplication quantified at ~4% of rows at np2 — matches the docstring's disclosed approximation. Bonus: empty ranks now get the finite global value instead of 0/0 NaN.

### Verdict
Merge. Follow-ups #410 and #411 filed; neither blocks.

Issue #415, opened 2026-07-24
sympy.Max/Min(UWexpression, plain number) crashes at construction: AttributeError '_prec'


lmoresi:
Found by the #302 adversarial review (2026-07-24), pre-existing (reproduced identically on unmodified development):

```python
k = uw.expression(r'k', 1.0, 'constant')
sympy.Max(k, 0.5)   # AttributeError: 'UWexpression' object has no attribute '_prec'
```

UWexpression reports `is_comparable=True`, so sympy's Max/Min attempts an immediate numeric comparison and reaches for the Float internals (`_prec`) that a Symbol subclass doesn't have. The failure is loud (not a silent-bake hazard), but it blocks the natural way of writing clamped/floored constants — users must route through uw.maths.smooth_max or wrap differently.

Likely fix: UWexpression should not advertise `is_comparable` (it is symbolic until packed), or MathematicalMixin should intercept Max/Min construction. See `src/underworld3/utilities/mathematical_mixin.py`.

Underworld development team with AI support from Claude Code

lmoresi:
Fixed by PR #456, merged to development on 2026-07-28. (Manual close — PRs target development, so auto-close doesn't fire.)

PR #416, opened 2026-07-24
JIT constants nested in wrapper expressions are live-rampable — reveal before substitute

### LINKED ISSUE #302: JIT: a manifested constants[] UWexpression can be baked into the C source (not live-rampable)
## Summary

A constant `UWexpression` that **is** collected into `solver.constants_manifest` can nonetheless have its value **baked into the compiled C source**, so changing its `.sym` between solves has **no effect** — the parameter is not live-rampable even though it is listed as a live constant. A full solver rebuild (`is_setup = False` + re-set the constitutive parameter) is required to pick up the new value.

This breaks the documented contract that a manifested constant can be updated via `.sym` and repacked through `constants[]` without recompilation. It is the mechanism the δ-continuation / homotopy work relies on, and it works for some parameters (δ) but silently fails for others (a rate-strengthening viscosity floor ξ).

## Observed behaviour (Stokes, `ViscousFlowModel`, Drucker–Prager + Perzyna floor)

Viscosity contains two rampable constants defined identically as `uw.expression(...)`:

- **δ** (soft-min smoothing): appears as `δ²` inside `sqrt((f-1)² + δ²)` where `f` depends on the strain-rate field. **Live-ramps correctly** — setting `δ.sym` and solving picks up the new value (the δ-march converges as expected).
- **ξ** (Perzyna floor): appears as `ξ·η_bg` in `η_p = σ_y/(2·ε̇) + ξ·η_bg`. **Does NOT live-ramp** — setting `ξ.sym` between solves leaves the solution unchanged; only a rebuild propagates the new ξ.

Both appear in `constants_manifest` (verified by printing `[e.name for i,e in solver.constants_manifest]` — `\xi` is present at a valid index). So ξ is *listed* as a live constant but its value is *baked* in the generated C — updating `constants[]` for its slot does not change the residual.

Verification was via **full solves** (a warm ξ-descent): without a rebuild between ξ steps the converged solution (and a band-thickness diagnostic) is frozen across all ξ; with a rebuild per ξ it varies correctly. (Note: a quick `snes_max_it=0` + `getFunctionNorm()` residual probe is **not** a reliable test here — it returns a lagged norm — so please validate any fix with a real solve.)

## Likely mechanism (hypothesis for the maintainers)

`constants_manifest` is built by `_extract_constants` / `_collect_constant_atoms` on the **pre-unwrap** callbacks (so ξ is found and listed). The C source is generated from the **post-unwrap** expression. The nondimensionalisation/unwrap appears to **fold pure-constant subexpressions to numbers**: `ξ·η_bg` is a product of two constants and collapses to a literal, so ξ never reaches the substitution that would route it through `constants[]`. δ survives only because its sole occurrence is inside a **field-dependent** subexpression (`(f-1)² + δ²`) that cannot fold. Net: a manifested constant whose every occurrence is inside an otherwise fully-constant subexpression gets baked, and the manifest/​C-source disagree.

If correct, reformulating ξ so it always multiplies a field (e.g. the natural Perzyna numerator `ξ·η_bg·2·ε̇` before dividing by `2·ε̇`) *might* keep it live — but this is **untested** (the fold may still occur after sympy simplification), and the deeper fix is to make `_extract_constants` and the C-source generation agree (either don't fold a subexpression that contains a manifested constant, or substitute manifested constants before the fold).

## Impact

- Silent: no error; the parameter simply doesn't update, which can invalidate continuation/homotopy studies that ramp such a parameter in place.
- Same *family* as the previously-noted "constant in an exponent doesn't ramp" JIT gotcha.

## Suggested checks

1. Make `constants_manifest` and the emitted C source consistent — a listed constant must be read from its `constants[]` slot, never baked.
2. Add a guard/warning when a manifested constant is detected to have been folded out of the C source (manifest lists it but the code never references its slot).
3. A minimal repro: two `uw.expression` constants in a viscosity, one multiplying a field, one multiplying only other constants; ramp each `.sym` and confirm via a real solve which propagates.

Reported from the yield-homotopy investigation (Spiegelman notch, harness `convergence.py`); happy to provide the full harness or help build a minimal repro.

lmoresi:
Closes #302. The prerequisite for the model-owned-time design recorded on #410.

## The bug — broader than reported

A constant `UWexpression` listed in `constants_manifest` could still be baked into the generated C as a literal. The lowering substituted `constants[]` placeholders with a TOP-LEVEL `xreplace` and only then unwrapped — so a constant hidden inside a nested `UWexpression` never met the substitution, and the unwrap folded it to a number. Since **every `Parameters.*` value is template-wrapped in a `UWexpression`**, this froze every constant inside constitutive parameter expressions, not just the issue's ξ case: the reproducer shows even the 'δ pattern' freezes when it arrives through a parameter (ramping changed nothing; a rebuild changed the answer x10). It also explains why rebuilds 'worked': the cache key baked the same values, so a ramp changed the hash and forced the recompile a live constant exists to avoid.

## The fix

Three-phase lowering, identical in codegen and cache hashing (shared `_reveal_constants`): (1) expand non-constant `UWexpression`s to fixpoint while KEEPING truly-constant atoms symbolic — the predicate is the same one the manifest scan uses, so the revealed set equals the manifested set by construction; (2) `xreplace` manifested constants → `constants[]` placeholders; (3) resolve remaining atoms. Two RuntimeError guards make silent baking structurally impossible: a truly-constant atom with no manifest slot, or a manifested constant surviving to the final expression, refuses loudly. Nested-constant ramps no longer invalidate the JIT cache.

## Adversarial review (public, posted below) — all six attacks HELD

Eight nested topologies probed with real solves against independent literal-rebuild ground truths — constant exponents (the known sibling gotcha), `sqrt(f² + δ²)`, `Piecewise`, `Max` including a ramp that **switches the runtime branch**, double nesting: every ramp matched truth to 1e-9, zero silent bakes. Guards: zero false positives across rotated free-slip consistent-Newton, ND poisson/stokes, units-active constants, and a VEP subset (31 tests green). Cache: ramp → 0 new .so, structural change → recompiles. Consistent-Jacobian: ramped nested amp AND exponent give bit-identical solution and iteration count vs a rebuild. Performance: 20-deep nesting, no measurable cost vs development. The review's one in-diff finding (a silent fallback in the cache-key path that would have re-created recompile-per-ramp) is fixed in the response commit; its pre-existing discovery is filed as #415 (`Max/Min(UWexpression, number)` construction crash).

## Verification

- New regression tests (`tests/test_0103_jit_rampable_constants.py`): top-level and nested constants ramp via real solves with no rebuild; ramped == rebuilt to 1e-10; ramp-back restores base exactly.
- Level_1 + tier_a gate: 445 passed pre-rebase; post-rebase (onto #404) re-run reported in the review thread before merge.

Manual close on merge: #302.

### DIFF
--- a/src/underworld3/utilities/_jitextension.py
+++ b/src/underworld3/utilities/_jitextension.py
@@ -299,11 +299,11 @@ def prepare_for_cache_key(fn, constants_subs_map):
     3. Unwrap the remaining UW atoms to pure SymPy so the hash is
        deterministic.
     """
-    # Phase 1: reveal constants nested inside other UWexpressions
-    try:
-        fn_structural = _reveal_constants(fn)
-    except Exception:
-        fn_structural = fn
+    # Phase 1: reveal constants nested inside other UWexpressions. Loud on
+    # failure, exactly like the codegen path — a silent fallback here would
+    # hash constant VALUES into the cache key and force a recompile on
+    # every ramp (adversarial-review finding).
+    fn_structural = _reveal_constants(fn)
 
     # Phase 2: Substitute constants with _JITConstant placeholders
     if constants_subs_map and fn_structural is not None:
--- a/src/underworld3/utilities/_jitextension.py
+++ b/src/underworld3/utilities/_jitextension.py
@@ -260,32 +260,60 @@ class JITCallbackSet:
                 len(self.bd_residual), len(self.bd_jacobian))
 
 
+def _reveal_constants(fn):
+    """Expand non-constant UWexpressions to fixpoint, KEEPING truly-constant
+    atoms symbolic — so constants nested at ANY depth surface as atoms.
+
+    This must run BEFORE the ``constants[]`` substitution: a top-level
+    ``xreplace`` cannot see a constant hidden inside a nested UWexpression
+    (every ``Parameters.*`` value is template-wrapped in one), so nested
+    constants were silently folded to C literals while the manifest still
+    listed them as live — issue #302. The keep-constants predicate here is
+    the same ``_is_truly_constant`` used to build the manifest, so the set
+    of atoms revealed is exactly the set the manifest routes to
+    ``constants[]``.
+    """
+    from underworld3.function.expressions import (
+        unwrap_expression,
+        UWDerivativeExpression,
+    )
+
+    if fn is None:
+        return fn
+    if isinstance(fn, UWDerivativeExpression):
+        fn = fn.doit()
+    if isinstance(fn, sympy.MatrixBase):
+        return fn.applyfunc(
+            lambda e: unwrap_expression(e, mode='symbolic_keep_constants'))
+    return unwrap_expression(fn, mode='symbolic_keep_constants')
+
+
 def prepare_for_cache_key(fn, constants_subs_map):
     """Prepare a single expression for JIT cache hashing.
 
-    Two-phase process:
-    1. Substitute constant UWexpressions with ``_JITConstant`` placeholders
+    Three-phase process (mirrors the codegen lowering in ``_createext`` so
+    the cache key and the generated C agree — issue #302):
+    1. Reveal nested constants (``_reveal_constants``).
+    2. Substitute manifested constants with ``_JITConstant`` placeholders
        so that changing a constant's *value* does not invalidate the cache.
-    2. Unwrap remaining (non-constant) UWexpressions to pure SymPy so the
-       hash is deterministic.
-
-    Parameters
-    ----------
-    fn : sympy expression or None
-        The expression to expand.
-    constants_subs_map : dict or None
-        Mapping from UWexpression symbols to ``_JITConstant`` placeholders.
+    3. Unwrap the remaining UW atoms to pure SymPy so the hash is
+       deterministic.
     """
-    # Phase 1: Substitute constants with _JITConstant placeholders
-    if constants_subs_map and fn is not None:
-        try:
-            fn_structural = fn.xreplace(constants_subs_map) if hasattr(fn, "xreplace") else fn
-        except Exception:
-            fn_structural = fn
-    else:
+    # Phase 1: reveal constants nested inside other UWexpressions
+    try:
+        fn_structural = _reveal_constants(fn)
+    except Exception:
         fn_structural = fn
 
-    # Phase 2: Unwrap remaining (non-constant) expressions
+    # Phase 2: Substitute constants with _JITConstant placeholders
+    if constants_subs_map and fn_structural is not None:
+        try:
+            if hasattr(fn_structural, "xreplace"):
+                fn_structural = fn_structural.xreplace(constants_subs_map)
+        except Exception:
+            pass
+
+    # Phase 3: Unwrap remaining (non-constant) expressions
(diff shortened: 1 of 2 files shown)


lmoresi:

Independent skeptic pass, live probes in the branch's own build. Instructed to REFUTE; could not.

### Attack 1 — sympy folding during the reveal phase: HELD
Eight nested-constant topologies, each ramped via .sym with REAL solves and checked against an independent literal-inlined rebuild (different kernel, separate compile): constant exponent over a field base; constant exponent with numeric base inside a wrapper; Integer*constant Mul auto-simplification; sqrt((T−1)²+δ²) regularisation; Piecewise on a field threshold; DOUBLE nesting; Max(constant, field); and Max(k1,k2) where the ramp switches the runtime branch (kernel evaluates fmax(constants[i],constants[j]) live). Every ramped answer matched truth to 1e−9. No silent bakes; the guards never needed to fire.

### Attack 2 — guard false positives: HELD
test_0103 + test_1018 (rotated free-slip, consistent-Newton power-law) + test_0817/0818 (ND) + test_0812 (units-active constants): 29 passed, zero guard errors. VEP stability subset: 2 passed. No spurious RuntimeError anywhere.

### Attack 3 — cache behavior: HELD
First solve: exactly one new .so. Ramp nested constitutive constant: zero new .so, identical cache key, correct answer. Ramp a BC constant: same. Structural change: recompiles as it must. A fully-constant wrapper chain collapses onto the template constant's slot and inner ramps still propagate (values are re-packed per solve).

### Attack 4 — consistent Jacobian: HELD
Power-law config with consistent_jacobian=True: ramping BOTH a nested amplitude and a nested EXPONENT (the historically-broken family), no rebuild — solution and Newton iteration count bit-identical to a fresh rebuild at the same values (vmean 0.00065017399 both, its=6 both).

### Attack 5 — performance: HELD
20-deep nested chain (20 manifested constants): setup 0.10s / first solve 4.19s vs development's 0.26s / 4.24s. No regression.

### Attack 6 — BC and boundary kernels: HELD
Constants in Dirichlet expressions and natural-BC fluxes are manifested, ramp live without recompile, and match truth (natural-flux ramp: exactly 3x on a linear problem). bd_residual/bd_jacobian slots get the same reveal.

### Findings → responses
- In-diff: prepare_for_cache_key silently swallowed reveal failures (would hash constant VALUES → recompile-per-ramp for that expression while codegen raised). Fixed in the response commit — both paths now fail loudly and identically.
- Pre-existing, filed as #415: sympy.Max/Min(UWexpression, plain number) crashes at construction (AttributeError '_prec' — is_comparable over-promise). Loud, so not a #302-family hazard, but blocks the natural spelling of clamped constants.

### Verdict: MERGE.

lmoresi:
Post-rebase verification (onto development at #404): level_1+tier_a gate **449 passed, 0 failures** on a clean rebuild — including #404's four new boundary-flux/topography tests running against this branch's lowering. (An intermediate gate run showed 4 failures that traced to a half-built environment from a timeout-killed build — old compiled extension without the #412 insert paired with post-rebase sources; `rm -rf build/` + rebuild resolved it, and the same four tests pass identically to clean development. Recorded here for honesty of the record.)

Issue #417, opened 2026-07-25
Repeated rotated free-slip Stokes solves rebuild solver state and grow RSS until OOM


gthyagi:
## Summary

Repeated **linear Stokes solves with rotated strong free-slip boundary conditions**
grow resident memory substantially on every timestep. In an 8-rank Zhong et al.
(2008) A1 thermal-convection run on a 16 GiB Apple Silicon Mac, aggregate Python
RSS reached 54-59 GiB before macOS watchdog panics/restarts.

This is distinct from the constrained-multiplier scaling problem in #244. Single
rotated solves are accurate and fast; the failure appears during repeated
transient solves because the rotated path reconstructs PETSc matrices and the
fieldsplit/GAMG solver hierarchy for every call.

## Reproducer

Benchmark script:

- https://github.com/gthyagi/uw3-mantle-convection-benchmarks/blob/main/benchmarks/bench_100_zhong2008_thermal_convection_A1.py

Detailed postmortem:

- https://github.com/gthyagi/uw3-mantle-convection-benchmarks/blob/main/docs/thermal_A1_memory_postmortem.md

The unsafe production case used `cellsize=1/8`, P2/P1, 8 MPI ranks, and rotated
free slip. The mesh had only about 48,202 Stokes unknowns, so solution-vector
storage does not explain the observed memory.

## Crash evidence

| Run | Aggregate Python RSS before failure | Per-rank RSS | System state |
|---|---:|---:|---|
| 1 | 54.29 GiB | 6.17-7.28 GiB | watchdog panic |
| 2 | 59.14 GiB | 7.05-7.60 GiB | watchdog panic |
| 3 | 59.18 GiB | 7.12-7.61 GiB | watchdog panic |

At failure, the compressed-segment pool was effectively full and macOS had
created 41-42 swapfiles.

## Bounded coarse probes

Serial `cellsize=1/4` repeated-solve probes isolate the boundary-condition path:

| BC / experiment | RSS increase | Approx. increase per subsequent solve |
|---|---:|---:|
| Rotated free slip | +102.03 MiB over initialization and 12 solves | 8.54 MiB/solve |
| Rotated, explicitly destroy prior `_rotated_freeslip_info` | +88.65 MiB over 11 subsequent solves | 8.06 MiB/solve |
| Native Nitsche | +25.57 MiB over 9 subsequent solves | 2.84 MiB/solve and near plateau |

Destroying the Python-visible prior result does not remove the slope. This points
[...]

PR #418, opened 2026-07-25
Reuse linear rotated free-slip solver workspace

gthyagi:
## Summary

Fixes the unbounded resident-memory growth observed when a linear rotated free-slip Stokes system is solved repeatedly, as in transient mantle-convection models.

The change caches the rotated linear solver workspace and reuses it while the operator is unchanged. It also adds explicit invalidation for coefficient, time, boundary-condition, constitutive-model, solver, and DM changes.

Tracks #417.

## Root cause

Repeated solves rebuilt two expensive classes of PETSc state:

1. `_residual_is_nonlinear()` assembled and compared two Jacobians on every solve.
2. The rotated solve path rebuilt the rotation operators, transformed matrix, pressure mass, nullspaces, field splits, KSP, and GAMG hierarchy.

Stage-isolation measurements showed approximately 76.77 MiB growth across 12 repeated nonlinearity probes and 117.23 MiB across 12 direct rotated solves. Explicitly destroying only the previous result vectors did not address the retained PETSc solver state.

## Implementation

- Cache the rotated residual linearity classification.
- Keep one owned rotated linear workspace containing `Q`, `Q^T`, transformed operators, pressure mass, nullspaces, KSP/PC, coefficient state, and deterministic operator probes.
- Skip Jacobian/preconditioner assembly for RHS-only changes when coefficient `MeshVariable._state` values are unchanged.
- Refresh the operator when coefficient fields change.
- Force refresh/comparison when an explicit physical `time=` is supplied, covering `mesh.t`-dependent symbolic operators.
- Destroy the workspace when solver, DM, BC, or constitutive state is rebuilt.
- Store the Cartesian reaction directly for topography instead of retaining matrix/RHS copies solely for later reconstruction.
- Expose and forward `time=` separately from `timestep=` in `Stokes.solve()`.
- Preserve the direct-LU path.

The design and invalidation contract are documented in `docs/developer/design/ROTATED_FREESLIP_LINEAR_REUSE.md`.

## Validation

The isolated fix was cherry-picked onto the current `development` tip. The branch is exactly one commit ahead with no merge conflicts; `git diff --check` and Python syntax checks pass.

Focused serial and MPI validation covered repeated RHS solves, viscosity invalidation, explicit-time refresh, spherical velocity and leakage, Zhong 3-D reaction topography, boundary traction, dynamic topography, and partition independence.

| Case | Repeated solves | RSS change after fix | Correctness |
|---|---:|---:|---|
| Serial spherical, cell size 1/4 | 30 | peak +1.91 MiB | RHS scaling error < 1e-9 |
| 8-rank coarse box | 30 | +11.84 MiB aggregate, +1.48 MiB/rank | scaling error < 8e-14 |
| 8-rank spherical, cell size 1/4 | 10 | +12.78 MiB aggregate, +1.60 MiB/rank | scaling error < 3e-10 |
| Serial viscosity x2 invalidation | 1 rebuild | old KSP destroyed; cache not reused | expected half-velocity error 8.10e-11 |
| Serial forced rebuild | 1 rebuild | old KSP destroyed; cache not reused | scaling error 1.60e-10 |

Before this change, the serial spherical probe grew by 102.03 MiB over 12 solves (about 8.54 MiB/solve). The production 8-rank Zhong A1 run had reached 54.29-59.18 GiB aggregate RSS on a 16 GiB machine and triggered macOS watchdog restarts.

MPI checks used the Pixi environment's Open MPI launcher. The Homebrew Open MPI launcher is ABI/version-mismatched with the environment and can hang during startup; no `FI_PROVIDER=tcp` workaround was used.

## Scope

This PR changes only the UW3 rotated free-slip solve lifecycle, focused tests, and developer documentation. Benchmark-repository scripts and result documents are not included.

### DIFF
--- a/docs/developer/design/ROTATED_FREESLIP_LINEAR_REUSE.md
+++ b/docs/developer/design/ROTATED_FREESLIP_LINEAR_REUSE.md
@@ -1,0 +1,108 @@
+# Rotated Free-Slip Linear Solver Reuse
+
+## Scope
+
+This note documents the repeated-solve workspace introduced for linear Stokes
+problems using rotated strong free slip. It addresses
+[issue #417](https://github.com/underworldcode/underworld3/issues/417).
+
+The change does not alter the rotated boundary condition or its discrete
+equations. It changes PETSc object ownership and determines when the assembled
+operator and its preconditioner can be reused.
+
+## Previous Lifecycle
+
+Every linear rotated solve previously rebuilt:
+
+1. the nodal rotation matrices `Q` and `Q.T`;
+2. the transformed saddle operator `Ahat = Q A Q.T`;
+3. the pressure mass block;
+4. the coupled and pressure nullspaces;
+5. the fieldsplit Schur KSP/PC;
+6. the velocity GAMG hierarchy.
+
+The linear/nonlinear detector also assembled and copied two trial Jacobians on
+every call. Long transient runs therefore accumulated a large PETSc and system
+allocator high-water mark even when only the body-force field changed.
+
+## Reused Workspace
+
+The Stokes solver now owns one `_rotated_linear_cache` containing:
+
+- `Q`, `Q.T`, and constrained normal rows;
+- the transformed matrix `Ahat`;
+- two deterministic operator-probe vector triples;
+- operator coefficient mesh variables and their UW data-version counters;
+- the pressure mass block, nullspaces, fieldsplit KSP/PC, and GAMG hierarchy;
+- the constraint-row diagonal scale.
+
+For an unchanged linear operator, a subsequent solve:
+
+1. assembles only `F(0)` and forms the new right-hand side;
+2. rotates and constrains that right-hand side;
+3. solves with the existing KSP/PC hierarchy;
+4. rotates the solution back;
+5. forms and retains the Cartesian reaction needed by topography recovery.
+
+The prior per-solve solution and reaction vectors are explicitly destroyed
+before replacement.
+
+## Invalidation
+
+The cache is destroyed before a solver or DM rebuild. The cached
+linear/nonlinear classification is invalidated at the same time.
+
+Mutable operator coefficients are collected from the constitutive parameters,
+constraint, penalty, and saddle preconditioner. Their base
+`MeshVariable._state` counters decide whether Jacobian assembly is required.
+
+- A body-force field change does not refresh the operator.
+- A viscosity field change refreshes `J`, `Jp`, and the rotated operator.
+- An explicit `solve(time=...)` refreshes and compares the operator, covering
+  expressions that depend on `mesh.t`.
+- Mesh, field-layout, boundary-condition, constitutive, or forced setup changes
+  destroy the full workspace.
+
+After a coefficient-triggered assembly, two deterministic matrix-vector
+products decide whether matrix values actually changed. This avoids exact
+`Mat.equal`, which rejects harmless distributed-assembly roundoff, and avoids
+copying CSR arrays, which recreated the memory high-water problem.
+
+The high-level `uw.systems.Stokes.solve()` wrapper exposes and forwards
+`time=` for this purpose. It remains distinct from the viscoelastic integration
+`timestep=`.
+
+## PETSc Ownership
+
+`_destroy_rotated_ksp_ctx` explicitly destroys the KSP, pressure mass matrix,
+and owned nullspaces. `_destroy_rotated_linear_cache` additionally destroys the
+operator probes, transformed operator, and rotation matrices.
+
+Linear topography no longer requires retaining a copy of the native operator
+and right-hand side. The Cartesian reaction `J U - b` is computed immediately
+after the solve and stored in the solve result.
+
+## Validation
+
+The focused regression verifies that:
+
+- changing only a temperature/body-force field preserves the `Q`, `Ahat`, and
+  KSP handles;
+- the velocity scales with the changed right-hand side;
+- changing a viscosity mesh variable destroys the old KSP and refreshes the
+  operator;
+- the refreshed velocity has the expected inverse-viscosity scaling.
+
+Additional serial and eight-rank tests cover spherical velocity/leakage,
+reaction-derived topography, boundary traction, and dynamic-topography field
+recovery.
+
+Always launch MPI tests with the worktree MPI executable:
+
+```bash
+.pixi/envs/amr-dev/bin/mpirun -np 8 \
+  .pixi/envs/amr-dev/bin/python <script.py>
+```
+
+Using a system `mpirun` from a different Open MPI installation can stall during
+initialization and is not a solver failure.
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -898,6 +898,7 @@ class SolverBaseClass(uw_object):
 
     def _reset(self):
 
+        self._reset_rotated_solver_cache()
         self.natural_bcs = []
         self.essential_bcs = []
 
@@ -915,6 +916,21 @@ class SolverBaseClass(uw_object):
         self.is_setup = False
 
         return
+
+    def _reset_rotated_solver_cache(self):
+        """Release rotated-free-slip state before solver/DM teardown."""
+        result = getattr(self, "_rotated_freeslip_info", None)
+        cache = getattr(self, "_rotated_linear_cache", None)
+        if result is not None or cache is not None:
+            from underworld3.utilities.rotated_bc import (
+                _destroy_rotated_linear_cache,
+                _destroy_rotated_solve_result,
+            )
+            _destroy_rotated_solve_result(result)
+            _destroy_rotated_linear_cache(cache)
+        self._rotated_freeslip_info = None
+        self._rotated_linear_cache = None
+        self._rotated_residual_is_nonlinear = None
 
     def get_snes_diagnostics(self):
         """
@@ -1228,6 +1244,11 @@ class SolverBaseClass(uw_object):
 
         if self.is_setup:
             return
+
+        # Rotated matrices and KSP/PC objects reference the current DM/SNES
+        # layout. Any requested solver rebuild invalidates the whole workspace
+        # and the cached linear/nonlinear classification.
+        self._reset_rotated_solver_cache()
 
         # Resolve the preconditioner choice (auto/fmg/gamg) against the current
         # mesh hierarchy and push the option bundle before the per-class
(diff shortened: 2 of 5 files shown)


gthyagi:
## Bounded production A1 restart: timing and memory

I reran the actual Zhong A1 production path after this fix, rather than only the synthetic repeated-solve probe.

Configuration:

- restart from the durable step-500 checkpoint;
- advance to step 510 (10 coupled steps);
- 8 MPI ranks, `cellsize=1/8`, P2P1T1, `Ra=7000`;
- rotated strong free slip and the evolved production temperature field;
- Pixi Open MPI launcher, `UW_MEMPROBE=1`;
- external 0.25-second per-rank RSS sampling;
- hard abort at 12 GiB aggregate rank RSS or 900 seconds.

The run completed normally in 87.79 seconds and neither guard was reached.

The external RSS value is the first sample after each Stokes completion. `Stokes RSS delta` is the sum of the eight rank-local changes measured strictly inside that `stokes.solve()` call.

| Phase | Coupled step time (s) | Aggregate rank RSS (MiB) | Mean RSS/rank (MiB) | Stokes RSS delta (MiB) |
|---|---:|---:|---:|---:|
| Initial Stokes setup | - | 3467.58 | 433.45 | +1003.98 |
| Step 501 | 12.87 | 2175.98 | 272.00 | +178.07 |
| Step 502 | 6.15 | 2240.27 | 280.03 | +3.36 |
| Step 503 | 5.98 | 2289.95 | 286.24 | +0.90 |
| Step 504 | 5.20 | 2322.45 | 290.31 | +2.45 |
| Step 505 | 5.53 | 2353.30 | 294.16 | +0.68 |
| Step 506 | 5.42 | 2583.67 | 322.96 | +1.41 |
| Step 507 | 5.67 | 2822.48 | 352.81 | -26.38 |
| Step 508 | 5.39 | 2832.45 | 354.06 | +0.48 |
| Step 509 | 5.61 | 2816.72 | 352.09 | +1.34 |
| Step 510 | 5.38 | 2842.64 | 355.33 | +0.96 |

### Timing summary

| Measurement | Result |
|---|---:|
| Ten coupled steps, excluding initial Stokes setup | 63.21 s |
| Mean, steps 501-510 | 6.32 s/step |
| Post-warm-up mean, steps 502-510 | 5.59 s/step |
| Post-warm-up range | 5.20-6.15 s/step |
| PETSc `SNES_Stokes.solve`, 11 calls including setup | 40.41 s |
| Previous production rate, steps 400-500 | approximately 10.75 s/step |
| Bounded post-fix improvement | approximately 1.92x |

### Interpretation

- Initial rotated operator/fieldsplit/GAMG construction was the run's RSS peak.
- Setup temporaries were released before step 501.
- Whole-process thermal/SLCN workspace warmed through approximately step 507.
- Aggregate RSS changed by only +20.16 MiB from step 507 to step 510 (+2.52 MiB/rank over three steps).
- Stokes-local changes sum to -14.80 MiB over steps 502-510, so the repeated rotated solver workspace is not accumulating in this production run.

Step-510 diagnostics continue the checkpoint trajectory: paper time 0.242451, RMS velocity 57.8450, surface/CMB Nu 2.69543/3.20704, and mean temperature 0.317257.

This validates the fix over a bounded ten-step production continuation. It does not yet prove long-run whole-process stability; a guarded run of at least 100 steps on a larger-memory host remains appropriate.

Full benchmark report: [results_bench_100_zhong2008_thermal_convection_A1.py.md](https://github.com/gthyagi/uw3-mantle-convection-benchmarks/blob/0587cae/docs/results_bench_100_zhong2008_thermal_convection_A1.py.md)

gthyagi:
### Guarded production follow-up: 100 additional A1 steps

I extended the real Zhong A1 `cellsize=1/8` production trajectory from checkpoint 510 through checkpoint 610 on 8 ranks. To protect the 16 GiB Mac, this was run as two consecutive 50-step restart processes. Each used the Pixi MPI runtime, `UW_MEMPROBE=1`, 0.25 s rank-RSS sampling, an 8 GiB aggregate-rank hard stop, and a 1200 s timeout. Neither guard fired.

| Segment | Advancing steps | Process elapsed | Mean advancing interval | Peak aggregate RSS | Process-final RSS |
| --- | ---: | ---: | ---: | ---: | ---: |
| 510 -> 560 | 50 | 303.20 s | 5.56 s | 3456.72 MiB | 3011.14 MiB |
| 560 -> 610 | 50 | 310.50 s | 5.73 s | 3615.84 MiB | 3194.33 MiB |

Each process includes one extra restart-state Stokes solve, excluded from the advancing-step timing. In the first process, aggregate RSS reached 3220.98 MiB after step 559 and then released about 644 MiB at step 560. The second process released about 141 MiB at step 570; its external RSS slope over steps 590-609 was 0.74 MiB/step aggregate. Warm Stokes-local deltas were mostly zero or sub-MiB and included periodic net-negative releases. This is unlike the former per-step `PtAP`/fieldsplit/GAMG reconstruction growth.

The physical trajectory remained continuous. At step 610 (`t=0.272219` in paper `d^2/kappa` units):

| Vrms | Surface Nu | CMB Nu | Mean T |
| ---: | ---: | ---: | ---: |
| 57.8317 | 2.69388 | 3.20751 | 0.317261 |

This is deliberately a segmented 100-step validation, not a single uninterrupted 100-step memory profile. It supports the cache/reuse fix and validates a guarded checkpoint workflow; it does not justify an unmonitored full run on this machine. Full logs, RSS interpretation, and checkpoint metrics are documented in benchmark commit [`d797be4`](https://github.com/gthyagi/uw3-mantle-convection-benchmarks/commit/d797be4).

gthyagi:
### Extended production validation: checkpoint 810

The real Zhong A1 `cellsize=1/8`, 8-rank trajectory has now advanced another 200 steps, from checkpoint 610 through checkpoint 810, using four consecutive guarded 50-step processes. Every process used the Pixi MPI runtime, `UW_MEMPROBE=1`, 0.25 s RSS sampling, an 8 GiB aggregate-rank hard stop, and a 1200 s timeout. No guard fired.

| Segment | Process elapsed | Mean advancing interval | Peak aggregate RSS | Final aggregate RSS |
| --- | ---: | ---: | ---: | ---: |
| 610 -> 660 | 323.90 s | 5.88 s | 3126.52 MiB | 3087.09 MiB |
| 660 -> 710 | 281.43 s | 5.16 s | 3189.25 MiB | 3189.25 MiB |
| 710 -> 760 | 280.76 s | 5.16 s | 3277.72 MiB | 3277.72 MiB |
| 760 -> 810 | 286.31 s | 5.28 s | 3735.03 MiB | 3266.83 MiB |

The four processes completed in 1172.39 s (19.54 min). Advancing solve intervals averaged 5.37 s, with a 5.21 s median. The largest process peak was 3.74 GiB, less than half the configured limit. Warm Stokes-local RSS repeatedly flattened to zero/sub-MiB changes and included multiple collective releases rather than cumulative workspace growth.

Physical checkpoint results remained continuous:

| Step | Paper time | Vrms | Surface Nu | CMB Nu | Mean T |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 610 | 0.27222 | 57.8317 | 2.69388 | 3.20751 | 0.31726 |
| 660 | 0.28710 | 57.8491 | 2.69535 | 3.20769 | 0.31740 |
| 710 | 0.30196 | 57.8811 | 2.69825 | 3.20824 | 0.31767 |
| 760 | 0.31678 | 57.9300 | 2.70239 | 3.20925 | 0.31802 |
| 810 | 0.33156 | 57.9729 | 2.70582 | 3.21041 | 0.31830 |

Guarded post-fix production evidence now covers 310 evolved steps after checkpoint 500. Full details are in benchmark commit [`943a93e`](https://github.com/gthyagi/uw3-mantle-convection-benchmarks/commit/943a93e). This remains segmented validation; an unmonitored full local run is still excluded.

lmoresi:
# Adversarial review

Strong PR: the ownership discipline is careful (explicit destroys, `_owns_rotation`, nonlinear path tears the cache down), and the production validation trail — 310 guarded Zhong A1 steps with flat RSS and a continuous physical trajectory — this is deep evidence / well justified. The review concentrated on the one structural question: **when must the cache be invalidated, and can the machinery miss it?** Three suspected blind spots were probed live on the PR head (bec76bbd), serial and np2, in an isolated worktree.

## Probe results

### 1. Rampable UWexpression-constant coefficients — **FAIL, blocking (silent wrong answer)**

Viscosity expressed purely as a rampable constant (`k = uw.expression(...); Parameters.shear_viscosity_0 = k`) — the exact idiom the live-rampable-constants contract (PR #416) blesses, and what every δ-continuation / homotopy driver does between solves:

- `k.sym = 2.0` then `solve(zero_init_guess=False)` → solution **bit-identical to the k=1 solution** (`‖v2−v1‖/‖v1‖ = 0.0`), i.e. 100% error vs the correct halved velocity, with `workspace_reused=True` reported as if legitimate. Reproduced identically at np2 (silently wrong on both ranks, rank-symmetric).
- Mechanism is **double blindness**: `_operator_coefficient_variables` collects only *mesh variables* via `mesh_vars_in_expression`, so a pure constant yields `coefficient_states == ()` which matches forever; and `coefficients_unchanged` gates BOTH the `computeJacobian` call AND the matrix probe (`coefficients_unchanged or _matrix_matches_probe(...)`), so the safety-net probe can never fire on precisely the path where the counter heuristic fails — it would be probing a never-reassembled `J` anyway.
- The constants machinery underneath is fine: forcing `solve(time=0.0)` refreshes and gives the exact correct answer. The ramp just never reaches the operator on the fast path.

**Fix shape**: include the float values of rampable atomic constants reachable from the constitutive parameters / penalty / constraints in the invalidation key alongside the `_state` tuple — or more conservatively, when the tracked-variable set is empty, forfeit the shortcut: always reassemble and let the matrix probe decide (that keeps most of the win — the probe path still skips the ptap/GAMG rebuild when values are unchanged).

### 2. `mesh.deform()` between solves (the free-surface pattern) — **PASS**

A 2% top-boundary bump between solves flows correctly through `is_setup=False` → `_setup_solver` → `_reset_rotated_solver_cache`: next solve rebuilds (`workspace_reused=False`, new `Q` handle), and matches a fresh solver on the deformed mesh to machine zero, with no stale-rotation leakage. Moving-mesh workflows are safe.

### 3. np2 rank symmetry of the reuse decision — **PASS**

Four solves alternating RHS-only and viscosity-field changes: per-rank `workspace_reused` flags allgathered every solve agree at every step (`[F,F],[T,T],[T,T],[F,F]`), no hangs; the viscosity refresh is exact (`‖v−v/2‖ = 0` scaling check).

## Also: now CONFLICTING with development

The base has moved since this branch was cut: PR #377 (merged) added `_capture_rotated_report` reading the rotated result dicts, and a follow-up bugfix PR (imminent) adds `"rnorm"` (linear) / `"rnorm"`,`"rnorm0"` (nonlinear) to those dicts and computes the true rotated residual right where this PR restructures the solve tail. The rebase should preserve those keys — happy to help coordinate.

## Recommendation

Hold merge until (1) the rampable-constant blind spot is closed (probe 1) and (2) the branch is rebased over current development preserving the solve-report keys. Everything else probed is sound, and the memory/performance result stands.

lmoresi:
Superseded by #543 (merged as e475246f), which carries your commit with authorship preserved plus the rework onto the current rotated loop and the constants-aware invalidation key from the July review. Your 310-step Zhong A1 production validation remains the load-bearing evidence for the mechanism and is credited in the PR — when you have a slot, a 50-step guarded A1 restart segment on current development would re-anchor it against the reworked loop. Thanks for the mechanism and the exceptional validation trail.

Issue #425, opened 2026-07-26
Use the nested hierarchy for MG transfers on adapt children (topological prolongation)


lmoresi:
Design note: `docs/developer/design/nested-vs-geometric-mg-transfers.md` (commit e984a3b6).

## The inconsistency

`mesh.adapt()` maintains an exact refinement hierarchy — that is what newest-vertex bisection with conforming closure is *for*, and what the escape/halo cost buys. The MG transfer then discards it and re-derives an approximation geometrically:

```python
tri  = Delaunay(coarse_coords)          # re-triangulate the coarse DOF cloud
simp = tri.find_simplex(fine_coords)    # locate each fine DOF in one simplex
```

We pay for exact hierarchy maintenance and then do not use it.

This is the root cause of #424, not a separate concern: the barycentric builder has only **local support**, so moving fine coordinates (`relax()`, snapping, free-surface deformation) can leave a coarse DOF with no fine image -> zero column -> singular `PᵀAP`.

## The nested transfer is trivial

Every fine vertex is either an **inherited** coarse vertex (weight 1 on itself) or the **midpoint of a coarse edge** (weights ½, ½ on its endpoints). It is:

- exact for the un-deformed nested mesh (it IS the FE embedding);
- **structurally full rank** — every coarse DOF appears with weight 1 in its own inherited fine DOF, so #424 is impossible by construction;
- immune to node motion (topological, not geometric);
- 1–2 non-zeros per fine row, no Delaunay, no point location.

Both engines already record it: `edge2mid` in `NVBMesh` and `TaggedBisectionMesh`, and the native `nvb_transform.c` is a real `DMPlexTransform` with the same split-edge bookkeeping. Separately, `petsc_dm_set_regular_refinement` — our wrapper for the flag PETSc checks to take its exact nested interpolation path — appears exactly once in the repo: its own definition. Nothing calls it.

## Scope: keep BOTH paths

The geometric builder stays, for the non-nested pairs it was asked for (moved base vs adapted child, unrelated meshes, and the planned non-hierarchical variant with newly added seed points). The change is to stop using it where a topological relation exists.

Selection: **topological when the parent/child map is present**, geometric otherwise. The RBF retry from #424 then becomes a fallback that should essentially never fire on adapt children.

## The hard part — capture ordering

The map must be in **DM point numbering**, and the bridge from engine vertex ids to DM points is coordinate matching, which is exact only while midpoints are still exact float averages of their parents. Snapping and relaxation destroy that:

[...]

Issue #424, opened 2026-07-26
3D: shape relaxation can break the custom-P FMG build (zero-column transfer) — silent GAMG fallback


lmoresi:
Found while re-running the adaptivity demos with the new `mesh.relax()` / `adapt(..., relax=...)`.

In **3D**, relaxing an adapted mesh can make the mesh-owned custom-P geometric-MG hierarchy fail to build:

```
custom_mg: mesh-owned FMG build failed (transfer 6->7 has 26 zero columns
(coarse DOFs with no fine image) — BC-per-level reduction failed; coarse
operator would be singular.); using the solver default preconditioner.
```

The solver then **silently falls back to GAMG** — correct answers, but much slower, and nothing tells the user their geometric hierarchy is gone.

## Measured

Poisson MG gate, 3D box, `refinement=2`:

| config | cells | pc | its |
|---|---|---|---|
| adapt only | 24594 | `mg` | 2 |
| adapt + relax (at end) | 24594 | `mg` | 2 |
| adapt + relax `"per-generation"` | 24160 | **`gamg`** | **23** |

and in the larger `combine_mmpde_nvb_3d.py` demo (`max_levels=2`, ~36k cells), **relax-at-end also failed** — `transfer 6->7 has 26 zero columns`, `pc=gamg`, 26 its. So both placements are affected; whether it trips appears to depend on mesh size / depth rather than on the placement.

**2D is unaffected** in everything tested: `pc=mg` throughout, Stokes velocity block and Poisson, 2 its.

## Likely mechanism

The custom-P prolongation is built by locating fine DOFs inside coarse cells (barycentric). Relaxation moves the fine level coordinates; some coarse DOF then has no fine DOF mapping to it, giving a zero column and a singular coarse operator. 3D is presumably more exposed because tetrahedral point location is less forgiving than triangular.

## Why it matters

`adapt(..., relax=True)` is documented as safe for MG on the strength of 2D measurements. That claim is wrong in 3D and the docs are being corrected alongside this issue.

## Suggested directions

1. The fallback should be **loud** — an adapted+relaxed mesh silently losing its geometric hierarchy is the kind of thing that shows up as "why is my 3D solve slow" months later.
2. The transfer builder could tolerate zero columns by falling back to nearest-neighbour / RBF for the orphaned coarse DOFs rather than rejecting the whole level (an RBF builder already exists).
[...]

Issue #432, opened 2026-07-27
uw.function.evaluate returns wrong values for P1 fields at points exactly on cell edges (3D)


lmoresi:
Found while validating an MG prolongation against `evaluate` as the reference — the reference was the thing that was wrong.

## Symptom

Evaluating a **degree-1 scalar MeshVariable** at points lying exactly on coarse-cell **edges** (3D tetrahedral mesh) returns values that are not the FE interpolant at those points.

Independent check: each failing point lies on exactly **one** coarse edge `(a,b)`, so the P1 value there is unambiguously `0.5*(u_a + u_b)`. Comparing that against `evaluate`:

| point | independent edge-truth | `evaluate` |
|---|---|---|
| 504 | +0.700344 | +0.923709 |
| 565 | +0.323892 | −0.728195 |
| 2175 | −0.287598 | −2.759645 |
| 2177 | −2.122083 | −6.128067 |
| 2178 | −0.119434 | +0.782410 |
| 2205 | +0.361674 | +0.739306 |

Errors are O(1), not round-off. Affected points are a small fraction (2–12 per few thousand) but the wrong values are badly wrong.

## Why it matters

These are not exotic points. Every vertex introduced by edge bisection lies exactly on a parent edge, so any workflow evaluating a field at refined-mesh vertices, or at points shared between cells, is exposed. It is silent — no warning, no NaN.

## Reproducer sketch

```python
base  = uw.meshing.UnstructuredSimplexBox(minCoords=(0.,0.,0.), maxCoords=(1.,1.,1.),
                                          cellSize=0.5, refinement=1, qdegree=2)
child = base.adapt(metric, max_levels=2)          # child vertices lie on parent edges
uc = uw.discretisation.MeshVariable("uc", coarse_mesh, 1, degree=1)
uc.data[:, 0] = np.random.standard_normal(uc.data.shape[0])
fx  = child.dm.getCoordinatesLocal().array.reshape(-1, 3)
val = uw.function.evaluate(uc.sym[0], fx)         # wrong at points on coarse edges
```
Compare against `0.5*(u_a + u_b)` for the coarse edge each point sits on.

Full script: `/tmp/who_is_wrong.py` pattern — locate the coarse edge whose midpoint matches the query point, take the P1 value along it.

## Suspected cause

[...]

PR #433, opened 2026-07-27
Mesh shape relaxation, and the exact nested MG transfer for adapt children

lmoresi:
Follows the merged round-3 work (#409). Rebased onto current `development` — the old `feature/adaptivity-round3` branch had gone stale behind #416/#404/#412.

Closes nothing outright; advances #424 and #425, and depends on #420 being merged for the mover to work at all on graded meshes.

## What this adds

**`mesh.relax()` — shape relaxation.** Refinement picks node positions from combinatorics (which edge the tagging rule nominated), never from geometry, so a refined mesh carries needles that reflect the base mesh rather than the problem. The MMPDE mover could not fix this because its *reference frame* is the mesh it was handed, making a distorted mesh its own optimum — `redistribute_nodes(1)` moved exactly nothing. Adding `reference="ideal"` (a regular simplex, per-cell volume) measures deviation from equilateral-in-the-metric instead.

Exposed as `mesh.relax()` and `adapt(metric, max_levels=N, relax=...)` with two placements, **neither dominant**:

| property | unrelaxed | relax at end | relax per generation |
|---|---|---|---|
| P1 interpolation error | 2.75e-2 | 2.56e-2 | **2.38e-2** |
| 99th pct max angle | 112 deg | 110 deg | **96 deg** |
| on-fault size spread | 2.80 | **2.31** | **2.31** |
| far-field closure halo | 42.8 | **27.7** | 43.9 |

Default is at-end (cheaper, cannot change topology). Documented side by side rather than ranked.

**Exact nested MG prolongation.** `adapt()` now records the true coarse-to-fine embedding per refinement pass, built from the parent/child relation rather than re-derived by point location. Verified: partition of unity exact, **zero** zero-columns, exact reproduction of an arbitrary coarse field, 1.0-1.4 nnz/row. Installed for **degree-1** fields; higher degree still uses the geometric builder (stage 2, #425).

**Orphan repair in the geometric builder (#424).** Point location has local support, so moving fine coordinates can leave a coarse DOF with no fine image -> zero column -> singular `PtAP`. Each orphan now takes its nearest fine DOF as a pure injection, preserving partition of unity and ~d+1 sparsity. 3D `relax="per-generation"`: `gamg` 23 its -> `pc=mg` 2 its.

**scipy removed from the nested path.** `nvb.py` is now scipy-free; the coordinate lookups were exact-equality identifications wearing a nearest-neighbour search as a disguise, and a kNN search always returns *something* — it would have silently bound the wrong vertex if its assumption broke.

## Testing

41 tests pass across relax, nested prolongation, graded adapt and the custom-MG suites. New: `test_0752` (7 cases), `test_0753` (11 cases).

Adversarial review to follow as a comment before this is marked ready.

### DIFF
--- a/docs/developer/design/nested-vs-geometric-mg-transfers.md
+++ b/docs/developer/design/nested-vs-geometric-mg-transfers.md
@@ -1,0 +1,127 @@
+# Two MG transfer paths, each used properly
+
+**Status**: PROPOSED (2026-07-26)
+
+## The problem
+
+`mesh.adapt()` maintains an exact refinement hierarchy — that is the entire
+point of newest-vertex bisection with conforming closure, and it is what the
+"escape"/halo cost buys. The multigrid transfer then **throws that hierarchy
+away** and re-derives an approximation to it geometrically:
+
+```python
+def barycentric_prolongation(coarse_coords, fine_coords):
+    tri  = Delaunay(coarse_coords)          # re-triangulate the coarse DOF CLOUD
+    simp = tri.find_simplex(fine_coords)    # locate each fine DOF in one simplex
+```
+
+The coarse mesh's own cells are discarded; the parent/child relation is never
+consulted. We pay for exact hierarchy maintenance and then do not use it.
+
+This is not merely inelegant — it is the direct cause of issue #424. The
+barycentric builder has only **local support**: a coarse DOF is reached only if
+some fine DOF happens to land in a Delaunay simplex touching it. Move the fine
+coordinates — `mesh.relax()`, surface snapping, free-surface deformation — and
+a coarse DOF can lose every fine image. Its column in `P` goes to zero,
+`PᵀAP` acquires a zero row and column, and the coarse operator is singular.
+
+## Why the geometric builder exists (and should stay)
+
+It was built for genuinely **non-nested** pairs, where no topological relation
+exists at all:
+
+* a moved base mesh against an adapted child,
+* two independently generated meshes,
+* the planned non-hierarchical variant with newly added seed points.
+
+That capability was asked for and is worth keeping. The error was applying it
+to the *nested* case as well, where the exact answer is already known.
+
+## The nested transfer is trivial and cannot go singular
+
+For a bisection hierarchy every fine vertex is exactly one of two things:
+
+* an **inherited** coarse vertex → weight 1 on itself;
+* the **midpoint of a coarse edge** `(a, b)` → weights ½, ½ on `a` and `b`.
+
+Properties that matter:
+
+* **Exact** for the un-deformed nested mesh (it *is* the FE embedding).
+* **Structurally full rank** — every coarse DOF appears with weight 1 in its
+  own inherited fine DOF, so a zero column is *impossible by construction*.
+  #424 cannot occur in this formulation.
+* **Immune to node motion.** The relation is topological, so relaxation,
+  snapping and deformation do not disturb it.
+* **Cheap** — 1 or 2 non-zeros per fine row, no Delaunay, no point location.
+
+The engines already record exactly this. `nvb.py` keeps
+
+```python
+self.edge2mid = {}     # (a, b) -> midpoint vertex id
+```
+
+for both the 2D `NVBMesh` and the 3D `TaggedBisectionMesh`, and the native
+`nvb_transform.c` is a real `DMPlexTransform`, whose split-edge bookkeeping
+carries the same information.
+
+Note also that `petsc_dm_set_regular_refinement` — our own wrapper for the flag
+PETSc checks to take its exact nested interpolation path — appears exactly once
+in the repository: its own definition. Nothing calls it.
+
+## The one hard part: *when* the relation can be captured
+
+The relation must be recorded in **DM point numbering**, and the bridge from
+engine-internal vertex ids to DM points is coordinate matching (this is how
+`to_dm` already transfers labels). Coordinate matching is exact only while
+midpoints are still exact float averages of their parents.
+
+Both snapping and relaxation destroy that:
+
+* **native path** — `_nvbx.refine()` produces the DM, *then*
+  `snap_level_boundaries()` moves boundary vertices. There is a clean window
+  between the two where coordinates are pristine.
+* **cell-list path** — the snap is applied to `nvb.coords` *before* `to_dm()`
+  is called, so by the time a DM exists the window has closed.
+
+So the capture must be designed in, not bolted on:
+
+1. capture the parent/child map immediately after refinement, **before** any
+   snap or relaxation, and store it on the child;
+2. or have `to_dm()` return its engine-id → DM-point map directly, removing the
+   dependence on coordinate matching altogether (preferable — it is exact
+   regardless of ordering).
+
+Option 2 is the more robust and is the recommended route.
+
+## Proposed shape
+
+* Each generation stores a sparse vertex-level prolongation on the child, e.g.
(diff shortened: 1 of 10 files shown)


lmoresi:
## Adversarial review (author, pre-review)

Trying to break my own work. Six things a reviewer should push on, roughly in order of how much they would cost if I am wrong.

**1. The nested transfer is validated against an oracle I built myself.** `test_0753::test_reproduces_an_arbitrary_coarse_field` computes the reference by locating the coarse edge each vertex lies on and taking the P1 value along it. That is independent of `uw.function.evaluate` **on purpose** — evaluate is wrong at exactly these points (#432) — but it shares my assumption that every bisection vertex lies on a coarse edge. The test asserts `checked > 0.9 * n_vertices`, and in 3D 1-2 vertices per pass are *not* on any coarse edge and go unchecked. If those are the ones that are wrong, this test cannot see it.

**2. I got the correctness of this code wrong twice today, in both directions.** I first claimed the transfer was the exact FE embedding on the strength of a linear-reproduction test that cannot distinguish a correct embedding from a mis-attributed one. I then removed a working component after a test using `evaluate` as the reference showed 2-12 wrong DOFs per 3D pass, and wrote a confident and entirely wrong explanation of why it was unsound. Both were caught only by building a third, independent oracle. **Weight my claims here accordingly** — the code is now measured, but my track record on reasoning about it unaided is poor.

**3. The stage-1 restriction is narrower than it sounds.** `degree == 1` means P1 Poisson-like problems. **Stokes never touches this path** — velocity is P2, and instrumentation shows the nested transfer declined 3 times out of 3. Advection-diffusion typically wants P3. So the headline "exact nested transfer" currently does nothing for either production workhorse. A reviewer might reasonably ask whether it is worth merging before stage 2; my answer is that it is the reference implementation the geometric path is priced against, but that is a judgement call.

**4. The nested transfer is measurably WORSE on iterations in 3D.** Poisson P1, `k=8pi`, `rtol=1e-12`: nested 5,6,5 against geometric 4,5,4 — consistent across three configurations, not noise. Likely stencil width (nested is 1.0-1.4 nnz/row, near-injection; barycentric spreads over ~4). Its coarse operators are ~3x sparser so it may still win on wall-clock, **but I have not measured wall-clock**. If someone merges this expecting a speedup they may be disappointed.

**5. The orphan repair is incomplete and I know it.** On a deliberately pathological pair (fine cloud covering 60 % of the coarse hull) orphans remain and the RBF retry still fires — and that retry produces a **fully dense** transfer (`nnz/row == n_coarse`), which would be ruinous at production size. It does not arise for genuine coarse/fine mesh pairs, where both cover the same domain, but the failure mode is a silent performance cliff rather than an error.

**6. Unexercised code paths.** The cell-list refinement engine keeps its own `edge2mid`-based capture and its own relax hook; **neither is covered**, because both 2D and 3D prefer the native transform when it is built. I previously claimed the 3D test covered the cell-list hook — that was wrong and is corrected in the test docstring, but the coverage gap is real. Also `_adapt_parent_cells` is recorded but consumed by nothing, and declines on 3 of 6 3D passes.

**Dependencies:** this needs #420 (per-cell collapse floor) merged, or the mover silently does nothing on any graded mesh and `relax()` is a no-op.

**What I am confident about:** the recorded transfers satisfy partition of unity, have zero zero-columns, and reproduce an arbitrary coarse field to ~1e-15 across every pass in 2D and 3D; `adapt()` output is bit-identical to before this branch when `relax` is off (checksummed against the pre-change commit); and 41 tests pass.

lmoresi:
## Staleness and review follow-ups

### Merge hazard found and defused

This branch **duplicates PR #420** (the per-cell collapse floor). It was cherry-picked from a branch that already carried that fix, while `development` still has the old median-based version.

The two copies were functionally identical but differed in em-dashes and a missing `See #419.`, which would have produced a pointless comment-only conflict. The block is now **byte-identical to #420 s**, so:

**Merge order: #420 first, then rebase this branch.** The duplicate hunks drop out silently rather than conflicting. Merging this one first would leave #420 as a no-op against a fix it cannot see.

### Staleness, for the record

The original `feature/adaptivity-round3` branch is **stale and should not be used**. PR #409 was squash-merged on 2026-07-24, and `development` has since moved on via #416, #404 and #412 — a PR from that branch would have shown ~3445 lines of spurious deletions. All 14 commits were cherry-picked onto current `development` here and re-verified (built, 41 tests). That branch can be deleted; I have not done so, since it is not mine to remove.

### Review item 5 addressed

The RBF retry now **says** it is dense. It rescues correctness when the orphan repair cannot place every coarse DOF, but returns `nnz/row == n_coarse`, making the Galerkin coarse operators dense — harmless on the meshes where it has fired, a silent performance cliff at production size, exactly the failure mode the GAMG fallback had. The warning now states that and says to fix the cause rather than accept the symptom.

### Review items NOT addressed, and why

- **(1) Oracle gap.** 1-2 vertices per 3D pass do not lie on a coarse edge and go unchecked by `test_reproduces_an_arbitrary_coarse_field`. Closing it needs a reference for points interior to a coarse face/cell — real work, and it is the gap most likely to hide a genuine error. Worth doing before this is relied on.
- **(4) Wall-clock unmeasured.** The nested transfer is one iteration worse in 3D but ~3x sparser; whether it wins on time is unknown. It should not be merged as a claimed speedup.
- **(6) Cell-list engine paths still uncovered.** Both its prolongation capture and its relax hook are unexercised, because both 2D and 3D prefer the native transform when built. Forcing that path in a test is straightforward (stub `_nvbx`) and has not been done.

23 tests re-run green after these changes.

lmoresi:
## Restacked — the chain is no longer fragile

The duplication is **gone**, not merely aligned. This PR is now stacked properly:

```
#420  bugfix/mmpde-graded-area-floor -> development   [MERGEABLE]
#433  feature/mesh-relax-and-nested-mg -> #420        [MERGEABLE]
```

Verified: `git diff #420...HEAD -- mmpde.py` contains **zero** references to `a0_own` / `_min_area_frac`. The collapse-floor fix now exists in exactly one place — #420 — and this branch carries only its own changes.

What that buys:

- **This PRs diff shows only its own work.** Reviewing it no longer means mentally subtracting #420.
- **No merge-order trap.** GitHub retargets this PR to `development` automatically when #420 merges. Previously, merging in the wrong order would have left #420 as a no-op against a fix it could not see.
- **No cosmetic conflict.** The only conflict during the restack was em-dashes and a missing `See #419.`; resolved in favour of #420s text, so #420 is canonical for that hunk.

Rebuilt and re-tested on the new base: **37 tests pass** (relax, nested prolongation, graded adapt, custom-MG Stokes). Pre-restack state is preserved locally as `backup/mesh-relax-pre-restack` in case the history needs checking.

### Remaining chain hygiene, not done by me

`feature/adaptivity-round3` is dead and should be deleted — #409 was squash-merged on 2026-07-24 and `development` has moved on via #416/#404/#412. Everything from it lives here. I have left the deletion to you rather than removing a branch I do not own.

### Still open from the adversarial review

Unchanged by the restack, and listed again so they are not lost:

1. **Oracle gap** — 1-2 vertices per 3D pass lie on no coarse edge and go unchecked. The likeliest place for a real error to hide.
2. **Wall-clock unmeasured** — nested is one iteration worse in 3D but ~3x sparser. Do not merge as a claimed speedup.
3. **Cell-list engine paths uncovered** — both its capture and its relax hook; forcing that path needs `_nvbx` stubbed in a test.

Issue #431, opened 2026-07-27
CBF sigma_nn recovery on deformed/faceted boundaries: ~2.4% amplitude deficit (chord-vs-arc boundary mass suspected)


lmoresi:
> **CURRENT SCOPE (2026-07-27)** — the report below is the ORIGINAL filing and its
> headline mechanism is **retracted** (probe artifact: radius-mask surface selection on a
> deformed ring; see the correction comment below). The recovery is exonerated at ~0.976
> of exact. What this issue now tracks is only:
>
> **The CBF σ_nn recovery on a deformed, faceted boundary under-reports amplitude by
> ~2.4%** (load-only 0.976, driving 0.972 on a 5%-deformed ring at cellSize 0.06).
> Chord-vs-arc boundary mass on the faceted ring is the suspected bulk. Likely related to
> the #414 curved-boundary error family. Next step before any fix: measure how the
> deficit scales with deformation amplitude and resolution. The free-surface pathology
> originally attributed here was a sign error in the FreeSurface manager, fixed in PR #435.

---

## Summary

On a deformed boundary, the CBF reaction recovery behind `dynamic_topography` /
`boundary_normal_traction` returns only **~0.16 of the current-shape load component**
rho_0*g*h of the constraint reaction, while recovering the dynamic (driving) component
exactly. Any sigma_nn measurement on a deformed boundary under a full-density body force
is biased by this; in the FreeSurface manager it made the full-Boussinesq scheme park the
surface at 24% of isostasy with steady O(vrms) flow through a stationary surface.

## Decisive probes (annulus r_in=0.5, r_out=1.0, cellSize=0.06, rotated free-slip lid)

**A — pure statics, the recovery has nothing to hide behind:** T=0, body force
-rho_0*g*rhat (rho_0*g=1e6), mode-4 boundary deformation amplitude 0.05. The discrete
solve is exact hydrostatics (u=0, p'=const); the constraint reaction is rho_0*g*h(theta)
to solver precision. Recovered (mass="lumped"): **m=4 amplitude 0.158 of exact**, phase
correct (-179.9 deg for the -(sigma-mean)/rho_g convention).

**B — the dynamic leg is clean:** driving-only vs driving+full-background on the
UNDEFORMED mesh (self-load = 0): recovered m=4 identical to 4 decimals (ratio 1.0000).
The huge hydrostatic RHS does not corrupt the recovery; the loss is specific to the
deformed-boundary load.

## Suspected mechanism (unconfirmed)

[...]

Issue #423, opened 2026-07-26
Old-frame SL reach-back amplifies exponentially on per-step-moving meshes (free-surface T blow-up; affects movers/adapt identically)


lmoresi:
## Summary

The serial free-surface temperature unboundedness (T reaching [-1.6, 3.2] at ~8% surface
deformation in annulus FS convection) is an **exponential instability of the semi-Lagrangian
record -> trace -> solve loop that requires per-step mesh motion and `old_frame_traceback=True`**.
It is not a landing-point problem, not the monotone limiter, not the flux-history frame, and not
the timestep. The free surface itself is exonerated as physics: its only necessary role is
moving the mesh between solves, so **MMPDE movers and adapt-on-top workflows are exposed
identically**.

## Evidence chain (all measured)

**1. Stage attribution** (annulus FS convection, gap-Ra 3e4, rho_g 2e5, res 0.06): per-step
bounds of T_pre / psi_star-after-clamped-traceback / T_post show the clamp invariant
`range(psi*) ⊆ range(T_pre)` holds at every step — the traceback *contracts* the range —
while the FE solve grows BOTH extremes ~8-10%/step. The worst violator is mesh-locked at
r=0.952 for 22 consecutive steps (hot BC is at r=0.5): a local linear instability, not
advected material. Onset at h/R ≈ 5.4%.

**2. Lever scan**: theta_flux=1.0 (BE) delays onset and saturates at a period-2 limit cycle
([-0.38, 1.51]) instead of running away — an accelerant removed, not a cause. Old-frame for
the *flux* history: no effect. **True dt halving makes it WORSE at matched deformation** —
the gain is per loop *cycle*, not per unit time.

**3. Minimal reproducer** (`slcn_minimal_repro.py`, attached path below): no Stokes, no free
surface — prescribed 4-cell streamfunction velocity, sharp (1-2 cell) cold boundary layer,
mode-4 squeezed annulus:

| variant | 60-step outcome |
|---|---|
| undeformed, static mesh | bounded [-0.05, 1.02] |
| squeezed 9%, static | mild linear creep, max -> 1.26 |
| squeezed 9%, moving ±0.09%R/step | **exponential runaway, T -> 160** |

**4. Localisation** (2x2 under motion):

| | theta 0.5 | theta 1.0 |
|---|---|---|
| `old_frame_traceback=True` | T -> 160 | T -> 46 |
[...]

Issue #421, opened 2026-07-26
FreeSurface: prescribed surface datum loses ~7% net flux in parallel (surface-array <-> field round trip)


lmoresi:
## Summary

`uw.systems.FreeSurface`'s consistent (third) solve prescribes `u.n = ũ_n` so the surface
stays a *material* boundary. The datum field `_un_target` carries a **~6-7% net flux on 2
ranks versus ~0.06% in serial** — a ~100x degradation that is **independent of how the
surface mean is removed**, so it is not the demean.

An incompressible interior over a closed no-slip base can neither gain nor lose volume, so
`∮ ũ_n ds = 0` is required. In parallel it is badly violated.

## Measurement

Isoviscous annulus (`r_in=0.5, r_out=1.0, cellSize=0.1`), gap-Ra 3e4, `rho_g=5e5`,
3 steps at `advect_scale=10`. Net flux = `|∮ ũ_n ds| / ∮|ũ_n| ds`, both by exact FE
`uw.maths.BdIntegral` over the surface (not the manager's own nodal sum, so it cannot pass
by construction):

| ranks | node-count demean | arc-length demean |
|---|---|---|
| 1 | 8.96e-3 | **5.63e-4** |
| 2 | 7.06e-2 | 5.75e-2 |

## What is NOT the cause

The ring machinery and the arc-length quadrature are **parallel-exact**. On np2 the
gathered ring has 66 entries for 64 distinct nodes (the 2 partition-seam vertices appear on
both ranks) and the trapezoidal rule splits their weight correctly:

- `sum(weights)` = 6.2831853072 = `2*pi*r_out` to machine precision
- weighted mean of `1` = 1.000000
- weighted mean of `cos(k*theta)`, k = 1,2,3 = ~1e-16

Pinned by `test_freesurface_ring_quadrature_is_exact_in_parallel` (runs on any rank count).

So the loss is in the **surface-array <-> field round trip across the partition seam**: the
demeaned ring array is flux-free by construction, but reading the values back out of
`_un_target` is not. Suspects, in order:

1. `_field_rows` matches local surface coords to local field rows with a KD-tree; behaviour
   at seam vertices that are owned by one rank and ghost on the other is unverified.
2. Writing `_un_target.array[rows] = u_tilde` sets local values only — a ghost/owner scatter
   may overwrite the seam DOF, or the two ranks may disagree on its value.
3. `_surface_node_coords` returns seam vertices on both ranks (by design, for the ring);
[...]

Issue #402, opened 2026-07-23
FreeSurface: composition transport is level-set-only; can't drive temperature convection without private access


lmoresi:
## Summary

`uw.systems.FreeSurface` accepts a `composition` field and advects it with the material-consistent surface velocity (the correct way to keep a free surface a material boundary). The constructor docstring lists **temperature** as an intended use ("a mesh level set holding a signed distance, or temperature"). But `_build_composition_transport` hardcodes level-set defaults that are wrong for a temperature/convection field:

- `diffusivity = 1.0e-7` (a level set carries no real diffusion; temperature needs the thermal κ),
- `bcs = []` (temperature needs Dirichlet BCs at the surface/base),
- `monotone_mode = None` (a bounded [0,1] temperature typically wants monotone limiting).

## Impact

Building a free-surface **convection** driver on the manager currently requires reaching into the private transport solver after construction:

```python
fs = uw.systems.FreeSurface(stokes, "Upper", ..., composition=T)
fs._comp_adv.constitutive_model.Parameters.diffusivity = kappa   # private
fs._comp_adv.add_dirichlet_bc(1.0, "Lower")                      # private
fs._comp_adv.add_dirichlet_bc(0.0, "Upper")                      # private
```

This works (a clean annulus FS-convection run is up and stable on it) but is fragile — it depends on the private `_comp_adv` name and on BCs being addable post-construction.

## Suggested resolution (either)

1. Parameterize the composition transport: `composition_diffusivity`, `composition_bcs`, `composition_monotone` on the constructor; or
2. Expose the material-consistent advection velocity publicly (e.g. `fs.advection_velocity`) and run the consistent solve even when `composition is None`, so callers can drive their own AdvDiff with proper κ/BCs/monotone.

Option 2 is more flexible (temperature convection wants its own AdvDiff anyway); option 1 is more turnkey. Worth a design decision.

## Context

Found while building a fresh annulus free-surface convection driver on the hardened `FreeSurface` manager (`feature/free-surface-tangent`). Reference: `docs/developer/design/FREESLIP_DYNAMIC_TOPOGRAPHY_FREESURFACE.md`, skill `free-surface-convection`.

Underworld development team with AI support from Claude Code

PR #435, opened 2026-07-27
Free-surface hardening: material-boundary gauge, moving-mesh transport fix, full-density (Boussinesq) formulation

lmoresi:
## What this delivers

The free-surface toolkit (uw.systems.FreeSurface) taken from "stable only for stiff
surfaces at small deformation" to a validated capability: bounded temperature transport
to at least 17% surface deformation, a correct material-boundary constraint, and the
full-density (Boussinesq) formulation with its surface restoring force — the setting in
which topography relaxes at the physical rate and equilibrium states have closed
streamlines with no flow through the surface.

## The four main fixes, each with its measurement

**1. Material-boundary gauge (2c522c5d, 8f490069).** The rate prescribed to the
consistent solve must carry zero net flux on a closed boundary; the demean removed the
node-count mean instead of the arc-length mean, leaving ~1% net flux that a strong
constraint cannot absorb. With the arc-length gauge the strong rotated datum beats the
penalty on every measure: rate error 5.9e-3 vs 3.8e-2, net volume leak 7.0e-5 vs
4.4e-2, and the hot-injection Nu inflation disappears (3.06 vs 5.69 at matched steps).

**2. Old-frame reach-back retired from the manager (b507aca1, issue #423).** The serial
temperature blow-up beyond ~5% deformation was localised by ablation to
old_frame_traceback=True — an exponential instability of the record/trace/solve loop
whose gain is per loop cycle (halving dt makes it worse). A minimal reproducer with no
free surface at all (prescribed velocity + a 0.1%/step mesh wobble) reproduces it;
with the standard ALE path the same configuration is bounded at any theta. The manager
now uses the standard ALE path; envelope measured from 2.6% to 17% deformation with
T in [0,1] to 1e-3. Guard test: test_1071_slcn_moving_mesh.py.

**3. Timestep estimate on deformed meshes (84f86794).** The no-composition CFL path
evaluated velocity by point location at mesh vertices, which legitimately NaNs on a
deformed mesh; it now reads nodal data directly.

**4. Full-density formulation support (997a6b0d, 98961147).** With
rho = rho_0(1 - alpha dT) the free solve carries the surface restoring force
(topographic relaxation at the Cathles rate; without the background there is provably
no restoring force and topography grows without limit). The recovered held-lid
reaction then contains the self-load of the current shape, and the pre-existing sign
convention flips it — the surface parks at half its equilibrium with steady flow
through it. Fix: background_buoyancy="analytic" subtracts the geometric current
height (no extra solve); a twin-solve expression mode is retained as the exact
reference; both validated identical on the relaxation benchmark and an equilibrium
restart (surface climbs from 24% to the ~80%-of-isostasy dynamic equilibrium, with
the through-flow ratio collapsing 2.15 -> 0.075 in one step). The CBF recovery itself
was verified essentially exact (0.976/0.972 on a 5%-deformed ring, exact
superposition) — issue #431 records the retraction of an interim mis-attribution and
is rescoped to the benign 2.4% residual.

Default behaviour (no background_buoyancy, reduced formulation) is unchanged and
covered by the existing tests.

## Tests

23 serial (test_1018, test_1070, test_1071) + 6 at np2 green. Two new datum tests are
skipped at np>1 against pre-existing issue #421 (parallel datum round trip), with the
measurement in the skip reason.

## Issues touched (none auto-close; to be closed by hand where appropriate)

- #421 parallel datum round trip — pre-existing, documented, open.
- #423 old-frame amplification — manager-side fix landed here; the source-level "why"
  remains open.
- #431 CBF recovery residual — rescoped after retraction; open, low priority.
- #402/#403 — unchanged scope, not addressed here.

An adversarial self-review follows as the first comment.


lmoresi:
## Adversarial self-review (house rule: posted before marking ready)

Attacking the weakest points of this PR honestly:

**1. The parallel exposure is real and is shipped as a known hole.** The strong
rotated datum is now the consistent-solve default, and #421 shows the datum field
carries ~6% net flux at np2 — the exact over-determined case the strong constraint
cannot absorb. Serial free-surface runs are validated; parallel free-surface runs
with composition should be treated as unsupported until #421 lands. Mitigation
available today: consistent_constraint="penalty" per run. Reviewers should weigh
whether the default should be parallel-size-gated instead.

**2. The commit history contains a corrected mis-attribution.** 997a6b0d's message
claims the CBF recovery loses 84% of the load; 98961147 retracts that (probe
radius-mask artifact) and identifies the real one-sign defect. History is honest in
sequence but a reader of 997a6b0d alone will be misled — the PR description carries
the corrected narrative, and squash-merge would remove the hazard entirely.
Recommend squash.

**3. The 'analytic' subtraction leans on the recovery being faithful (0.976).** The
2.4% residual acts as a small per-step error in h_inf. At the validated cases it is
invisible (bit-identical trajectories at reported precision vs the exact two-reaction
mode), but at much larger deformation or much coarser resolution the residual could
grow — nobody has measured its scaling. The exact mode is retained precisely for
this; a scaling check is cheap follow-up work.

**4. Old-frame retirement is manager-scope only.** old_frame_traceback=True remains
available on SemiLagrangian and is now known to be an amplifier on per-step-moving
meshes (#423) with no guard outside the manager. Any other mover/adapt workflow that
enables it inherits the hazard. A warning in the ddt docstring exists; a runtime
warning when it is enabled alongside per-step deforms would be stronger.

**5. Test-time cost.** test_1070 grew two level_2 tests (~90s) and test_1071 adds
~60s. Within house norms for tier_b but noted.

**6. What is NOT validated here:** 3D (all measurements are 2D annulus/box);
nonlinear rheology with the strong datum (raises NotImplementedError with a clear
message — deliberate); statistical steady state (all runs are developing or
restarted-near-equilibrium states).

None of these block the merge in my judgement; 1 and 2 deserve reviewer attention.

lmoresi:
Update to the adversarial review, point 1: the parallel datum defect (#421) is now FIXED in this PR (67d811d5 — ring seam deduplication + direct nodal velocity reads). The datum tests run and pass at np2; no parallel gating of the strong-constraint default is needed. Point 1 is withdrawn; points 2-6 stand. Also added in the same commit: an explicit NotImplementedError for 3D (previously would run silently wrong; follow-on work via #404's boundary mass), per discussion.

lmoresi:
Roadmap for the follow-on work is now in the PR: docs/developer/design/FREE_SURFACE_ROADMAP_2026-07.md — three workstreams (A: land this PR + teaching rebuild + stationarity run; B: nonlinear datum + rotated-path unification, joint session, tracking #438; C: 3D, structural, builds on #404, suggested for @gthyagi), with sequencing and the deferred list. C is assembly work, not new mathematics — the integrator, datum constraint, deform carrier and full-density h_inf are all dimension-free; the 2D-only pieces and their 3D replacements are tabulated.

PR #446, opened 2026-07-27
Share the nnn=1 guard between both interpolation entry points; delete the dead RBF surface

### LINKED ISSUE #443: interpolation_matrix(nnn=1, order=1) silently returns a nearest-neighbour operator where the value path raises
Found in a retrospective adversarial review of #430 after merge.

## The inconsistency

`KDTree.rbf_interpolator_local` rejects `nnn=1` with `order=1`, because one neighbour cannot determine an affine tail:

```python
kdt.rbf_interpolator_local(tgt, f, 1, 2, False, order=1)
# ValueError: order=1 needs at least dim + 2 neighbours ...
```

`KDTree.interpolation_matrix` does not:

```python
T = kdt.interpolation_matrix(tgt, nnn=1, order=1)
# no error; nnz/row = 1 -- a nearest-neighbour operator
```

Measured on `development` at `4d043aeb`.

## Why it matters

The caller asked for a linear-exact operator and silently got a nearest-neighbour one, which reproduces *constants* only. Nothing in the returned object records the downgrade. The whole point of the `order=1` path is a guarantee; handing back an operator that does not have it, without saying so, is the failure mode the rest of that PR was written to avoid.

It also contradicts the PR's own claim that the value and operator paths are built on one core and "cannot drift apart". They share `_local_stencil`, but `_local_stencil` early-returns for `nnn == 1` *before* reaching the `order` handling, and only `rbf_interpolator_local_from_kdtree` carries the guard — `interpolation_matrix` does not.

## Fix

Move the `nnn == 1 and order == 1` guard into `_local_stencil`, so both callers inherit it. That is where the rest of the shared behaviour already lives.

## Also worth tightening while there

- `interpolation_matrix` with `nnn > self.n` raises `RuntimeError: Error in rbf_interpolator_local_from_kdtree - a nearest neighbour wasn't found`, naming a function the caller never invoked. Correct behaviour, confusing message.
- The degenerate-fallback `warnings.warn(..., stacklevel=3)` is tuned for the value path's call depth; from `interpolation_matrix` the depth differs by one, so attribution is off by a frame.

## Test gap

`test_0102_kdtree_linear_rbf.py::test_interpolation_matrix_agrees_with_the_value_path` parametrises `order` but not `nnn`, and never exercises `nnn=1`. A guard test on both entry points would have caught this.

lmoresi:
Fixes #443, #428.

## #443 — the two paths drifted exactly where I said they could not

`interpolation_matrix(nnn=1, order=1)` silently returned a **nearest-neighbour** operator where the value path raised:

```python
kdt.rbf_interpolator_local(tgt, f, 1, 2, False, order=1)   # ValueError
kdt.interpolation_matrix(tgt, nnn=1, order=1)              # no error, nnz/row = 1
```

The caller asked for a linear-exact operator and got a constants-only one, with nothing recording the downgrade.

**Cause**: `_local_stencil` early-returns for `nnn == 1` *before* any `order` handling, and the guard sat in `rbf_interpolator_local_from_kdtree` alone. Sharing a core is not the same as sharing its guards. #430's body claimed the two paths "cannot drift apart"; this is how they did.

The guard moves into `_local_stencil`, so both entry points inherit it. Verified on both, with `nnn=1, order=0` still legal on both.

Two smaller things at the same site:

- The stencil-too-large error read `Error in rbf_interpolator_local_from_kdtree - a nearest neighbour wasn't found`, naming a function the caller never invoked. It now reports the requested stencil size and the tree size: `Cannot build a 20-point stencil: the kd-tree holds 5 point(s)...`
- The degenerate-fallback warning used `stacklevel=3`, tuned for the value path's depth. The two entry points sit at different depths below `_local_stencil`, so it is now `2` — the deepest frame correct for both.

## #428 — dead RBF surface removed

**`MeshVariable.rbf_interpolate(meth=, rubbish=)`.** Neither was ever read. `meth` was documented as "reserved, currently unused", which Charter §5 rules out.

Removing `meth` also fixes a latent test bug: `tests/test_0505_rbf_swarm_mesh.py` calls `var.rbf_interpolate(rx, nnn, p)`, so its `nnn` landed in `meth` and was discarded — every parametrised case actually ran at the default stencil, including the `(2, 1, 2)` case that believed it was testing the raw nearest-neighbour branch. With `meth` gone the positional argument means what the test always intended, and the file passes unchanged.

Checked: no other caller passes a second positional argument.

**`_rbf_reduce_to_meshVar`, `Swarm._get_map`, `Swarm._nnmapdict`, `_nn_proxy`.** Zero callers. Deleting the trio is the disposition already recorded for SWARM-23 in `design/SWARM_MODERNIZATION_DESIGN_2026-07.md`, which notes `_get_map` caches particle indices against a stale key and is a live trap if anything revives it. `_nn_proxy` existed only to gate a branch inside the dead method and was plumbed out to `Swarm.add_variable`.

## Regression

- level_1 `tier_a`/`tier_b`: **551 passed, 0 failed**
- level_2 `tier_a`/`tier_b`: **319 passed, 0 failed**
- np=2 parallel proxy/index-swarm: 4 passed
- Deprecated-pattern scan: clean

105 lines deleted, 65 added.

### DIFF
--- a/src/underworld3/ckdtree.pyx
+++ b/src/underworld3/ckdtree.pyx
@@ -716,7 +716,9 @@ cdef class KDTree:
         # must trip this guard, so the comparison is >= (issue #399).
         if np.any(closest_n >= self.n):
             raise RuntimeError(
-                "Error in rbf_interpolator_local_from_kdtree - a nearest neighbour wasn't found"
+                f"Cannot build a {nnn}-point stencil: the kd-tree holds "
+                f"{self.n} point(s) and a nearest neighbour wasn't found. "
+                "Reduce nnn, or check that the tree is not empty."
             )
 
         # np.bool_, not bool: this module cimports the C++ `bool` from libcpp,
@@ -724,6 +726,16 @@ cdef class KDTree:
         degenerate = np.zeros(coords_converted.shape[0], dtype=np.bool_)
 
         if nnn == 1:
+            # The guard lives HERE, not in the callers: this early return
+            # precedes all `order` handling, so a caller-side check leaves the
+            # other entry point silently returning a nearest-neighbour stencil
+            # when a linear-exact one was asked for (issue #443).
+            if order == 1:
+                raise ValueError(
+                    "order=1 needs at least dim + 2 neighbours to determine "
+                    "the affine tail; nnn=1 selects the raw nearest-neighbour "
+                    "path, which reproduces constants only."
+                )
             return closest_n, np.ones(closest_n.shape), None, degenerate
 
         # can decompose weighting vecotrs as IDW is a linear relationship
@@ -771,7 +783,10 @@ cdef class KDTree:
                 f"{degenerate.size} stencils could not support an affine fit "
                 "(collinear/coplanar neighbours) even after widening, and fell "
                 "back to inverse-distance weighting.",
-                stacklevel=3,
+                # 2 = whichever public method called _local_stencil. The two
+                # entry points sit at different depths below that, so this is
+                # the deepest frame that is correct for both.
+                stacklevel=2,
             )
 
         return closest_n, linear_weights, wide, degenerate
@@ -921,24 +936,14 @@ cdef class KDTree:
             print(f"Mapping values with nnn - {nnn} & p {p}  ... start", flush=True)
 
         if nnn == 1:
-            # only use nearest neighbour raw data
-            if order == 1:
-                raise ValueError(
-                    "order=1 needs at least dim + 2 neighbours to determine the "
-                    f"affine tail; nnn=1 selects the raw nearest-neighbour path."
-                )
-            closest_n, _ = self.find_closest_n_points(
-                1, np.ascontiguousarray(coords_converted, dtype=np.float64)
+            # Only the raw nearest-neighbour value; _local_stencil holds the
+            # order=1 guard so both entry points share it (issue #443).
+            closest_n, _, _, _ = self._local_stencil(
+                coords_converted, nnn, p, order
             )
-            # (n, 1) -> (n,): the nearest-neighbour path returns data rows
-            # directly, so the stencil axis must not survive into the result.
-            # query(k=1) used to do this reshape for us.
-            closest_n = closest_n.reshape(-1)
-            if np.any(closest_n >= self.n):
-                raise RuntimeError(
-                    "Error in rbf_interpolator_local_from_kdtree - a nearest neighbour wasn't found"
-                )
-            return data[closest_n]
+            # (n, 1) -> (n,): this path returns data rows directly, so the
+            # stencil axis must not survive into the result.
+            return data[closest_n.reshape(-1)]
 
         closest_n, n_weights, wide, degenerate = self._local_stencil(
             coords_converted, nnn, p, order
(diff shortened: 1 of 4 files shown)


lmoresi:
## Adversarial review

### The deletions are broader than "dead code" implies

`_nn_proxy` was a **public-facing keyword** on `Swarm.add_variable`, underscore-prefixed but reachable. Anything passing `_nn_proxy=` now gets a `TypeError` rather than a deprecation warning. Justified — it gated a branch inside a method with zero callers, so it could not have done anything — but it is an API removal without a shim, and the PR title says "dead surface", which undersells it. Nothing in the repo or `docs/examples` passes it; user scripts outside the repo are not checked.

Same for `meth`/`rubbish` on `MeshVariable.rbf_interpolate`. Removing `meth` **changes what the second positional argument means**: a caller writing `rbf_interpolate(coords, 4)` previously set the dead `meth` and got the default `nnn`; now it sets `nnn=4`. That is the fix, but for an out-of-repo caller it is a silent behaviour change, not an error. The only in-repo instance was the test, which is what the change repairs.

### The #443 fix is narrower than the class of bug

Moving one guard into `_local_stencil` fixes the instance. It does not establish that the two entry points agree everywhere else — I checked `nnn=1`, `nnn > n`, empty targets and `order` validation, but there is no systematic property test asserting the paths are equivalent across the parameter space. `test_interpolation_matrix_agrees_with_the_value_path` parametrises `order` and `dim`; it still does not sweep `nnn`, `p`, or `monotone`. A caller could find the next divergence the same way this one was found.

Worth stating plainly: `interpolation_matrix` ignores `monotone` entirely — correctly, since limiting is data-dependent and an operator has no data — but the signature simply omits it, so `interpolation_matrix(..., monotone=True)` is a `TypeError` while a reader might expect a no-op. Not changed here; noted.

### What the tests do and do not cover

The new guard test asserts both entry points raise, and that `order=0` stays legal on both. It does **not** assert that the *operator* and the *value* path produce the same result at `nnn=1, order=0` — the one combination where the shapes differ most (`data[closest_n.reshape(-1)]` versus a CSR with one non-zero per row). They should agree; it is untested.

The error-message test pins a string (`"20-point stencil.*holds 5 point"`). String-matching tests rot when messages are reworded. It is checking the thing that actually regressed, but a future editor will have to update it.

### What I did not do

- **No parallel run of the deleted paths.** `_get_map`/`_rbf_reduce_to_meshVar` were dead, so there is nothing to exercise, but `_nn_proxy` removal touches `SwarmVariable.__init__` and `Swarm.add_variable` signatures, which every swarm test constructs. Covered by np=2 only, not np=4.
- **No check that `xxhash` is still needed.** `_get_map` was the only user of the `import xxhash` inside it; the import was local to the deleted function, so it went with it, but I did not check whether `xxhash` remains a required dependency elsewhere or is now removable from the environment spec.

### What holds up

- The defect was found by a retrospective adversarial review of already-merged code, reproduced, filed with a reproducer, and fixed at the layer that makes it structurally impossible rather than at the call site.
- Deleting the `_get_map` trio matches a disposition already recorded and argued in `SWARM_MODERNIZATION_DESIGN_2026-07.md` (SWARM-23) rather than being an opportunistic cleanup.
- 105 lines removed against 65 added, with level_1 551 / level_2 319 / np=2 4 passing and the style gate clean.

PR #448, opened 2026-07-27
Inverse-distance weights use the actual distance; p defaults to 1 (#427)

### LINKED ISSUE #427: rbf_interpolator_local: inverse-distance weights use SQUARED distances, so the documented p semantics are wrong (effective exponent is 2p)

lmoresi:
Fixes #427.

## The defect

`KDTree.query` / `find_closest_n_points` return **squared** distances, and the weighting used them directly:

```python
weights = 1 / np.power(epsilon + distance_n, p)     # distance_n is r²
```

So the decay was `r^(-2p)`, not the `r^(-p)` the argument is named and documented for. At the old default `p=2` the scheme was inverse **fourth** power — close to nearest-neighbour — not the "inverse distance (squared)" the docstring claimed.

The retained `old_*` implementations took `sqrt` explicitly, so this changed silently at a rewrite. Nothing caught it: no test pinned the exponent, and the only accuracy test used `rtol=5e-2`, loose enough that `1/r²` and `1/r⁴` both pass.

## The change (maintainer decision)

1. Take the square root, so `p` applies to the actual distance.
2. Default `p` 2 → **1**, so the default is genuinely *inverse distance*.

`epsilon` stays `1e-12` but now floors `r` rather than `r²`, so its scale is the one it reads as — it was an effective `1e-6` length floor.

## Verification

Two sources at distance 1 and 2 from the target; interpolating a field that is 1 at the near point and 0 at the far one recovers the weight ratio directly:

| `p` | measured `w₁/w₂` | expected `2^p` |
|---|---|---|
| 1 | 2.0000 | 2 |
| 2 | 4.0000 | 4 |
| 3 | 8.0000 | 8 |

The exponent had **no test at all** before; it does now, including one pinning the default at `p=1`.

## How much it actually changes

Linear field through the swarm-proxy stencil (`nnn=dim+1`), max relative error:

| | `p=1` | `p=2` |
|---|---|---|
| 2D | 8.79e-2 | 8.77e-2 |
| 3D | 2.05e-1 | 2.04e-1 |

The choice of `p` barely matters. What mattered was that `p` meant something other than its name — and that since #430, `order=0` is the fallback rather than the primary path, so this is a correctness-of-contract fix rather than an accuracy one.

## Call sites

Stopped hard-coding `p=2`: `SwarmVariable.rbf_interpolate` lets the default apply, `MeshVariable.rbf_interpolate`'s own `p` default follows to 1, and the `nnn=1` checkpoint call drops a positional `p` it never used (`nnn=1` short-circuits before any weighting).

## Test gap closed

`test_interpolation_matrix_agrees_with_the_value_path` now sweeps `nnn` and `p` as well as `order`. It previously pinned `order` only and hard-coded `p=2` on the value side — so it compared two different weightings the moment the default changed, and duly failed here. That gap was flagged in this series' own adversarial review; this is it biting.

## Regression

- level_1 `tier_a`/`tier_b`: **567 passed, 0 failed**
- level_2 `tier_a`/`tier_b`: **319 passed, 0 failed**

### DIFF
--- a/docs/developer/subsystems/interpolation.md
+++ b/docs/developer/subsystems/interpolation.md
@@ -25,6 +25,11 @@ That is what `order` selects on
 
 $$ w_j = \frac{d_j^{-p}}{\sum_k d_k^{-p}} $$
 
+$d_j$ is the **actual** distance to the neighbour, and `p` defaults to 1 — so
+the default really is inverse distance. (Until #427 the kd-tree's squared
+distances were used directly, making the decay $r^{-2p}$: the documented
+default of `p=2` was in fact $1/r^4$.)
+
 Weights are positive and sum to one. Consequences, both of them important:
 
 - A **constant** field is reproduced exactly.
--- a/src/underworld3/ckdtree.pyx
+++ b/src/underworld3/ckdtree.pyx
@@ -434,7 +434,7 @@ cdef class KDTree:
             coords,
             data,
             nnn = None,
-            p=2,
+            p=1,
             verbose = False,
             order = 0,
             monotone = False,
@@ -476,7 +476,7 @@ cdef class KDTree:
             nearest-neighbour values without distance weighting.
         p : int, optional
             Power index for distance weighting: ``weight = 1/distance^p``
-            (default 2). Used by ``order=0`` only.
+            (default 1, i.e. inverse distance). Used by ``order=0`` only.
         verbose : bool, optional
             Print progress messages (default False).
         order : int, optional
@@ -738,13 +738,15 @@ cdef class KDTree:
                 )
             return closest_n, np.ones(closest_n.shape), None, degenerate
 
-        # can decompose weighting vecotrs as IDW is a linear relationship
-        # build normalise weight vectors and multiply that with known data
-        # TODO(BUG): issue #427 — `distance_n` holds SQUARED distances, so the
-        # decay is r^(-2p), not the documented r^(-p), and `epsilon` floors r
-        # at ~1e-6 rather than 1e-12.
+        # Inverse distance weighting: w = 1 / (eps + r)^p, normalised.
+        # `find_closest_n_points` returns SQUARED distances, so the square root
+        # is taken here -- without it the decay is r^(-2p) rather than the
+        # r^(-p) the argument names (issue #427). epsilon is a length floor for
+        # a target sitting exactly on a source point; it is on r, not r^2, so
+        # its scale is the one it reads as.
         epsilon = 1e-12
-        weights = 1 / np.power(epsilon + distance_n[:], p)
+        distance = np.sqrt(distance_n[:])
+        weights = 1 / np.power(epsilon + distance, p)
         n_weights = (weights.T / np.sum(weights, axis=1)).T
 
         if order == 0:
@@ -791,7 +793,7 @@ cdef class KDTree:
 
         return closest_n, linear_weights, wide, degenerate
 
-    def interpolation_matrix(self, coords, nnn=None, p=2, order=0):
+    def interpolation_matrix(self, coords, nnn=None, p=1, order=0):
         """Sparse operator mapping values on the KD-tree points to ``coords``.
 
         ``T @ data`` is exactly what :meth:`rbf_interpolator_local` returns for
@@ -895,8 +897,8 @@ cdef class KDTree:
         nnn : int
             The number of neighbour points to sample from. If `1`, no distance averaging is done.
         p : int
-            The power index to calculate weights, i.e., pow(distance, -p).
-            Used by ``order=0`` only.
+            The power index to calculate weights, i.e., pow(distance, -p),
+            on the actual distance. Used by ``order=0`` only.
         verbose : bool
             Print when mapping occurs
         order : int, optional
--- a/src/underworld3/discretisation/discretisation_mesh_variables.py
+++ b/src/underworld3/discretisation/discretisation_mesh_variables.py
@@ -886,7 +886,7 @@ class _BaseMeshVariable(Stateful, uw_object):
 
         return self._kdtree
 
-    def rbf_interpolate(self, new_coords, nnn=None, p=2, verbose=False,
+    def rbf_interpolate(self, new_coords, nnn=None, p=1, verbose=False,
                         order=0, monotone=False):
         """Interpolate variable data to new coordinates using RBF.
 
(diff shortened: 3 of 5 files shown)


lmoresi:
## Adversarial review

### This changes numbers for every existing `order=0` caller, and I have not measured most of them

The weighting goes from an effective `1/r⁴` to `1/r`. That is a large change in stencil sharpness — the old scheme was close to nearest-neighbour, the new one is markedly smoother. I measured it on **one** thing: a linear field through the proxy stencil, where it is a wash (8.79e-2 vs 8.77e-2).

Consumers still on `order=0` that I did **not** measure:

- `MeshVariable.rbf_interpolate` — the RBF rung of the `uw.function.evaluate` point-location fallback ladder. Points that fail location now get a smoother extrapolant. Still bounded (positive weights, convex combination), so no contract is broken, but the values move.
- `adaptivity.py:903` mesh2mesh transfer.
- The `order=1` **degenerate fallback**, which uses these weights at the handful of points where the affine fit fails.

The suites pass, but "no test asserts a number that moved" is not the same as "nothing moved". A caller relying on the old near-nearest-neighbour sharpness will see different answers with no deprecation signal.

### The measurement I chose flatters the change

A **linear** field is close to the worst possible probe for a weighting-exponent change: inverse distance smears a gradient regardless of `p`, so the error is dominated by the scheme's inability to reproduce linears, not by the exponent. That the two agree to three significant figures says little.

A field with curvature, or a discontinuity, would separate them properly — `1/r⁴` tracks a jump much more sharply than `1/r`. I did not run one. The honest statement is "`p` barely matters *for the linear-field metric I used*", not "`p` barely matters".

### `epsilon` is now arguably in the wrong place

`epsilon = 1e-12` on the actual distance is a `1e-12` length floor. For non-dimensional coordinates on an O(1) domain that is fine. For a kd-tree built in metres over a planetary domain it is still fine; for one built in units where the domain is O(1e-12) it is not. It was previously an effective `1e-6` floor, which was wrong differently. Neither version scales `epsilon` to the data, and nothing tests coincident points.

### The default change is a second decision riding on a bugfix

Taking the square root is unambiguously a bug fix. Moving `p` from 2 to 1 is a **taste decision** about default smoothing, and it is bundled into the same commit. They are separable: the fix alone (`p=2` on actual distance → `1/r²`) would have been the conservative landing. Bundled, a reviewer cannot approve one without the other, and a future bisect cannot separate them.

Maintainer chose `p=1` explicitly, so this is recorded rather than contested — but it should be visible that the PR does two things.

### What holds up

- The exponent now has a direct test, on a two-point geometry where the expected ratio is exact arithmetic rather than a tolerance. That test would have caught the original regression the moment it was introduced.
- The consistency test's `p`/`nnn` sweep — flagged as missing in this series' own earlier review — failed on this change and was fixed rather than loosened. That is the gap closing as intended.
- The call sites stopped hard-coding `p`, so the default is now expressed in exactly one place.

lmoresi:
## Closing the gap I flagged: measured on curved and discontinuous fields

The review above said the linear-field metric was the worst possible probe for an exponent change and that I had not run a better one. Running it now. `p=4` on the actual distance reproduces the **old** effective `1/r⁴`, so the last column is the previous behaviour.

Max error, `nnn=dim+1`, 4000 sources / 400 targets:

| | `p=1` (new default) | `p=2` | `p=4` (old behaviour) |
|---|---|---|---|
| 2D linear | 2.79e-2 | 2.61e-2 | 2.57e-2 |
| 2D curved (`sin 6x + \|x\|²`) | **5.73e-2** | 5.86e-2 | 6.11e-2 |
| 2D jump | 1.00 | 1.00 | 1.00 |
| 3D linear | 1.29e-1 | 1.30e-1 | 1.33e-1 |
| 3D curved | 1.69e-1 | 1.66e-1 | **1.77e-1** |
| 3D jump | **8.30e-1** | 8.90e-1 | 9.60e-1 |

So the conclusion survives contact with a better probe, and mildly favours the change:

- **Curved fields**: `p=1` is best in 2D and within 2% of best in 3D; the old `1/r⁴` is worst in both. Sharper weighting does not help a smooth field — it just narrows the effective stencil toward one point.
- **Discontinuity**: this is where I expected `1/r⁴` to win, and it does not. In 2D all three saturate at 1.0 (a target between two materials gets a value near 0.5 either way). In 3D `p=1` is clearly best, 8.30e-1 against 9.60e-1.
- **Linear**: the old behaviour is marginally *better* in 2D (2.57e-2 vs 2.79e-2) and worse in 3D. This is the flattering-metric case the review called out — the differences are noise against a scheme that cannot reproduce linears at all.

That removes the substantive objection. The remaining ones stand: this still changes numbers for `order=0` consumers I have not individually exercised (`evaluate`'s fallback rung, `adaptivity` mesh2mesh, the `order=1` degenerate fallback), `epsilon` is still an unscaled absolute length floor with no coincident-point test, and the bugfix and the default change are still bundled in one commit.

PR #450, opened 2026-07-28
Proxy mesh variables inherit the swarm variable's units (#439)

### LINKED ISSUE #439: Swarm proxy MeshVariable does not inherit the swarm variable's units
Split out of #426, which was closed by PR #434. #434 fixed the coordinate half (dimensional `.coords` into a non-dimensional kd-tree). This is the other half, deliberately left out of that PR.

## The defect

`SwarmVariable._create_proxy_variable` builds the proxy `MeshVariable` without forwarding `units=`:

```python
var = swarm.add_variable(name="rho", size=1, proxy_degree=1, units="kg/m^3")
var.units            # kilogram / meter ** 3
var._meshVar.units   # None
```

So `var.sym` — which is the proxy's symbol, and the thing that ends up in constitutive expressions — is dimensionless even though the swarm variable it stands for is not.

Confirmed on `development` after #434.

## Why it was not fixed alongside the coordinate bug

Attaching units to the proxy makes `var.sym` unit-aware, which changes symbolic behaviour everywhere a proxied swarm variable is composed into an expression — constitutive models, buoyancy terms, DDt history. That is a design decision about the units boundary, not a one-word coordinate fix, and bundling it into a bugfix PR would have made the diff unreviewable.

## What needs deciding

- Should the proxy carry the swarm variable's units, or is the proxy deliberately a non-dimensional internal object with units living only on the `SwarmVariable`?
- If the former: what happens to existing expressions that currently compose a dimensionless `var.sym`?
- `MeshVariable.data` is documented as always non-dimensional, so the storage does not change either way — this is purely about what `.units` and `.sym` advertise.

Reference: `docs/developer/design/UNITS_SIMPLIFIED_DESIGN_2025-11.md` is the governing document for the units boundary.

lmoresi:
Fixes #439.

## The defect

`SwarmVariable._create_proxy_variable` did not forward `units=` to the proxy `MeshVariable`. Since `var.sym` resolves to the proxy, reading a proxied variable through the symbolic path returned the **non-dimensional** number with no units attached, presented as though it were the answer.

Measured, with `density=3300 kg/m³` as the reference quantity and a stored non-dimensional value of `1.0`:

| | before | after |
|---|---|---|
| `var.units` | `kg/m³` | `kg/m³` |
| `var._meshVar.units` | **`None`** | `kg/m³` |
| `var.array` | 3300.0 | 3300.0 |
| `var._meshVar.array` | **1.0** | 3300.0 |
| `evaluate(var.sym)` | **`ndarray` 1.0, no units** | `UnitAwareArray` 3300.0 kg/m³ |

The two rows in bold are the bug: the swarm variable and its own proxy disagreed about the value, and `evaluate` handed back the raw non-dimensional number silently.

Stored data stays non-dimensional on both sides — only what the proxy *advertises* changes, so the transfer itself is untouched (asserted by test).

## Context

This is the half of #426 that #434 deliberately left out. It was split off at the time because attaching units to the proxy makes `var.sym` unit-aware everywhere proxies are composed into expressions, which is a units-boundary decision rather than a coordinate fix. Maintainer decision was to forward the units and measure what moves.

## Blast radius: none

- level_1 `tier_a`/`tier_b`: **571 passed** (567 baseline + the 4 new tests), 0 failed
- level_2 `tier_a`/`tier_b`: **319 passed**, 0 failed
- np=2 parallel swarm/proxy: 4 passed

**No existing test both declares units on a swarm variable and reads it symbolically.** That is why the change moves nothing — and also why the defect survived. Worth recording as a coverage gap in its own right, not just as a convenient result.

## Tests added

To `tests/test_0116_swarm_proxy_units.py` (created by #434):

- proxy `.units` equals the variable's `.units`
- `.array` on both sides reaches the same dimensional number, while `.data` stays non-dimensional on both
- `evaluate(var.sym)` returns a `UnitAwareArray` with the right units and value — the silent-wrong-value case
- a variable declaring no units keeps a dimensionless proxy

### DIFF
--- a/src/underworld3/swarm.py
+++ b/src/underworld3/swarm.py
@@ -1065,6 +1065,12 @@ class SwarmVariable(DimensionalityMixin, MathematicalMixin, Stateful, uw_object)
                 continuous=self._proxy_continuous,
                 varsymbol=r"\left<" + self.symbol + r"\right>",
                 remesh_policy="reinit",
+                # The proxy is what `var.sym` resolves to, so it advertises
+                # the same units as the variable it stands for. Without this,
+                # evaluating a proxied symbol returned the NON-DIMENSIONAL
+                # number with no units attached, as though it were the answer
+                # (issue #439). Stored data stays non-dimensional either way.
+                units=self._units,
             )
             # The remesh helper calls this on REINIT vars after an
             # adapt. Bound here so the closure captures ``self`` (the
--- a/tests/test_0116_swarm_proxy_units.py
+++ b/tests/test_0116_swarm_proxy_units.py
@@ -105,3 +105,107 @@ def test_proxy_refresh_still_works_without_units():
 
     del swarm
     del mesh
+
+
+# ---------------------------------------------------------------------------
+# The proxy advertises the swarm variable's units (issue #439).
+#
+# The proxy MeshVariable is what `var.sym` resolves to, so if it carries no
+# units then reading a proxied variable through the symbolic path silently
+# returns the NON-DIMENSIONAL number as though it were the answer. Measured
+# before the fix, with density=3300 kg/m^3 as the reference and a stored
+# value of 1.0:
+#
+#     swarm.array 3300.0   proxy.array 1.0        <- disagree
+#     evaluate(rho.sym) -> ndarray 1.0, no units
+#
+# Stored data is non-dimensional on both sides either way; only what the
+# proxy declares changes.
+# ---------------------------------------------------------------------------
+@pytest.fixture
+def dimensional_model():
+    orchestration_model = uw.get_default_model()
+    orchestration_model.set_reference_quantities(
+        length=uw.quantity(1000.0, "km"),
+        viscosity=uw.quantity(1e21, "Pa*s"),
+        density=uw.quantity(3300.0, "kg/m^3"),
+    )
+    yield orchestration_model
+    uw.reset_default_model()
+
+
+@pytest.fixture
+def dimensional_mesh(dimensional_model):
+    return UnstructuredSimplexBox(
+        minCoords=(0.0, 0.0), maxCoords=(1.0, 1.0), cellSize=1.0 / 8.0
+    )
+
+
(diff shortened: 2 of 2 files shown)


lmoresi:
## Adversarial review

### "Blast radius: none" is a weaker result than it reads

571/319/4 passing means **no existing test exercises the changed path at all**, not that the change is safe. The condition to trip it is: a swarm variable that declares `units=`, with a proxy, read symbolically, under an active units model. Nothing in the suite does all four.

So the honest claim is "nothing regressed *that is tested*". Anything downstream that composed a proxied symbol and relied on it being dimensionless — arithmetic against a bare float, a constitutive term assembled without unit checking — now sees a unit-aware operand. The suite cannot tell us whether such code exists, and the `docs/examples` tree is not run by CI.

The place I would expect trouble is multi-material constitutive assembly, where a density proxy multiplies a level-set fraction. `IndexSwarmVariable` level sets carry no units, so that product is now `unit-aware x dimensionless`. That composes fine in principle; it is not exercised.

### The type of `_units` being forwarded is not pinned

`self._units` is a `pint.Unit` by the time it reaches the constructor, not the string the user passed. `MeshVariable` accepts both — but the test asserts `proxy.units == var.units`, which would pass even if both were wrong in the same way. A test pinning the *string* (`"kilogram / meter ** 3"`) exists only in the `evaluate` test, indirectly.

### The vector/tensor case is untested

Every new test uses `size=1`. A vector swarm variable with units gets a proxy with `self.shape` components and one `units=` for all of them, which is presumably right, but "presumably" is doing work there. Same for `dtype=int` variables, where units are meaningless and now get forwarded anyway if declared.

### What holds up

- The before/after is measured on both sides of the change rather than asserted, including the `.array` disagreement that is the concrete symptom.
- The dimensionless case is pinned, so the change cannot silently start attaching units to variables that never asked for them.
- The fix is one keyword at the single site that constructs the proxy; there is no second path to keep in step.

PR #455, opened 2026-07-28
Fix GenericFluxModel flux storage — preserve symbolic field dependencies

bknight1:
## Summary

`GenericFluxModel.Parameters.flux` setter (and `__init__`) wrapped each flux component in a UWexpression via `validate_parameters()`, creating intermediate symbols (`q_0`, `q_1`) that the JIT treated as independent constants. This lost the dependency on the unknown field variable and produced a singular matrix.

## Fix

Store raw sympy expressions directly in the flux setter, bypassing the UWexpression wrapper. The `__init__` now also sets a raw `sympy.zeros` vector instead of validated components.

## Tests

New test file `tests/test_1060_generic_flux_model_update.py` with three tier_a, level_1 tests:

- **test_flux_update_via_force_setup**: Changing flux via `_force_setup=True` works (solution scales correctly with k)
- **test_flux_update_via_solver_link**: Changing flux via `_solver` back-link triggers automatic rewire (no `_force_setup` needed)
- **test_flux_stores_raw_sympy**: Verifies flux stores raw sympy, not UWexpression-wrapped symbols (no `q_0`, `q_1` artifacts)

## Verification

All three tests pass with the fix; before the fix they would produce singular matrices.

### DIFF
--- a/src/underworld3/constitutive_models.py
+++ b/src/underworld3/constitutive_models.py
@@ -2593,17 +2593,10 @@ class GenericFluxModel(Constitutive_Model):
         def __init__(inner_self, _owning_model):
             inner_self._owning_model = _owning_model
 
-            default_flux = sympy.zeros(_owning_model.dim, 1)
-            elements = [default_flux[i] for i in range(_owning_model.dim)]
-            validated = []
-            for i, v in enumerate(elements):
-                flux_component = validate_parameters(
-                    rf"q_{{{i}}}", v, f"Flux component in x_{i}", allow_number=True
-                )
-                if flux_component is not None:
-                    validated.append(flux_component)
-
-            inner_self._flux = sympy.Matrix(validated)
+            # Raw sympy zero vector — NOT wrapped in UWexpressions, so the
+            # JIT sees the symbolic content directly (field variable dependencies).
+            # See the flux setter for the rationale.
+            inner_self._flux = sympy.zeros(_owning_model.dim, 1)
 
         @property
         def flux(inner_self):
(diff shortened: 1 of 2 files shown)


lmoresi:
# Adversarial review — live probes on the PR head and its exact parent commit

**Verdict: the fix is correct and merge-worthy; the stated mechanism is wrong in a way that matters for the record.** No change to the fix itself is needed — a wording correction and one test cleanup follow in a response commit, then this merges.

## The true mechanism (probed, decisive)

Reproduced pre-fix on the PR's exact parent (`967fb30b`): Poisson + GenericFluxModel, `flux = grad(u)` → `DIVERGED_LINEAR_SOLVE` after 0 iterations, solution identically zero. Diagnostics:

| Evidence | Value (pre-fix) |
|---|---|
| `constants_manifest` | **empty** — q_0/q_1 never routed to `constants[]` |
| `_is_truly_constant(q_0)` | **False** (it contains the unknown's gradient) |
| `_G3` (dF1/d∇u) | **zero matrix** |
| Same pre-fix code, `consistent_jacobian=True` | `G3 = Identity`, solves correctly (peak 0.07365, the textbook value) |

So this is **not** "the JIT treated the components as independent constants" — the manifest was empty and the residual was never corrupted (PR #416's reveal machinery handled the wrappers correctly, which is why the Newton-mode run converges with the identical residual). The defect is in the **default Picard tangent**: `_jacobian_source` deliberately differentiates the RAW wrapped flux (frozen-coefficient semantics; unwrap-before-differentiate runs only on the Newton path). For every other model that's safe — the wrapper holds only a coefficient, and d(κ·∇u)/d∇u = κ ≠ 0. GenericFluxModel wrapped the **entire flux**, so the opaque q_i swallowed the ∇u structure and "freeze the coefficient" froze everything: G0–G3 = 0 → structurally singular operator.

This distinction matters: the current docstring/comment wording implicitly indicts the #416 constants machinery, which is exonerated, and would misdirect the next person who hits a singular matrix.

## Findings

1. **Fix works — PROBED.** All 3 new tests pass; independent probe: GenericFluxModel `q = 3∇u` vs DiffusionModel `κ = 3` agree to 0.0 relative difference; flux updates between solves work via both `_force_setup` and the `_solver`-link reset.
2. **Mechanism wording — PROBED, correction required** (above). Applied in the response commit.
3. **Units survive raw storage — PROBED.** Dimensional κ as `uw.expression` under active reference quantities: bit-identical to DiffusionModel, and κ still lands in `constants_manifest` (coefficient-level rampability retained). Nothing in the units chain depended on the whole-flux wrapper.
4. **Behavioral note — PROBED.** With raw storage, GenericFluxModel's default tangent is now effectively full Newton: probe `q = (1+u²)∇u` converges in 2 SNES iterations under the *Picard default* (there is no wrapper left to freeze, so the Picard/Newton switch is a no-op for this model). Usually a win; documented in the response commit.
5. **Collateral clean — PROBED.** Diff touches only `GenericFluxModel._Parameters`; `validate_parameters` and all other models untouched; `test_0103` (rampable constants), `test_0200`, `test_0812` (units) + the new file: 14/14 on the PR build.
6. **Test cleanups (minor).** `Parameters._solver = solver` in the test setup is redundant (the `constitutive_model` setter wires it) and masks a regression in the automatic wiring — deleted in the response commit. Tests hard-code MUMPS; house preference is plain `lu` (all review probes used it) — left as-is to keep the response commit minimal, noted here.

## The trap that remains (scoped precisely)

The general hazard is **narrower** than "wrapped field-dependent parameters break Jacobians": coefficient-level wrapping under the Picard default is *intentional* frozen-coefficient semantics and stays nonsingular. The dangerous pattern is exactly one — **a wrapped Parameter that swallows the entire ∇u dependence of a flux term** — and GenericFluxModel was the only in-tree instance. Future models (or user-built constitutive models) can re-create it; a warn-if-wrapped-sym-contains-the-unknown's-gradient guard is conceivable but out of scope here.

PR #456, opened 2026-07-28
Issue quick-fix batch 2: six small independent fixes (#415 #447 #413 #411 #451 #414)

### LINKED ISSUE #411: compute_volume_residual_fields / boundary sibling still strip Dirichlet values (the #407 fix was applied to one of three duplicated variants)

### LINKED ISSUE #413: boundary_flux: 2D P1 traces crash with a bare KeyError in the lumped line-mass assembly
Found in the PR #404 round-2 review (2026-07-24), pre-existing: the 2D lumped de-smear assembles P2 line masses (vertex-midpoint-vertex triples), so a P1 scalar/velocity trace has no midpoint keys and dies with a bare `KeyError: (x, y)` (also hit independently while building the #407 regression tests — the workaround was degree=2). With PR #404 merged, the docstring enumerates 'auto selects lumped for 2D traces', which now over-promises for P1. Fix: assemble the P1 two-node line mass when the trace has no midpoints (mirror of the 3D P1/P2 dispatch #404 added), or raise a legible NotImplementedError.

### LINKED ISSUE #414: Curved-boundary P2 VERTEX flux/traction values converge only slowly — document and steer consumers to midpoint/fitted quantities
From the PR #404 round-2 review (2026-07-24). The consistent P2 trace-mass recovery is exact on FLAT boundaries (pointwise 1e-11), but on CURVED boundaries the vertex values are unusable at practical resolutions: the P2 vertex basis has zero surface mean, so vertex reactions carry only the O(h) facet-normal/geometry error, and the consistent solve faithfully reconstructs that error. Measured (spherical-shell conduction, exact flux): vertices 93%→55% error under refinement cs=0.25→0.125 (moving SLOWLY), midpoints 2.8%→0.7% (superconvergent). The Zhong vertex-only harmonic fit drifts AWAY from the reference with refinement (0.395→0.371 vs 0.419) while the midpoint fit converges onto it.

Actions: (1) docstring caveat on boundary_normal_traction / dynamic_topography / boundary_flux — on curved boundaries, pointwise consumers should use midpoint values or integral/fitted quantities, not vertices; (2) consider having dynamic_topography on curved boundaries return a midpoint-weighted or fitted field; (3) the test_1018/test_1064 vertex goldens encode the bias (rtol 0.10-0.12 absorbs it) — annotate so nobody tightens them against the wrong reference.

### LINKED ISSUE #415: sympy.Max/Min(UWexpression, plain number) crashes at construction: AttributeError '_prec'

### LINKED ISSUE #447: Self-assigning a constitutive Parameter (read then write back) recurses to stack death
Reading a constitutive `Parameters` value and assigning it straight back is not a no-op — it nests a `UWexpression` inside itself, and the next `.value` access recurses until the stack dies.

```python
cm = stokes.constitutive_model
saved = cm.Parameters.yield_stress      # read
...
cm.Parameters.yield_stress = saved      # write back  -> RecursionError later
```

Traceback (abridged):

```
constitutive_models.py:_apply_floor      return uw.maths.smooth_max(value, floor, rounding)
maths/functions.py:smooth_max            return (a + b + sympy.sqrt((a - b) ** 2 + epsilon**2)) / 2
function/expressions.py:__add__          result_value = self.value + other.value
function/expressions.py:947 value        if hasattr(self._sym, 'value'):
  [Previous line repeated 977 more times]
RecursionError: maximum recursion depth exceeded
```

`UWexpression.value` (`expressions.py:947`) walks `self._sym` looking for a nested `.value`, and self-assignment makes `_sym` reference the expression that owns it, so the walk never terminates.

**Why it bites in practice.** Save/restore around a temporary parameter change is the natural way to probe a model without disturbing it — e.g. removing a regularisation to measure a residual against the unregularised law, then putting it back. That pattern is silently a trap. The workaround is to re-derive the value rather than restore the object, which is only possible if you know how it was built.

Two things worth considering:

- the setter could unwrap an incoming `UWexpression` that is already this parameter's own expression (or, more generally, refuse to nest an expression inside itself);
- `UWexpression.value` could carry a cycle guard so the failure is a clear error rather than a stack overflow 977 frames deep.

Related: the same class of self-referential blow-up was seen with `sympy.Max(<UWexpression>, x)` recursing to stack death during the Drucker–Prager consistent-Jacobian work, which is why `uw.maths.smooth_max` exists.

Found while measuring true hard-Min residuals during viscoplastic continuation work (PETSc 3.25, `development`).

### LINKED ISSUE #451: Spiegelman notch benchmark example builds a 3600 x 900 km domain instead of 120 x 30 km
`docs/examples/WIP/Benchmark/Ex_VP_Spiegelman_Benchmark.py` on `development` builds its gmsh geometry in **kilometres** and then hands the plex to `uw.discretisation.Mesh` unscaled. UW3 reads raw DM coordinates as **already non-dimensional**, so the model it actually solves is 30x too large in every direction.

The script's own comment says the domain is `4H x H = 120 km x 30 km`. It builds 3600 x 900 km.

### Where

```python
H = uw.quantity(30, "km")
model.set_reference_quantities(length = H, ...)
S = H.magnitude          # 30.0 - scale factor: non-dimensional -> km
...                      # every gmsh point is multiplied by S
mesh1 = uw.discretisation.Mesh(plex[1], boundaries=boundaries, ...)   # no rescale
```

### Measured

Loading the mesh both ways and printing what the solver sees:

| build | non-dim extent | dimensional |
|---|---|---|
| as written | x = -60..60, y = -30..0 | **3600 x 900 km** |
| DM coordinates divided by H first | x = -2..2, y = -1..0 | 120 x 30 km |

`eta_bg` non-dimensionalises to exactly 1 in **both** cases, so viscosity is not the tell — only the mesh extent is.

### Why it is fatal rather than cosmetic

Non-dimensionally the momentum equation has one group, `rho*g*L^2/(eta*V)`, and the yield law adds `C*L/(eta*V)`. Inflating L by 30 multiplies the first by **900** and the second by 30, so the body force overwhelms the imposed convergence and the cohesion/friction balance is shifted to a point no correctly-scaled version of this benchmark reaches.

Concretely, at the benchmark's own default parameters (`eta_bg = 1e24`, `V = 2.5 mm/yr`), taking the domain-average viscous stress `2*eta*edot` with `edot = 2V/width`:

| build | tau_y / viscous stress |
|---|---|
| as written | **136.5** — nothing yields anywhere |
| corrected | 0.188 — partially yielded |

So as shipped, the plasticity benchmark runs with **no active plasticity at its own reference parameters**.

### Fix

Divide the DM coordinates by H before constructing the Mesh, keeping `length = 30 km` as the reference:

```python
plex = uw.discretisation._from_gmsh(msh, useMultipleTags=True, useRegions=True)
_cv = plex[1].getCoordinates()
_cv.scale(1.0 / H.magnitude)
plex[1].setCoordinates(_cv)
mesh1 = uw.discretisation.Mesh(plex[1], ...)
```

Equivalently, build the gmsh geometry in non-dimensional units in the first place (drop the `* S`), which is simpler and removes the trap at source.

### Note on the misleading comment

The script says *"The Model must be set up before mesh creation so that the mesh inherits the correct coordinate units (km)"*. That is true but does not do what the comment implies: `patch_coordinate_units` only **labels** the coordinate symbols with units, it never rescales the DM. Worth correcting the comment wherever it appears, since it is the reason the bug looks safe on reading.

This was found while auditing a research harness that had copied the same mesh-construction block; the same defect is in `drivers/convergence.py` and `drivers/fmg_contraction_probe.py` in that harness, so it has propagated.

lmoresi:
## Summary

Six small, independent fixes, each with fail-first evidence where demonstrable. Closes #415, closes #447, closes #413, closes #411, closes #451, closes #414 (manual close on merge — PRs target development).

| Issue | Fix |
|---|---|
| #415 | `UWexpression.is_comparable` → `False` (plain-Symbol semantics). `sympy.Max/Min(expr, number)` previously crashed on Float internals (`_prec`); now constructs symbolically, resolves on unwrap, stays live to constant ramps. Pre-fix `AttributeError` reproduced. |
| #447 | Parameter assignment guards both self-reference cycles: read-back write is a no-op (was: container nested in itself → `RecursionError` on `.value`); composite `P.x = P.x * 2` snapshots current contents. Pre-fix `RecursionError` reproduced. |
| #413 | 2D boundary-flux recovery detects trace order from the reaction data and assembles the matching P1 two-node / P2 three-node line mass (lumped + consistent); mixed traces raise legibly. Was a bare `KeyError` on P1. |
| #411 | `compute_volume_residual_fields` + `compute_boundary_residual_fields` get the #407 `DMPlexInsertBoundaryValues` insert (with `residual_time` plumbing). Regression test pins exact agreement with `_assemble_volume_reaction` and **fails on the pre-fix build**. |
| #451 | Spiegelman notch example builds gmsh geometry in ND units (`S = 1.0`) — was 30× too large in every direction. |
| #414 | Curved-boundary warning (P2 vertices slow, midpoints superconvergent, with measured numbers) on the three consumer docstrings. |

## Validation

- New: `tests/test_0104_expression_parameter_guards.py` (3 tests); `test_1019_boundary_flux.py` +3 (P1 trace lumped+consistent, #411 exact-agreement).
- Fail-first: #415/#447 repros crash on the pre-fix build; the #411 test FAILS on the pre-fix build and passes here.
- Worktree gate `level_1 and tier_a`: **523 passed, 0 failed** (17 skipped, 1 xfail).

Adversarial review to follow on this thread before merge, per project policy.

### DIFF
--- a/docs/examples/WIP/Benchmark/Ex_VP_Spiegelman_Benchmark.py
+++ b/docs/examples/WIP/Benchmark/Ex_VP_Spiegelman_Benchmark.py
@@ -120,7 +120,7 @@ d_eta = np.log10(params.uw_eta_background.magnitude) - np.log10(params.uw_eta_ba
 # calls needed.
 #
 # The Model must be set up **before** mesh creation so that the mesh
-# inherits the correct coordinate units (km).
+# coordinates carry the model's reference length (H = 30 km per ND unit).
 
 # %%
 # Reference quantities that define the scaling (Table 1, Spiegelman et al. 2016)
--- a/docs/examples/WIP/Benchmark/Ex_VP_Spiegelman_Benchmark.py
+++ b/docs/examples/WIP/Benchmark/Ex_VP_Spiegelman_Benchmark.py
@@ -160,12 +160,15 @@ if uw.mpi.rank == 0:
 
 
 # %% jupyter={"source_hidden": true}
-### Set up the mesh — geometry in km (Table 1, Spiegelman et al. 2016)
+### Set up the mesh — geometry in units of H (Table 1, Spiegelman et al. 2016)
 #
-# The gmsh geometry is defined in non-dimensional units of H (model height),
-# then scaled to km.  Domain is 4H × H = 120 km × 30 km.
+# The gmsh geometry is defined in NON-DIMENSIONAL units of H (model height):
+# UW3 reads raw plex coordinates as already non-dimensional against the model's
+# reference length (H = 30 km), so the 4H × H = 120 km × 30 km domain is
+# [-2, 2] × [-1, 0]. Building the geometry in km here (S = H.magnitude) made
+# the solved model 30x too large in every direction — issue #451.
 
-S = H.magnitude  # 30.0 — scale factor: non-dimensional → km
+S = 1.0  # geometry is non-dimensional; reference length H supplies the km scale
 
 if problem_size <= 1:
     cl_1 = 0.25 * S
@@ -198,7 +201,7 @@ if uw.mpi.rank == 0:
     gmsh.option.setNumber("General.Verbosity", 0)
     gmsh.model.add("Notch")
 
-    # Domain outline (non-dimensional × S → km)
+    # Domain outline in non-dimensional units of H (S = 1: 4H × H → [-2,2] × [-1,0])
     Point1  = gmsh.model.geo.addPoint(-2 * S,      -1 * S,    0, cl_1)
     Point3  = gmsh.model.geo.addPoint(+2 * S,      -1 * S,    0, cl_1)
     Point4  = gmsh.model.geo.addPoint( 2 * S,    -3/4 * S,    0, cl_1)
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -2853,7 +2853,7 @@ class SolverBaseClass(uw_object):
            (measured: 93%→55% error under one refinement, while midpoints go
            2.8%→0.7%). Pointwise consumers on curved boundaries should use
            **edge-midpoint values** or integral/fitted quantities, never vertex
-           values (issue #414). Flat boundaries are pointwise exact."""
+           values (issue #414). Flat boundaries are exact up to solver tolerance."""
         from underworld3.utilities.boundary_flux import boundary_flux as _bf
         return _bf(self, boundary, mass=mass, remove_mean=remove_mean, normal=normal)
 
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -2844,7 +2844,16 @@ class SolverBaseClass(uw_object):
         traces raise explicitly. Reaction and mass assembly are partition-independent.
         For vector fluxes, supply an analytic ``normal`` when strict partition
         independence of the normal projection is required; geometric facet-normal
-        averaging at partition seams has a small pre-existing partition sensitivity."""
+        averaging at partition seams has a small pre-existing partition sensitivity.
+
+        .. warning::
+           On CURVED boundaries, P2 **vertex** values converge only slowly: the P2
+           vertex basis has zero surface mean, so vertex reactions carry only the
+           O(h) facet-geometry error, which the recovery faithfully reconstructs
+           (measured: 93%→55% error under one refinement, while midpoints go
+           2.8%→0.7%). Pointwise consumers on curved boundaries should use
+           **edge-midpoint values** or integral/fitted quantities, never vertex
+           values (issue #414). Flat boundaries are pointwise exact."""
         from underworld3.utilities.boundary_flux import boundary_flux as _bf
         return _bf(self, boundary, mass=mass, remove_mean=remove_mean, normal=normal)
 
@@ -5947,7 +5956,14 @@ class SNES_Stokes_SaddlePt(SolverBaseClass):
         triangles, and the required consistent surface-mass solve for 3D P2 triangles.
         Explicit ``"lumped"`` and ``"consistent"`` choices remain available where
         mathematically valid. Three-dimensional recovery currently supports triangular
-        P1/P2 traces only."""
+        P1/P2 traces only.
+
+        .. warning::
+           On CURVED boundaries, P2 vertex values of :math:`\sigma_{nn}` converge
+           only slowly (the vertex basis has zero surface mean, so vertex reactions
+           carry only the O(h) facet-geometry error); edge-midpoint values are
+           superconvergent. Pointwise consumers on curved boundaries should use
+           midpoint or integral/fitted quantities (issue #414)."""
         if self._rotated_freeslip_info is None:
             raise RuntimeError(
                 "boundary_normal_traction requires a completed rotated-free-slip solve.")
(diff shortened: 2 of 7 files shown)


lmoresi:
# Adversarial review — live probes on the batch worktree

**Verdict: APPROVE with one scope caveat, addressed in a response commit.** All six changes hold under attack. The caveat: the #447 guard as submitted covers the two direct cycles but not a self-reference nested *inside a wrapper expression* — and that hole's worst mode is silent garbage, not a crash. Hardened in the response commit (recursive cycle detection → legible `ValueError`), so #447 closes fully. One pre-existing (not-a-regression) P3 defect discovered en route is filed separately.

## Findings

**MAJOR (scope) — deeper #447 cycle was unguarded — PROBED, fixed in response commit.** `wrapper = uw.expression("w", Params.yield_stress * 2); Params.yield_stress = wrapper` was accepted; afterwards `constitutive_model.flux` → `RecursionError` (the original crash site) and `unwrap_expression(...)` → **silently returns 2⁵⁰·τ_y** — the cycle unrolled ~50 times against an internal depth cap and handed back as a live expression. The response commit adds recursive containment detection through nested `UWexpression` contents (visited-set bounded): any assignment that would create a cycle through wrappers now raises a `ValueError` telling the user to snapshot explicitly. Cross-container cycles (a→b→a) are caught by the same walk.

**MINOR — Matrix-valued composite snapshot produced malformed sympy — PROBED, fixed in response commit.** `xreplace` of a Matrix-valued container inside a Matrix entry stored nested-matrix garbage silently. Now raises for Matrix-shaped contents instead of substituting.

**NOTE — pre-existing P3 silent-wrong boundary flux — PROBED both sides, filed as a separate issue.** A degree-3 trace returns finite but wrong flux (−0.57…−0.74 vs exact −1.0) — byte-identical on `origin/development`: both P3 edge DOFs collapse onto the midpoint coordinate key and one reaction overwrites the other, on the old code and the new. Not a regression from this batch; the recovery should reject degree > 2 traces explicitly.

**NOTE — doc touch-ups, fixed in response commit:** stale "km" comment in the Spiegelman example (line ~123) contradicting the corrected framing; "Flat boundaries are pointwise exact" softened to "exact up to solver tolerance" (the tests assert 1e-3).

## Attacks that failed (evidence)

- **#415 blast radius (the big risk)**: Poisson with `diffusivity = sympy.Max(k, 0.5)` SOLVED and ramped k = 1.0 → 0.1 → 2.0 across the threshold *without rebuild*: T_max = 0.125001 / 0.250001 / 0.062501 vs analytic 1/(8·max(k,0.5)), matching fresh solvers to ≤3e-6. `Piecewise` conditions resolve live; `default_sort_key`/hash fine; **full `srepr` byte-identical across two fresh processes** (parallel JIT codegen determinism holds); `uw.function.evaluate` of `Max(k, ·)` correct before and after ramp.
- **#447 guards that ARE in the submitted diff**: read-back no-op skips `_reset` correctly while all other branches still reset (instance-level spy); cross-parameter alias (`P.a = P.b`) stays a live reference (no false no-op — 2.5 → 4.0 after updating the source); lazy mixed composite `P.x = P.x*2 + P.y*3` gives exactly the intended semantics (self snapshotted, other container live).
- **#413/#411 at np2**: test_1019 8 passed (both new P1 mass modes + the #411 exact-agreement test hold per-rank); parallel boundary-flux suite 3 passed. `time=0.5` path matches `_assemble_volume_reaction(time=0.5)` to rtol 1e-12; Nitsche boundary-residual path runs clean with the new insert.
- **#451 reading sweep**: every geometry entity scales with S (arcs, centres, mesh sizes); no residual km assumptions (viz windows derive from mesh bounds; the symbolic lithostatic pressure rescales with the frame — and was itself 30× wrong before this fix, consistently repaired).
- Mixed-trace raise is rank-symmetric by construction (allgathered inputs).

Issue #457, opened 2026-07-28
TransverseIsotropicFlowModel: consistent (Newton) tangent inconsistent with the residual when anisotropy is active


lmoresi:
## Evidence (PETSc `-snes_test_jacobian`, native essential-BC path — no rotated machinery involved)

Annulus, power-law background (`eta_0 = eps_II^(1/3-1)`), localised weak-plane band (`eta_1 = eta_0*(1-0.99*band)`, radial director), `consistent_jacobian=True`. Checker run at **bounded curvature** (regularisation `1e-4` inside the sqrt) so the finite-difference reference is valid:

| state | ViscousFlowModel (control) | TI isotropic limit | TI with anisotropy |
|---|---|---|---|
| rest / first iterate | 1.4e-10 | 1.4e-10 | 1.4e-10 |
| early | 9.2e-7 | 9.2e-7 | 9.2e-7 |
| developed flow | **4–5e-6** (FD-limited) | — | **1.8e-3 → 4.5e-3** |

The isotropic power-law tangent is correct (FD-limited). The defect appears only when `eta_1 != eta_0` — the anisotropy part of the consistent tangent (the `dC/d(eps_II)` director-coupled terms, `C(eps_II) = eta_0*I + (eta_0-eta_1)*[n-projector terms]`) is missing or wrong. The band is localised, so the global Frobenius mismatch is small (~0.2–0.5% of ||J||) while being locally O(1) — which matches the observed behaviour: TI solves under `consistent_jacobian=True` converge at Picard pace (24 iterations on the #438 TI smoke vs ~11 for the isotropic power-law), and early trajectories are near-identical to the frozen tangent.

## Methodological note (matters for anyone re-running this)

With the customary `1e-12` strain-rate regularisation the checker is **invalid** for shear-thinning rheology: the viscosity curvature at the floor (`~eps_II^(-8/3)`) breaks the FD reference, and every model — including the known-good isotropic Newton tangent — reads `||J-Jfd||/||J|| ~ 0.013–0.17` identically. Bound the curvature (regularisation ~`1e-4` in the sqrt) before trusting the numbers.

## Impact

- TI + `consistent_jacobian=True` silently runs at (roughly) Picard convergence rates; no wrong answers observed (the residual is the true one), just lost Newton performance exactly where fault/anisotropy physics needs it.
- Blocks trusting Newton-rate claims for TI in the free-surface/fault programme (workstream C inherits TI rheology in 3D).
[...]

Issue #403, opened 2026-07-23
Finish rotated free-slip prescribed-normal datum: wire FreeSurface consistent solve, pyx API, parallel, nonlinear


lmoresi:
## Context

The rotated strong free-slip constraint now supports a prescribed non-zero wall-normal
datum `u.n = ũ_n` on the LINEAR solve path (commit `1a08d398` on
`feature/free-surface-tangent`): one rotated matrix, a switchable datum (0 = free-slip /
held, ũ_n = prescribed radial), differing only in the constraint RHS. Validated serial
in `tests/test_1018_rotated_freeslip.py` (datum-zero matches free-slip to round-off;
`u.n = cos θ` imposed to machine precision).

This lets the free-surface **consistent** solve drop its ill-conditioned penalty natural
BC (`penalty=1e5·(n̂·v − ũ_n)·n̂`) — the term that trips `newtonls` line search on an
otherwise linear solve — in favour of the strong rotated constraint.

## Remaining work

1. **Wire the FreeSurface manager.** In `free_surface.py::_build_consistent`, register
   rotated free-slip on the surface and set `consistent._rotated_freeslip_datum =
   {surface: un_target.sym}` (updated each step in `_solve_consistent`), removing the
   `add_natural_bc` penalty. Confirm the material-surface advection is unchanged /
   improved (no penalty leak).

2. **Clean pyx API.** Let `add_rotated_freeslip_bc(conds, boundary, normal=...)` accept a
   non-zero `conds` (constant / sympy expr / field `.sym`) and store it, replacing the
   private `solver._rotated_freeslip_datum` route. Needs a Cython rebuild. Update
   `_classify_walls` (currently unpacks 2-tuples) if the stored form changes.

3. **Parallel np2/np4.** The datum application (`_set_rows_local`, the xhat ownership
   loop, `build_rotation` per-node datum eval) is written ownership-safe but untested in
   parallel. Add an np2 case to the datum test.

4. **Nonlinear path.** `solve_rotated_freeslip_nonlinear` currently raises on a datum.
   Extend it (the datum enters the constrained residual / initial-feasible guess) if a
   prescribed normal is ever needed with a nonlinear rheology.

5. **CBF σ_nn regression.** datum=0 is bit-path-identical (so σ_nn recovery is safe), but
   add an explicit test that `boundary_normal_traction` is unchanged with the new code.

## Origin

Found while building a fresh annulus free-surface convection driver and tracing why a
[...]

PR #458, opened 2026-07-28
Rotated free-slip: prescribed-normal datum through the SNES path; one unified Newton loop (#438, #403)

### LINKED ISSUE #438: Free surface with nonlinear / anisotropic rheology: strong material-boundary constraint is linear-only (penalty fallback leaks)
## Why this matters

The rotated-BC free surface (uw.systems.FreeSurface, PR #435) is the only working
free-surface strategy, and nearly every target application is nonlinear — with
anisotropic nonlinearity (TI faults) required for the flagship cases, potentially WITH
the free surface (faults reaching the surface). This must not be a blocker, and it is
not: everything runs today. But one leg runs on a fallback whose deficiency we have
now measured precisely, so it needs to be first-class work, recorded here.

## Current compatibility matrix

| combination | status |
|---|---|
| rotated free-slip (u.n=0) + nonlinear | WORKS (SNES-integrated rotated path, honours consistent_jacobian) |
| rotated free-slip + anisotropic (TI) | works at constraint level (rotation is rheology-blind); solver-config caveats only (Schur/mass pmat choice for TI) |
| FreeSurface free + held solves + nonlinear | WORKS (held auto-dispatches nonlinear rotated; sigma_nn recovery is constitutive-agnostic) |
| FreeSurface consistent solve (u.n = u~_n != 0) + nonlinear | **GAP: linear-only.** Nonlinear raises NotImplementedError with a clear message; fallback = consistent_constraint="penalty" |
| FMG hierarchy + nonlinear | works by construction (Galerkin RAP of whatever tangent is assembled; re-formed each Newton iteration; performance-only caveats for localised viscosity) |

## Why the penalty fallback is not acceptable long-term

The consistent solve exists to keep the surface a MATERIAL boundary. Measured on the
isoviscous annulus (PR #435): the penalty leaks ~4% net volume flux through the surface
and misses the prescribed rate by 6.4x more than the strong datum — the exact mechanism
that put semi-Lagrangian departure feet outside the domain and injected hot material
into the cold boundary layer. Nonlinear runs on the fallback inherit that class of
error at whatever amplitude their regime produces.

## Required work

1. **Extend the rotated prescribed-normal datum to the SNES path** (#403 item 4): the
   linear path imposes the datum via zeroRowsColumns(x,b) with a solution-row restore;
   the Newton loop needs the equivalent per iteration — set the rotated-normal DOF of
   the iterate to the datum, zero the corresponding residual row, keep the tangent
   transparent (the machinery the u.n=0 nonlinear path already has, generalised to a
   non-zero, field-valued datum). Belongs with the pyx add_rotated_freeslip_bc datum
   argument (#403 item 2) and, ideally, with the planned linear/nonlinear rotated-path
   unification (the split itself is a design smell — flagged for the joint pyx session).
2. **Anisotropic extras** (small, config-level): a sensible scalar viscosity proxy for
   the Schur mass pmat under TI in the rotated solves; never-ksponly guidance carried
   over to the FreeSurface-derived solves.
3. **Fault-at-the-surface interaction** (open question, not solver work): the surface
   ring filter and tangential transport act along a ring that a TI weak zone may
   intersect; whether the filter should be fault-aware is a physics/design question to
   revisit once 1 exists and a fault-to-surface model runs.

## Acceptance

Power-law annulus free-surface convection with consistent_constraint="strong" running
through the SNES path: material-boundary error and net surface flux at the strong-datum
level (not the penalty level) at matched steps; a TI fault-bearing smoke test compiling
and converging with the FreeSurface manager.

Related: #403 (items 2, 4), #435 (the 2D linear-complete toolkit), #423, #404 (3D).

lmoresi:
## What this is

Workstream B of the free-surface roadmap (joint session with @lmoresi): the prescribed wall-normal datum `u·n̂ = ũ_n` now runs through the full nonlinear rotated machinery, and the separate linear/nonlinear rotated solve paths are unified into one Newton/Picard loop. Closes #438's acceptance checklist and #403 items 2 & 4.

## Design

- **Datum through Newton = feasible iterates.** The constraint is affine, so accepted iterates carry `v_n = ũ_n` exactly (an affine snap in the rotated frame) and every increment satisfies the homogeneous constraint `n̂·δ = 0` — the datum never touches the tangent, the increment RHS, the FMG prolongation, or the null-space handling.
- **Cold starts take the affine lift, not the snap.** Snapping a zero state onto a datum manufactures a boundary-strip strain state whose shear-thinning tangent produces a first Newton step ~4 orders too large (measured; no line-searched step descends). The first increment instead carries the datum jump via `zeroRowsColumns(x̂, b̂)` at the rest-state tangent — which is exactly the old linear path's operation sequence, and is what makes unification natural:
- **One path.** A linear model converges after that first increment; the loop self-terminates, so the up-front `_residual_is_nonlinear` probe leaves the dispatch (two Jacobian assemblies saved per linear rotated solve; the probe survives as a lazy guard in the `picard`+pure-Newton corner). Direct LU becomes a per-increment diagnostic option. One result-dict shape (`reaction = F(u)`; the linear-only `A`/`b` keys retire). Linear solutions change at solver-tolerance level only (bit-preservation deliberately relaxed — maintainer decision in-session).
- **API**: `add_rotated_freeslip_bc(conds, boundary, normal=)` accepts the non-zero scalar datum (value-first, per the Charter); the FreeSurface manager uses it and its penalty-fallback catch is gone — `consistent_constraint="strong"` works for nonlinear rheologies.

## Three latent bugs found and fixed (each was masking the next)

1. **Rank-local datum branching** — both paths branched on the rank-local `datum_map`; a rank owning no datum boundary nodes desyncs the collective sequences (np>1 deadlock class; latent in the linear path since the datum landed). The activity flag now rides a collective PETSc Vec norm of the lift vector.
2. **Essential-pinned rotation modes admitted to the null space** — `_mode_satisfies_constraints` checked only the rotated constraint rows, so a rigid-rotation mode pinned by an essential BC on *another* boundary was projected out of every increment RHS: Newton converged superlinearly to rel ~2e-5 and floored there, irreducibly. Modes are now also verified as null vectors of the assembled operator (`‖J·m‖` against the velocity diagonal scale); after the fix the same solve reaches rel 2e-13 in 11 iterations and correctly leaves the gauge in place.
3. **Step-norm false convergence** — a tiny Newton step was reported as converged even at rel |F̂| = 0.85 (stiff tangents make tiny steps far from the solution). The exit is now verified against the reference scale `max(r0, ‖F̂(0)‖)`, which also fixes the warm-start rtol trap (rtol relative to a good warm start's own small r0 demands ever-more absolute accuracy).

Plus: FreeSurface's consistent solve warm-starts from the free solve's converged fields (a cold start at the power-law regularisation floor stalls at rel 0.13 while the free solve converges beside it).

## Evidence

Acceptance runs in `~/+Simulations/FreeSurface/annulus_fs_convection/b_snes_datum_acceptance/` (README carries the full table and attempt history):

- Power-law annulus FS convection, `consistent_constraint="strong"`, matched steps: material-boundary error **5.3e-3** (strong-linear reference 7.8e-3; penalty 3.7e-2), net surface flux **1.4e-4** (penalty 2.0e-3).
- TI fault-bearing smoke (power-law background, weak-plane band ratio 0.01, cos θ datum): converges, 24 its to rel 6e-8, σ_nn recovery clean.
- Nonlinear cos θ datum imposed to 3e-16 through genuine Newton iteration (new regression in test_1018); nonlinear datum at np2/np4 (new test in test_1066).

Tests: `test_1018` 18/18 serial · `test_1066` np2 + np4 · `test_1070` serial + np2 (7) · `test_1071` serial.

## Known limitations / follow-ups (all recorded)

- **Deformed-ring datum compatibility floor** (rel ~2e-3, honestly reported unconverged): confirmed by block-split to be 100% pressure-row residual — the arc-length (trapezoid) demean is not FE-quadrature flux-free on a deformed ring; the momentum equations are at machine precision. Fix = trace-mass-weighted demean (also the 2D instance of workstream C's datum gauge). Surface metrics unaffected.
- **#457** (found during acceptance): the TI consistent tangent is inconsistent with its residual when anisotropy is active — native-path defect, rotated machinery exonerated; TI "Newton" runs at Picard pace. Separate constitutive-model work.
- The naive LU pressure pin stays serial-only (`TODO(BUG)` retained); `_eval_datum` inherits the #432 3D point-evaluation exposure; np4 is suggestive rather than a guaranteed empty-rank exercise (annulus partitions give every rank boundary).

New subsystem documentation: `docs/developer/subsystems/rotated-freeslip.md`; changelog entry included.

An adversarial review follows as the first comment, per house practice. Draft until @lmoresi signs off.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -326,7 +326,7 @@ The PETSc-based solvers are carefully optimized and validated. **NO CHANGES with
 
 ## Boundary Conditions: Free-slip
 
-**Prefer rotated strong free-slip** (`solver.add_rotated_freeslip_bc(conds, boundary, normal=None)`, value-first: `conds=0` is the only implemented datum)
+**Prefer rotated strong free-slip** (`solver.add_rotated_freeslip_bc(conds, boundary, normal=None)`, value-first: `conds=0` is free-slip; a non-zero scalar/expression `conds` prescribes the wall-normal datum `u·n̂ = ũ_n` strongly)
 to impose `v·n̂ = 0`:
 
 - Enforces zero wall-normal flow to **machine precision** (Nitsche / penalty leak ~1e-3).
--- a/docs/developer/CHANGELOG.md
+++ b/docs/developer/CHANGELOG.md
@@ -6,6 +6,55 @@ This log tracks significant development work at a conceptual level, suitable for
 
 ## 2026 Q3 (July – September)
 
+### One Rotated Free-Slip Path, Now With a Prescribed Wall-Normal Velocity (July 2026)
+
+**Rotated strong free-slip now takes a prescribed wall-normal velocity datum
+(`add_rotated_freeslip_bc(conds, boundary, ...)` with non-zero `conds`) through
+the full nonlinear Newton machinery, and the separate linear and nonlinear
+solve paths have been unified into one** (#438; #403 items 2 and 4).
+
+The rotated constraint `u·n̂ = ũ_n` is the primitive behind both the held free-slip
+lid and the free-surface material-boundary condition — they differ only in the
+constraint right-hand side. Previously the non-zero datum existed only on a
+linear one-shot path, so a power-law or anisotropic rheology silently fell back
+to a weak penalty (which leaks worst exactly where anisotropy makes it matter).
+Now every rotated solve runs a single Newton/Picard loop in which accepted
+iterates carry the datum exactly; a cold start imposes it through the first
+increment's affine lift at the rest-state tangent (snapping a zero state onto a
+datum creates a boundary strain state a shear-thinning tangent cannot recover
+from — measured, not assumed); a linear model simply converges after that first
+increment, so the up-front nonlinearity probe is gone from the dispatch and
+every linear rotated solve saves two Jacobian assemblies.
+
+Three latent solver defects were found and fixed by making the loop report
+honestly along the way:
+
+- Branching on rank-local datum bookkeeping desynchronised the ranks'
+  collective sequences (an np>1 deadlock class); the datum-activity decision is
+  now a collective PETSc reduction.
+- A rigid-rotation mode pinned by an essential condition on *another* boundary
+  could still enter the solver null space (only the rotated rows were checked),
+  silently projecting an irreducible component out of every increment — Newton
+  converged superlinearly and then floored, far above tolerance. Candidate
+  modes are now verified as null vectors of the assembled operator, which
+  catches any form of pinning.
+- A tiny Newton step was reported as convergence even when the residual was
+  still large (a stiff tangent also produces tiny steps); the step-norm exit is
+  now verified against the problem's rest-state residual scale, which is also
+  the convergence reference for warm starts (rtol relative to a good warm
+  start's own small initial residual demands ever-more absolute accuracy).
+
+The free-surface manager's `consistent_constraint="strong"` therefore works for
+nonlinear rheologies: the penalty fallback is removed, and the consistent solve
+warm-starts from the free solve's converged fields. Acceptance: power-law
+annulus free-surface convection holds the material boundary at the strong-datum
+level (5e-3, versus 4e-2 for the penalty) with net surface flux 1e-4; a
+transversely isotropic fault-bearing smoke test converges through the same
+path. Direct LU per Newton increment remains available as a serial,
+preconditioner-free diagnostic (`solver._rotated_use_lu`).
+
+New subsystem documentation: `subsystems/rotated-freeslip.md`.
+
 ### Local Interpolation That Reproduces Linear Fields (July 2026)
 
 **The local scattered-point interpolator now has a linear-reproduction
--- a/docs/developer/index.md
+++ b/docs/developer/index.md
@@ -36,6 +36,7 @@ same topic are reference or historical material subordinate to the governing doc
 | Coding style & API conventions | [UW3 Style Charter](UW3_STYLE_CHARTER.md) (detailed reference: [Style Guide](UW3_Style_and_Patterns_Guide.md)) |
 | Data access | [subsystems/data-access.md](subsystems/data-access.md) (internals reference: [NDArray System](UW3_Developers_NDArrays.md)) |
 | Local scattered-point interpolation | [subsystems/interpolation.md](subsystems/interpolation.md) |
+| Rotated free-slip & wall-normal datum | [subsystems/rotated-freeslip.md](subsystems/rotated-freeslip.md) |
 | Units | [design/UNITS_SIMPLIFIED_DESIGN_2025-11.md](design/UNITS_SIMPLIFIED_DESIGN_2025-11.md) |
 | Testing tiers | [TESTING-RELIABILITY-SYSTEM.md](TESTING-RELIABILITY-SYSTEM.md) |
 | Branching & releases | [guides/branching-strategy.md](guides/branching-strategy.md) |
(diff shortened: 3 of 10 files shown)


lmoresi:
## Adversarial review (pre-ready, per house practice)

Ordered by how much each item should worry a reviewer.

**1. The unconditional lift-step acceptance has no safety net.** On a cold start with a datum, iteration 0 is accepted without any line search. For the tested rheologies the lift lands on a smooth state, but a rheology whose rest-state tangent is a *bad* global model (e.g. strongly pressure-dependent viscosity, or yield already active at rest) could get a first iterate far uphill, and the loop would then need the line search to walk back from a worse place than zero. Mitigations: bounded by `max_it` + stall exit + honest reporting; no observed failure. But the reviewer should know this is argued-safe, not proven-safe.

**2. `converged=True` semantics changed twice in one PR.** The reference scale is now `max(r0, ‖F̂(0)‖)` and the step-norm exit verifies against it. This is *more* honest, but it means: (a) a warm-started solve can report converged at an absolute residual that is `rtol·‖F̂(0)‖` even if its own start was much better — by design, but it is a contract change; (b) any downstream code that compared `rnorm/rnorm0` to `rtol` will see different verdicts. I audited the in-repo consumers (`_capture_rotated_report`, FreeSurface) — clean — but out-of-repo scripts reading `_rotated_freeslip_info` will see behaviour change.

**3. Linear rotated solves are no longer bit-identical to `development`.** Deliberate (maintainer ruling in-session: within-tolerance is the acceptance), and test_1066's hardcoded ∫v·v reference still passes at 1e-6 — but that pass is *luck of magnitude* for the nullspace fix (the spurious projection's imprint on the linear isoviscous case happened to be < 1e-6). A different fixture could have needed a re-baseline; anyone bisecting a downstream tolerance-level change will land on this PR.

**4. The occasional polish increment is a real cost on linear solves.** When the first residual check sits at the tolerance boundary the loop pays a second (cheap, but nonzero) KSP solve. Measured: FS plume wall-time unchanged, test_1018 *faster* overall (probe removal wins). But a user with many marginal linear solves could see it; the mitigation (inner KSP slightly tighter than outer) was considered and deferred.

**5. The nullspace mode verification depends on an assembled J.** The build was moved inside the loop to guarantee that, but `_finalize_rotated_solution`'s gauge-removal check can run with an unassembled J on one exit path (trivially-converged warm start before any Jacobian assembly this call) — it then skips the operator test (treat-as-pass, matching old behaviour). In practice J persists from a previous solve; still, that path is weaker than the main one.

**6. The FreeSurface warm-start copies fields via `.array` every step.** Correct and cheap relative to the solves, but it assumes the free solve converged — if the free solve diverged, the consistent solve now inherits its garbage rather than starting clean. The manager already warns on free-solve divergence; the coupling is new though.

**7. Parallel coverage is honest but not exhaustive.** np2/np4 pass for both datum tests; np4 on an annulus still gives every rank a boundary segment, so the true empty-datum-rank path (a 3D cap datum at high rank counts) is exercised only by construction (collective flags), not by test. The `_eval_datum` empty-rank + #432 3D exposure is documented, unresolved.

**8. Deleted code / behaviour surface.** The linear one-shot's removal takes with it: the `Uhat` result key (nobody consumed it), the internal nsp/custom_Pl self-build in `_solve_rotated_iterative` (callers audited: one), and the `A`/`b` fallback in `boundary_normal_traction` (dead by construction). If an external script called `solve_rotated_freeslip_nonlinear` by name, it breaks loudly (name gone) — accepted, the pyx dispatch is the only in-repo caller.

**9. What we could not test.** The 3D spherical datum (no 3D FreeSurface yet — the guard stands); TI Newton-rate claims (blocked by #457, found during this work); behaviour under `_rotated_use_lu` at np>1 (still guarded by the naive-pin TODO(BUG), serial-only as before).

Net: the riskiest items are (1) and (2), both semantic rather than correctness regressions, both documented in the subsystem page. The three fixed bugs each have a regression guard (deadlock class → np2/np4 tests; nullspace floor → the 1018 nonlinear datum test converges to tolerance, which the floor made impossible; false convergence → the TI smoke asserts the honest verdict).

Issue #470, opened 2026-07-29
Stale rotated free-slip datum guard tests in test_0641 (nonzero datum is now implemented, #458)


lmoresi:
`tests/test_0641_wave_c_api_shims.py::TestRotatedFreeslipValueFirst` still asserts that a non-zero rotated free-slip datum raises `NotImplementedError`:

- `test_nonzero_datum_not_implemented`
- `test_symbolic_possibly_nonzero_datum_not_implemented`

#458 (merged 2026-07-29) implemented the prescribed-normal datum, so `add_rotated_freeslip_bc(1.0, "Top")` is now accepted and both tests fail on `development`.

```
FAILED tests/test_0641_wave_c_api_shims.py::TestRotatedFreeslipValueFirst::test_nonzero_datum_not_implemented
FAILED tests/test_0641_wave_c_api_shims.py::TestRotatedFreeslipValueFirst::test_symbolic_possibly_nonzero_datum_not_implemented
2 failed, 82 passed
```

They should assert the new contract instead: a scalar/symbolic datum is accepted and recorded on `_rotated_freeslip_bcs`, and whatever is still genuinely rejected (vector/matrix values) still raises.

Found while validating the FMG option-bundle unification (#468 / #467); unrelated to that change and not fixed there to keep the diff scoped.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
Fixed by #487, merged as `5c904836`. `level_1 and tier_a` is now 530 passed, 0 failed — these two were the only red tests on `development`.

Replaced with the contract that actually holds, read off live behaviour: a numeric datum is recorded, a symbolic one is kept (`is_zero` is True only for a *provable* zero), a provable zero records no datum, and a vector value is still a `TypeError`. The last two are additions rather than replacements, so "a datum is accepted now" cannot be over-read as "anything is accepted".

Issue #276, opened 2026-06-23
FMG (preconditioner="fmg"/"auto") fails with PETSc err62 (injection) for scalar solvers on a refined Plex hierarchy


lmoresi:
## Summary

`preconditioner="fmg"` (and `"auto"` when a refinement hierarchy is present) on a
**scalar** solver (`Poisson`, `Projection`, `AdvDiffusionSLCN`) built over an
`Annulus(refinement>=1)` hierarchy fails at solve time with

```
PETSc Error 62: Could not locate matching functional for injection
```

The **Stokes velocity-block** FMG works on the *same* hierarchy, so this is
specific to the scalar (single-field) FMG path.

## Reproducer (serial)

```python
import underworld3 as uw, sympy
mesh = uw.meshing.Annulus(radiusOuter=1.0, radiusInner=0.5, cellSize=1/6,
                          refinement=2, qdegree=3)
print("dm_hierarchy levels:", len(mesh.dm_hierarchy))   # -> 3

# SCALAR: fmg FAILS, gamg OK
T = uw.discretisation.MeshVariable("T", mesh, 1, degree=3)
p = uw.systems.Poisson(mesh, T)
p.constitutive_model = uw.constitutive_models.DiffusionModel
p.constitutive_model.Parameters.diffusivity = 1.0
p.add_dirichlet_bc(1.0, "Lower"); p.add_dirichlet_bc(0.0, "Upper")
p.preconditioner = "fmg"      # -> PETSc err 62 ;  "gamg" -> OK
p.solve()
```

Observed:

```
scalar Poisson  preconditioner='gamg':  OK
scalar Poisson  preconditioner='fmg':   FAIL -> PETSc Error 62
Stokes velocity preconditioner='fmg':   OK      (same mesh)
```

## PETSc stack

```
DMPlexComputeInjectorFEM()  plexfem.c:3706
DMCreateInjection_Plex()    plex.c:10343
DMCreateInjection()         dm.c:1362
PCSetUp_MG()                mg.c:1024
PCSetUp() / KSPSetUp() / KSPSolve() / SNESSolve()
```

`PCSetUp_MG` requests an **injection** operator for the scalar field and
`DMPlexComputeInjectorFEM` cannot build it for the **degree-3** field on the
refined, boundary-snapped Plex levels (no matching dual functionals between the
coarse and fine spaces). The Stokes velocity FMG path does not request this
injection, hence it succeeds.

## Workaround

[...]

PR #471, opened 2026-07-29
One owner for the geometric-MG option bundle; rotated free-slip picks up a mesh-owned hierarchy (#468, #467)

### LINKED ISSUE #467: Rotated free-slip silently ignores a mesh-owned custom-FMG hierarchy (adapt children fall back to GAMG)
A `mesh.adapt()` refinement child carries its static coarse tail as `mesh._custom_mg_coarse_meshes`. On the standard solve path, `auto_inject_custom_mg` turns that into a registered `solver._custom_mg` so **every** solver on an adapted mesh drives geometric MG with no per-solver call. Under rotated free-slip that never happens, and the solve silently falls back to GAMG.

## Measured

`velocity_pc` as reported by `solver._rotated_freeslip_info` (annulus, rotated free-slip on `Upper`):

| configuration | velocity_pc |
|---|---|
| no hierarchy at all | `GAMG` |
| **mesh-owned tail (what `adapt()` leaves)** | **`GAMG`** |
| explicit `set_custom_fmg` | `custom-FMG` |

The mesh-owned case is indistinguishable from having no hierarchy at all.

Repro: `~/+Simulations/rotated_pmat_audit/mesh_owned_fmg_pickup.py`.

## Cause

Two facts that only bite in combination:

- `auto_inject_custom_mg(self, field_id=0)` is invoked from the standard Stokes solve at `petsc_generic_snes_solvers.pyx:8825` — **after** the rotated free-slip dispatch at `:8682` has already early-returned into `rotated_bc.solve_rotated_freeslip`.
- `rotated_bc._build_rotated_custom_Pl` consults only `solver._custom_mg`. It never looks at `solver.mesh._custom_mg_coarse_meshes`.

So the auto-pickup that exists precisely to make adapted meshes get FMG for free is unreachable from the one BC that most needs geometric MG.

## Why it matters

This is the `adapt-on-top-faults` workflow's own configuration: a fault resolved by local refinement on an `adapt()` child, with rotated strong free-slip (chosen over Nitsche because it composes with transverse isotropy). That combination is exactly the one that loses its multigrid, and it loses it silently — the solve converges, just on a much weaker preconditioner.

The `adapt-on-top-faults` skill documents the *plain refined `Annulus`* case ("native FMG is ignored; to get FMG you must build a coarse-mesh tail and call `set_custom_fmg`"). The adapt-child case is worse and is not documented: the hierarchy genuinely exists and is discarded.

## Suggested fix

Not a reordering — the injection is late on the standard path because the velocity sub-PC is only reachable once the monolithic Jacobian is assembled, and the rotated path builds its own IS-based fieldsplit inside `rotated_bc` anyway.

Instead, `_build_rotated_custom_Pl` should fall back to the mesh-owned tail when no solver-set hierarchy is registered: build a `CustomMGHierarchy([*coarse, solver.mesh], field_id=0)` itself, with the same opportunistic `try/except` and barycentric→RBF builder fallback `auto_inject_custom_mg` uses (see #424), so a failed build degrades to GAMG rather than crashing the solve.

Worth also asserting `velocity_pc == "custom-FMG"` in a test on an adapt child, now that the rotated solve reports it (#465).

Found while auditing the rotated preconditioner in #465; not fixed there to keep that change scoped.

### LINKED ISSUE #468: custom-P FMG smoother has drifted from the native FMG bundle (richardson vs gmres, and an accidental iteration count)
Two code paths write the geometric-multigrid option bundle for the Stokes velocity block, and they have drifted apart. The custom-P path — which is the **mandatory** route for rotated boundary conditions and for `adapt()` children, because native DMPlex MG cannot serve either — is running the configuration the native path measured and rejected.

## Measured, read back off the live PC after setup

| setting | A native (DMPlex refinement) | B custom-P (standard path) | C custom-P (rotated path) |
|---|---|---|---|
| pc / levels / mg_type | mg / 2 / FULL | mg / 2 / FULL | mg / 2 / FULL |
| **smoother KSP** | **gmres** | richardson | richardson |
| smoother PC | sor | sor | sor |
| **smoother max_it** | **4** | 3 | 2 |
| smoother norm type | none | none | none |
| coarse KSP / PC | preonly / redundant | preonly / redundant | preonly / **svd** |

Repro: `~/+Simulations/rotated_pmat_audit/fmg_config_diff.py`.

The coarse `svd` on the rotated path is **deliberate and correct** — the Galerkin-coarsened rotated velocity block inherits the rigid-rotation null space and redundant/LU hits a zero pivot (the #306 fix). Not drift. The other two rows are.

## 1. The smoother kind

`petsc_generic_snes_solvers._configure_preconditioner` sets `gmres`+`sor` with a recorded, measured justification:

> Chebyshev needs eigenvalue estimates of the smoothed operator, which are fragile on the indefinite / variable-viscosity velocity block and diverge. Richardson is stationary and degrades on the NON-SYMMETRIC operator produced by the consistent-Newton tangent. Measured on the Spiegelman notch (Drucker-Prager, eta contrast 1e26) over a nested 4-level hierarchy: contraction per V-cycle rho = 0.75 (richardson) vs 0.56 (gmres) at the SAME four smoother iterations — and the gmres margin GROWS with depth (5% at 3 levels, 25% at 4).

`custom_mg._configure_pcmg` still writes `richardson`+`sor`. It never received that update.

## 2. The iteration count is accidental

`_configure_pcmg` never sets `mg_levels_ksp_max_it` at all, so the smoother inherits whatever is left in the options database under that prefix:

- **standard path → 3**, left over from the GAMG bundle (`opts[f"{prefix}mg_levels_ksp_max_it"] = 3`), which `_configure_pcmg` does not clear;
- **rotated path → 2**, PETSc's own PCMG default, because that prefix is fresh and never saw the GAMG bundle.

So the same function produces different smoothing depending on what ran before it. That is the drift mechanism, not just the symptom.

## Cost

Same rotated operator and right-hand side, transversely isotropic annulus with a weak plane reaching the constrained boundary (eta_1/eta_0 = 1e-3), timed sequentially in-process, three repeats:

| smoother | velocity its | median s |
|---|---|---|
| richardson/2 (rotated path today) | 11 | 0.896 |
| richardson/3 (standard custom-P path) | 9 | 0.720 |
| richardson/4 | 8 | 0.784 |
| **gmres/4 (native bundle)** | **5** | **0.525** |

**1.7x**, and the native path's Spiegelman-notch finding transfers directly to the anisotropic rotated case. Note this is a 2-level hierarchy — the native comment says the gmres margin grows with depth, so a deep adapt-child hierarchy should do better still.

Repro: `~/+Simulations/rotated_pmat_audit/smoother_drift_cost.py`.

## Suggested fix

Give `_configure_pcmg` the same smoother bundle as the native path — `mg_levels_ksp_type=gmres`, `mg_levels_pc_type=sor`, `mg_levels_ksp_max_it=4`, `mg_levels_ksp_norm_type=none`, `mg_levels_ksp_converged_maxits` — and **set** the iteration count explicitly rather than inheriting it. Keep the rotated path's `mg_coarse_pc_type=svd` override.

Better still, and the reason to do it now rather than patch it: have one place own the bundle and both routes read it, so this cannot drift again. The two sites are `_configure_preconditioner` (`petsc_generic_snes_solvers.pyx`) and `_configure_pcmg` (`utilities/custom_mg.py`).

**Validation this needs before merging** (larger than the change): `test_1017_custom_mg_stokes`, the adaptivity suites that rely on mesh-owned custom-P hierarchies, and the rotated suites `test_1018` / `test_1064` / `test_1066` — several carry iteration-count guards and goldens that this will move.

Found while auditing the rotated preconditioner (#465); related to #467 (mesh-owned custom-P hierarchy is not picked up on the rotated path at all).

lmoresi:
Closes #468. Closes #467. Stacked on #465.

## The problem

Three routes reach a multigrid Stokes velocity block, and they are the **same preconditioner reached three ways**, not alternatives:

| route | when | prolongation |
|---|---|---|
| native | `refinement >= 1`, ordinary BCs | PETSc `DMCreateInterpolation` between refined DMPlex levels |
| custom-P, standard | `set_custom_fmg`, or an `adapt()` child's mesh-owned coarse tail | barycentric / RBF, Galerkin coarse operators |
| custom-P, rotated | rotated free-slip, via `rotated_bc` | as above, fine prolongation rotated `P̂ = Q_v·P` |

custom-P is *mandatory* wherever native cannot go — rotated BCs (the DM-coupled hierarchy cannot express a per-node rotation) and `adapt()` children (no DMPlex refinement relation). So the drifted bundle was on the route that matters most.

**#468** — the bundle was written in two places. The native path had been moved to a measured `gmres`+`sor` smoother; `custom_mg._configure_pcmg` never received that update, and never *set* `mg_levels_ksp_max_it` at all, so the smoother inherited whatever last wrote that prefix: 3 on the standard path (left over from the GAMG bundle) and 2 on the rotated path (PETSc's own PCMG default). The same function smoothed differently depending on what had run before it.

**#467** — `mesh.adapt()` leaves a coarse tail on its refinement child so every solver on an adapted mesh gets geometric MG with no per-solver call. The rotated path never consulted it (the standard path's injection hook runs *after* the rotated dispatch has already returned), so an adapt child under rotated free-slip reported `velocity_pc == "GAMG"` — indistinguishable from having no hierarchy at all. That is the `adapt-on-top-faults` workflow's own configuration.

## The change

`src/underworld3/utilities/multigrid_options.py` owns both bundles (geometric MG and the GAMG fallback); all three writers read from it. Two design points worth review:

- **The stale-key list is derived, not written.** A bundle must clear every key a sibling bundle sets that it does not, or `setFromOptions` re-reads a leftover. Deriving it from the union reproduces all three previous hand-written lists *exactly* — which is the check that it is the right abstraction, and it is self-maintaining when a key is added.
- **The coarse solve is a named variant** (`geometric_mg_bundle(coarse="svd")`), not a call-site override, so the one legitimate per-route difference is visible in the same place as everything else. The other native/custom-P asymmetry — native FMG being unusable for single-field solvers because `DMCreateInjection` is unreliable on a refined DMPlex (#276) — deliberately stays *out* of the bundle: it is a routing decision, not an option value, and belongs with the route choice.

`custom_mg.build_transfers` is now the shared "which hierarchy does this solver get?" rule for both the standard injection hook and `rotated_bc._build_rotated_custom_Pl`, with the same opportunistic barycentric→RBF fallback and degrade-to-GAMG on failure.

Also: the rotated FMG branch now drops its option keys after setup, as the KSP's own keys already did. The per-solve prefix is unique, so those keys previously accumulated in the global options database once per timestep.

## Measured

Same operator, RHS and coarse solve. Two levels — where the native measurement says the gmres margin is *smallest*.

| route | pre-fix | unified | isolated linear solve |
|---|---|---|---|
| custom-P standard (isotropic annulus) | 5 vel its (richardson/3) | 4 vel its (gmres/4) | — |
| custom-P rotated (TI annulus, η₁/η₀ = 1e-3) | 11 vel its (richardson/2) | 5 vel its (gmres/4) | 0.683 s → 0.392 s (**1.74×**) |

Reproduce: `~/+Simulations/rotated_pmat_audit/bundle_unification_gain.py` (iterations, cold solve each repeat) and `smoother_drift_cost.py` (the linear solve timed in isolation, setup warm). `fmg_config_diff.py` now shows the three routes agreeing on every row except the deliberate `coarse_pc = svd`.

## Tests

`tests/test_1021_mg_option_bundle.py` reads the smoother configuration back off the **live PETSc objects** for all three routes and asserts they agree, with the coarse difference asserted rather than tolerated. An options-database assertion would not have caught the original drift — the drift was precisely a key nobody wrote. Verified to *fail* when the pre-fix bundle is reimposed.

Also covered: the mesh-owned pickup and its negative control; the rotated FMG PC surviving eight Newton increments (the failure mode the option cleanup introduces); and the same mesh-owned pickup at np>1 in `tests/parallel/test_1064`, where the transfers are built cross-partition.

## Validation

Serial: `test_1014`, `test_1015`, `test_1016`, `test_1017`, `test_1018`, `test_1020`, `test_0753`, `test_0835`, `test_0836`, `test_0840` all pass. Full `level_1 and tier_a` sweep: **526 passed, 2 failed** — both pre-existing and unrelated (`test_0641` still asserts a non-zero rotated datum raises `NotImplementedError`, which #458 implemented; filed as #470).

Parallel np2 **and** np4: `test_1017_custom_mg_parallel_mpi` 6 passed, `test_1064_rotated_freeslip_parallel` 9 passed, `test_1066_rotated_datum_parallel` 2 passed.

**No goldens moved.** The brief anticipated re-recording; none was needed. `GOLDEN_ANNULUS_FMG` is a converged-solution quantity at 1e-6 rtol, so a better preconditioner reaches the same answer, and the iteration guards (`its <= 25`, `max(ksp_its) <= 3`) are upper bounds a better smoother only helps.

### DIFF
--- a/.claude/skills/nonlinear-solver/SKILL.md
+++ b/.claude/skills/nonlinear-solver/SKILL.md
@@ -1,0 +1,186 @@
+---
+name: nonlinear-solver
+description: How to make a hard nonlinear Stokes solve (Drucker-Prager / yield-stress viscoplastic) CONVERGE reliably in Underworld3 the way the working recipe actually does it — automatic warm-start (one Picard step on a cold start) plus a MULTI-SOLVE δ-continuation (constant δ per solve, warm-start the next, sharper δ), the consistent-Newton tangent, and a non-symmetry-safe multigrid smoother. Reach for THIS when a viscoplastic solve stalls / diverges and you are about to hand-tune PETSc options, ramp δ, or "just add a monitor". It carries the CONFIG TRAP LIST — the setup mistakes that each produce a different failure a few steps in — and the one thing you must NOT do (ramp δ inside a single SNES solve). For the yield-law maths and which tangent per model, see `plasticity-solvers`.
+---
+
+# nonlinear-solver
+
+The recipe that gets a **hard viscoplastic (Drucker–Prager) Stokes** problem to
+converge, and — more importantly — the list of setup mistakes that stop it. The
+central lesson from the Spiegelman hard-case study (`η_bg=1e26`, `V=10`): every
+failure was a **solver-configuration** error, not a bad Jacobian. If the correct
+setup is a minefield for an expert, that is an API regression — so the goal is to
+make the correct path the default path.
+
+Design of record: `docs/developer/design/nonlinear-solver-homotopy-warmstart.md`.
+Yield-law maths, tangent-per-model, quadratic-convergence check: `plasticity-solvers`.
+
+---
+
+## The recipe (what actually converges)
+
+1. **Warm start.** Start the continuation at **large δ**, where the yield surface is
+   smooth and the problem is easy, and take **one Picard step** into the Newton
+   basin. One Picard step is defect-correction iteration 1 — contractive, cheap. From
+   a *warm* iterate, take **no** Picard step (it wastes the good quadratic start).
+   A cold `v=0` start is safe on its own terms: `ε̇=0` makes `η_pl` infinite, which
+   the soft-min carries to the viscous branch (see the trap list for the one form
+   that must be written carefully).
+
+2. **Multi-solve δ-continuation** (NOT an in-solve ramp). Hold δ **constant** for a
+   full nonlinear solve to tolerance; warm-start the next, smaller δ from that
+   converged state; march δ down to the sharp surface. δ is a `constants[]` atom,
+   so each step is a recompile-free `PetscDSSetConstants` update.
+
+   **This is now one call** — the model advertises the homotopy and `solve()` marches it:
+
+   ```python
+   stokes.constitutive_model = uw.constitutive_models.ViscoPlasticFlowModel
+   cm.Parameters.shear_viscosity_0 = ...
+   cm.Parameters.yield_stress = ...          # a plain pressure-dependent yield
+   report = stokes.solve(homotopy=True)      # smooth mode, tangent, march: automatic
+   report["settled_delta"]                   # smallest δ that converged
+   ```
+
+   `solve(homotopy=True)` sets the smooth mode (softmin + power-mean), picks the
+   tangent the model asks for (Newton for DP, Picard for elastic VEP), and runs a
+   residual-guided march that accelerates on easy steps and reverts + retries a
+   failed δ more gently. Tune with
+   `homotopy_options=dict(delta0=…, down=…, dmin=…, entry_maxit=…, step_maxit=…)`;
+   the driver is also callable directly as
+   `underworld3.systems.yield_continuation`.
+
+3. **Consistent-Newton tangent** for non-elastic DP (`consistent_jacobian=True`);
+   **Picard** for elastic VEP — see `plasticity-solvers` for the per-model table.
+
+4. **`bt` line search** with the consistent tangent on a smooth (δ>0) surface.
+
+The δ-march is cheap: with the power-mean smoother a converged δ warm-starts every
+sharper δ in ≈0 Newton iterations, so a residual-guided auto-descent costs almost
+nothing.
+
+---
+
+## DO NOT ramp δ inside one SNES solve
+
+Ramping δ **inside a single SNES solve** (a `SNESSetUpdate` callback that sharpens
+the yield surface between Newton iterations) is **proven dead** — it diverges
+`DIVERGED_LINEAR_SOLVE` after ~2 iterations and grinds for ~2 hours, **even on the
+proven solver config**. Mechanism: the continuation only sharpens δ from a
+**converged**, well-conditioned iterate; the in-SNES ramp sharpens δ **mid-solve**
+at a far-from-solution iterate where the consistent-Newton Jacobian on a sharpening
+surface is ill-conditioned and the linear solve fails. **Hide the *continuation*,
+not the *ramp*.** (This supersedes the `enable_yield_homotopy()` in-SNES ramp still
+described in `plasticity-solvers`; that path is retained only as a dead-experiment
+record — use the multi-solve continuation above.)
+
+---
+
+## CONFIG TRAP LIST
+
+Each of these produces a *different* failure a few steps in — that is why the hard
+case felt like whack-a-mole. Check them first.
+
+| Trap | Symptom | Fix |
+|---|---|---|
+| Consistent Newton makes the velocity block **non-symmetric**; a Chebyshev/Richardson MG smoother assumes SPD | smoother diverges / stalls → `DIVERGED_LINEAR_SOLVE` or an endless grind | **Now the default** — the FMG bundle ships `mg_levels_ksp_type=gmres` + `pc_type=sor` + `norm_type=none`. Only an issue if you override it, or on GAMG (which uses PETSc's chebyshev default) |
+| `preconditioner="fmg"` (vs explicit `pc_type=mg` + manual mg opts) | outer KSP "converges" in **1 iteration** → no real Newton correction → stall → `DIVERGED_LINE_SEARCH` | use explicit `pc_type=mg` with the smoother opts above; bound the outer KSP (`ksp_max_it`~80) so a hostile step fails fast |
+| Cold plastic start `v=0`, or any rigid/unyielded point | `DIVERGED_FNORM_NAN` at iteration 0 | **Not** a div/0: `ε̇=0` gives `η_pl=+inf`, which `Min` and the sqrt soft-min carry correctly to the viscous branch. Only a soft-min form that computes `η_ve·η_pl/(η_ve+η_pl)` breaks (`inf/inf`). Fixed in the power-mean; if you hand-roll a blend, write the harmonic mean as `η_ve/(1+η_ve/η_pl)`. **Do not reach for a strain-rate floor** — it hides this rather than fixing it |
+| LU velocity block with all-Dirichlet-ish BC | pressure nullspace singular | attach the Stokes nullspace / avoid a bare LU there |
+| Hand-rolled `snes_monitor` to "see what's happening" | you read residuals but miss the tell | use `solve_with_diagnostics` / `get_snes_diagnostics` instead (below) |
+
+**The diagnostic tell:** `solver.get_snes_diagnostics()["linear_iterations"] ≈ 1`
+per Newton step means the linear solve is doing **no real work** (the FMG-1-iteration
+trap). A healthy consistent-Newton solve does real Krylov work each step and
+converges quadratically. Use `solve_with_diagnostics()`, not a hand-rolled monitor.
+
+---
+
+## Automatic warm-start (Layer 1 — landed)
+
+`solver.has_solution` is a **public, read-only** status flag: `True` only after a
+solve whose SNES converged; reset on a structural rebuild (remesh / adapt /
+mesh-mover — the `is_setup=False` hook); kept through coefficient changes (viscosity,
+δ, BC values, time step). A **diverged** solve leaves it `False`, so the next solve
+auto-cold-starts rather than warming off a corrupted iterate.
+
+On a **cold** (`zero_init_guess=True`) Stokes solve under the **consistent-Newton
+tangent**, a single Picard step is now taken automatically (reusing the existing
+`picard=1` machinery). The default (frozen) tangent path is bit-identical.
+
+```python
+stokes.consistent_jacobian = True
+stokes.solve()                 # cold → one automatic Picard step, then Newton
+if stokes.has_solution:
+    ...
+```
+
+---
+
+## Implementation status (this line of work)
+
+- **Layer 1a — DONE:** `has_solution` + cold consistent-Newton Picard warm-up
+  (`petsc_generic_snes_solvers.pyx`; test `test_0201`).
+- **Layer 1b — DONE:** `zero_init_guess` is tri-state — `None` (default) auto-detects
+  from `has_solution`, `True` forces fresh, `False` insists on warm. Note warm and cold
+  agree only to the *convergence tolerance*, not bitwise.
+- **Layer 3 — DONE:** the FMG velocity smoother defaults to `gmres`+`sor` with
+  `mg_levels_ksp_norm_type=none` (fixed-cost V-cycle), unconditionally — see
+  "Multigrid depth" below.
+- **Layer 2 — DONE:** the model advertises the homotopy
+  (`supports_yield_homotopy` / `_yield_homotopy_control`) and
+  `stokes.solve(homotopy=True, homotopy_options=...)` runs the residual-guided
+  continuation, returning the march summary.
+
+---
+
+## Multigrid depth — how to measure a smoother honestly
+
+**A two-level hierarchy is a coarse-grid correction, not a V-cycle.** Smoother
+comparisons made on one are misleading: the gmres-over-richardson margin measured on
+the Spiegelman notch is only 5 % at 3 levels but **25 % at 4** (ρ per V-cycle 0.746 →
+0.560), because a deeper cycle applies the smoother on more coarse operators. Judge a
+smoother at depth or not at all.
+
+To get depth without a monster problem, refine a **deliberately ultra-coarse NESTED
+base**: `make_notch_mesh.py 1` (492 cells) + uniform `refinement=N` gives 3 levels /
+7,872 cells at `N=2` and 4 levels / 31,488 at `N=3` — deeper *and* smaller than the old
+2-level 38,580-cell setup. In MG you want the coarsest grid as coarse as it can be
+before the problem breaks down.
+
+- **Never use a non-nested hierarchy** here — it does not give strong MG convergence
+  (maintainer ruling). Uniform refinement nests by construction.
+- Accepted tradeoff: uniform refinement does **not** snap new boundary nodes back to
+  the analytic notch arcs (no CAD/EGADS model attached), so the corner geometry is
+  frozen at the coarse mesh's chords on every level.
+
+Measure with `fmg_contraction_probe.py` (ρ_MG per V-cycle; `<0.5` healthy, `0.8–0.95`
+struggling, `≥0.98` hangs) or `smoother_depth_sweep.py` (pays the mesh build + viscous
+seed once, sweeps smoothers in-process) in the Spiegelman study.
+
+**`solve_report` cannot see the smoother.** It records the *Newton* contraction; the
+outer KSP is Eisenstat–Walker-collapsed to ~1 iteration/step, so the smoother's work
+hides inside the velocity sub-block. Probe the `fieldsplit_velocity_` sub-KSP directly.
+
+**A smoother will not rescue small ξ.** At the hard corner the failure is operator
+conditioning — the coarsest grid cannot represent the viscosity contrast — and at 4
+levels *every* smoother fails there (richardson outright, gmres with ρ>1). Use the δ/ξ
+continuation to stay in the solvable region.
+
+## Gotchas
+
+- **`./uw build` → `amr-dev` env**; verify `uw.__file__` is the worktree site-packages.
+- **Run VEP/consistent-Newton tests UNFORKED** — `pytest --forked` SIGABRTs (fork of
+  multithreaded PETSc).
+- Benchmark **every** default change — "Solver Stability is Paramount".
+- ξ (rate-strengthening) is a **non-homotopic** regularisation: put a user loop
+  *around* `solve()`, never inside the δ-march.
+
+## Reference
+
+- Design: `docs/developer/design/nonlinear-solver-homotopy-warmstart.md`,
+  `jacobian-consistent-tangent.md`, `solver-strategies-catalogue.md`.
+- Continuation driver: `underworld3.systems.yield_continuation`.
+- Diagnostics: `SNES_*.get_snes_diagnostics()` / `solve_with_diagnostics()`.
+- Related skills: `plasticity-solvers` (yield law + tangent per model),
+  `free-surface-convection`, `adaptive-meshing`.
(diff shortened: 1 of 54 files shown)


lmoresi:
## Adversarial review

We reviewed our own diff before marking this ready. Findings, most serious first.

### 1. The smoother we unified on costs wall clock on the standard route, at every depth we measured

This is the weakest point of the change and it deserves the top of the list rather than a footnote.

#468 asked for the native smoother bundle on the custom-P route. We measured what that costs. Outer Stokes KSP timed in isolation (setup warm), nested annulus hierarchy, **linear** solve, η contrast 1e6, varying only the velocity-block smoother on the live PC:

| depth | richardson/3 | gmres/4 | its | time |
|---|---|---|---|---|
| 2 levels | 16 vel its, 0.752 s | 11 vel its, 0.875 s | ×1.45 | **×0.86** |
| 3 levels | 8 vel its, 2.569 s | 5 vel its, 3.346 s | ×1.60 | **×0.77** |
| 4 levels | 6 vel its, 10.870 s | 5 vel its, 19.794 s | ×1.20 | **×0.55** |

gmres/4 wins on iterations everywhere and loses on wall clock everywhere, by a margin that *grows* with depth — the opposite of the depth behaviour the native comment records. Decomposing at 2 levels: richardson/3 → richardson/4 is 2.382 → 2.456 s (the fourth sweep, ~3%); richardson/4 → gmres/4 is 2.456 → 3.437 s (gmres itself, ~40% per cycle). Custom-P uses Galerkin RAP coarse operators built from barycentric transfers, denser than the native nested ones, so the smoother is a larger share of the cycle here than on the route the original measurement came from.

We do not think this refutes that measurement. It was taken on the Spiegelman notch with the consistent-Newton tangent — a **non-symmetric** operator, where richardson is stationary and degrades. The sweep above is linear and symmetric: richardson's best case. The complementary evidence is the rotated TI case in the PR body, which *is* anisotropic and where gmres/4 wins 11 → 5 iterations and 0.683 → 0.392 s.

**The gap we could not close:** an attempt to measure the non-symmetric case on the standard route (`smoother_depth_scaling.py -uw_nonlinear 1`) did not converge (`DIVERGED_LINE_SEARCH`), so there are no numbers for it. The script now refuses to time a failed solve rather than report whatever operator the line search abandoned. Someone should close this with a standard-path non-symmetric case that converges.

**The call we took:** unify on gmres/4 anyway. Richardson's failure on a non-symmetric operator is a robustness failure; gmres's cost on a symmetric one is a speed cost, and defaults err toward robust generality. The consistent-Newton tangent is the recommended setting for nonlinear rheologies, so the symmetric regime is the linear case only. But this is a defaults decision with a measured price on the linear path, and it is @lmoresi's to confirm, not ours.

### 2. We widened the scope past what #468 asked for

#468 is about the geometric-MG bundle. We also brought the **GAMG fallback** into the owner, because `rotated_bc` carried a hand-copied duplicate of the native GAMG bundle (commented "tuned to native parity") — the identical two-writer structure, one drift away from the same failure. Ten lines, and it removes a third copy.

We did **not** touch the `pc_mg_type=additive` / `mg_levels_ksp_max_it=3` blocks in the per-solver `__init__` defaults (`petsc_generic_snes_solvers.pyx`, five sites). Those are per-solver defaults rather than the managed bundle, and folding them in is a behavioural change well beyond this PR. They remain duplicated.

### 3. The rotated path now deletes its whole option bundle after setup — a new silent failure mode

Previously the FMG branch deleted one key (`mg_coarse_pc_type`) and leaked the rest. The prefix is unique per rotated solve, so those keys accumulated in the global options database once per timestep — the module docstring already claims the cleanup keeps it bounded, and the FMG branch did not honour that.

Deleting the rest is only safe if nothing re-reads them: a later `setFromOptions` on the velocity sub-PC would find `pc_type` gone and silently abandon the multigrid. We added `test_rotated_fmg_survives_repeated_newton_increments` in the same commit — eight Newton increments on a power-law rotated solve, all keeping `velocity_pc_type == "mg"` and `vel_its_last` flat at 5. It is a check, not a proof: a future PETSc that re-reads options inside `PCSetUp` would break it, and the test is the thing that would say so.

### 4. `_build_rotated_custom_Pl` mutates `solver._custom_mg` after an opportunistic pickup

Registering a mesh-owned hierarchy on the solver promotes it from opportunistic to solver-set, so a *later* build failure would raise instead of degrading to GAMG. `auto_inject_custom_mg` has done exactly this since #424, so this is parity rather than a new semantic — but it is a semantic worth being deliberate about, and we were.

### 5. The parity test's coverage is narrower than the claim

`test_1021` compares the three routes on a **2-level** hierarchy in serial. It would catch the drift this PR fixes (verified: it fails when the pre-fix bundle is reimposed) but it does not exercise depth, and the native route it compares against is the only one we can build natively on this geometry. The mesh-owned pickup is covered at np>1 in `test_1064`; the bundle parity is not.

### 6. What we checked and found clean

- The derived stale-key lists reproduce all three previous hand-written lists **exactly** — that equality is the evidence the derivation is the right abstraction, not a re-implementation with a different footprint.
- Value types crossing into the shared bundle: `pc_gamg_repartition` moved from the string `"true"` to `True`, and `mg_levels_ksp_converged_maxits` from `"true"` to a bare flag. Both reach PETSc identically (`setValue` maps `True` → `"true"`, `None` → a set flag, and a bare bool flag is TRUE). The GAMG-route tests in `test_1064` pass at np2 and np4 against unchanged goldens, which is the behavioural check.
- `_configure_pcmg`'s new `coarse=` keyword defaults to the previous behaviour; the two `_install_transfers` call sites are unaffected.
- No goldens moved. We expected to re-record and did not have to: `GOLDEN_ANNULUS_FMG` is a converged-solution quantity at 1e-6 rtol, so a better preconditioner reaches the same answer, and the iteration guards are upper bounds.
- Two failures in the `level_1 and tier_a` sweep are pre-existing and unrelated (#470).

lmoresi:
## Follow-up: the smoother's recorded mechanism is wrong (2550d9a4)

@lmoresi queried the "non-symmetric" framing in finding 1 above. He was right, and it does not survive measurement. Correcting it is now part of this PR — comments and docs only, no behavioural change.

**The claim.** Both the bundle comment we moved into `multigrid_options.py` and the Layer 3 section of `design/nonlinear-solver-homotopy-warmstart.md` state that richardson degrades because the consistent-Newton tangent makes the velocity block non-symmetric. We inherited it without checking.

**On paper it is wrong.** For an isotropic η(ε̇_II) the Newton term is `2 (η'/ε̇_II) (ε̇(u):ε̇(δu)) (ε̇(u):ε̇(v))` — a rank-one outer product a⊗a with a = ε̇(u), symmetric under swapping δu and v. η depends on ∇v only *through* its symmetric part, so both factors project onto the same tensor.

**Measured**, `||A - A^T||_F / ||A||_F` on the assembled velocity block:

| rheology | asymmetry | Newton term `||A_N - A_P||/||A||` |
|---|---|---|
| linear (calibration) | 5.1e-17 | — |
| shear-thinning, isotropic | 5.3e-17 | 2.1e-01 |
| power-law, isotropic | 5.3e-17 | 4.0e-01 |
| Drucker-Prager, pressure-dependent yield | 5.1e-17 | 5.1e-02 |
| transverse isotropic, **Picard** tangent | **7.2e-02** | — |
| transverse isotropic, Newton tangent | **8.9e-02** | 2.0e-01 |

The third column is the control — a null symmetry reading is vacuous if the consistent tangent contributed nothing. It contributed 5-40% of `||A||` every time, and the block stayed symmetric anyway.

Non-symmetry appears only under transverse isotropy, and there it appears under the **Picard** tangent too, where `int ε̇(δu):C:ε̇(v)` is symmetric by construction. Linear TI with anisotropy fully active is symmetric at 6.3e-17. That is a defect signature, not a property of the tangent — reported on #457, whose scope as written ("the anisotropy part of the *consistent* tangent") now looks too narrow.

**A trap worth recording**: petsc4py's `Mat.transpose()` transposes IN PLACE and returns `self`. Called on the matrix rather than a copy it measures `A - A` and reports an exact zero for every operator. Our first run "confirmed" everything was symmetric that way, and the tell was that it was *exactly* zero in cases that had no business agreeing.

**What this does and does not change.**

The Spiegelman-notch contraction measurement (ρ 0.75 richardson vs 0.56 gmres, margin growing with depth) is untouched — it is a measurement, not a derivation. Only the explanation attached to it was wrong. The mechanism that fits both it and the cost in finding 1 is **operator conditioning**: at η contrast 1e26 the SOR-preconditioned spectrum is spread far enough that a stationary iteration stalls where a Krylov smoother adapts its polynomial. That predicts the notch result *and* predicts that gmres loses on the well-conditioned problems we measured.

It also dissolves the gap we flagged in finding 1. We were looking for a "standard path × non-symmetric operator" case to justify gmres there. That cell does not exist for isotropic rheologies — the standard-path block is symmetric whether the tangent is Picard or Newton. So the discriminator is not symmetry but conditioning, and the honest position is:

- gmres/4 is the right default because it is the one that does not stall on high-contrast operators — the regime this code exists for;
- it costs up to ×1.8 in the linear solve on well-conditioned problems, at every depth we measured;
- both facts are now recorded in `multigrid_options.geometric_mg_bundle` and `subsystems/solvers.md` so removing the guardrail is a deliberate act.

The design doc's Layer 3 claim is marked with a `TODO(DESIGN)` rather than rewritten — it is #441's document, and the correction is bigger than a drive-by edit.

lmoresi:
## Scope note: this does not un-gate geometric MG for single-field solvers (#478)

A parallel session flagged that geometric FMG is currently gated off. Checking what this PR does and does not close: the gate is real, it is the #276 single-field lockout, and **#471 does not remove it** — deliberately, as the one design question the brief asked us to settle.

Measured off the live PC after setup:

| solver | mesh | `preconditioner=` | live PC |
|---|---|---|---|
| Stokes | refinement=1 | auto | **`mg(2)`** |
| Stokes | refinement=2 | auto | **`mg(3)`** |
| Poisson | refinement=1 | auto | `gamg` |
| Poisson | refinement=2 | **fmg (explicit)** | **`gamg`** + warning |

Stokes is fine. Every single-field solver gets algebraic multigrid at any depth, even when geometric is requested explicitly, via `if want_fmg and prefix == "": want_fmg = False`.

We kept that out of the shared bundle owner on the grounds that it is a **routing** decision — which route a solver may take — not an option value, and so belongs with the route choice rather than with the settings. That reasoning still holds. But it does mean the headline "all three routes now agree" is scoped to the routes a solver is *allowed* to reach, and for scalar and vector solvers there is only one.

What this PR does change for them: the sanctioned workaround (`set_custom_fmg`, which the lockout warning points at) previously installed the drifted bundle — richardson at an iteration count nobody set. It now installs the same bundle as everything else. Same for the mesh-owned auto-pickup on adapt children, which already worked for single-field solvers. So single-field geometric MG got better where it was reachable; it did not become reachable.

Filed as #478 with a proposed fix, because #471 makes it small: `mesh.dm_hierarchy` is already a coarse tail, barycentric transfers with Galerkin RAP need no injection (which is the whole of err62), and `custom_mg.build_transfers` is now the single hierarchy-resolution rule with a safe degrade. It wants its own PR — it is a default change across every scalar and vector solver.

Repro: `~/+Simulations/rotated_pmat_audit/fmg_gate_audit.py`.

lmoresi:
## Holistic adversarial review — all six commits

The earlier review covered `99761e02` only. This is a fresh pass over the PR as it now stands, since three commits landed after it. **Two real defects found, both introduced by this PR's own later commits, both fixed in `ebb2d337` with tests written to fail on the old code.**

### 1. `strategy="fast"` was honoured on one route and silently dropped on the other two

The worst possible defect for this PR to ship, because it is the exact disease the PR exists to cure.

`_configure_pcmg` had no smoother argument, so both custom-P routes always got `"robust"` regardless of what the user asked for:

| route | asked `fast` | got |
|---|---|---|
| native | richardson/3 | **richardson/3** |
| custom-P standard | richardson/3 | **gmres/4** |
| custom-P rotated | richardson/3 | **gmres/4** (+svd) |

And it is dropped on the two routes that matter *most* — custom-P is mandatory for rotated BCs and `adapt()` children, which is the argument in this PR's own description.

**Why it survived:** the only test of the strategy axis used a `refinement=1` mesh, i.e. the native route. A three-route feature tested on one route. The parity test (`test_all_three_routes_share_one_bundle`) did not catch it either, because it compares the three routes to *each other* at the default strategy, where they agree.

Fixed: `_configure_pcmg(smoother=...)`, every install path passes `solver._mg_smoother_variant`, and `test_strategy_reaches_every_route` asserts all three.

### 2. The strategy value broke `pickle`, `copy` and `deepcopy`

`_StrategyName.__new__(cls, name, summary)` takes two arguments; `str.__reduce_ex__` reconstructs via `cls(value)` with one. So every one of these raised `TypeError`:

```
pickle   -> TypeError: __new__() takes exactly 3 positional arguments (2 given)
copy     -> TypeError: ...
deepcopy -> TypeError: ...
```

`13bd82ce`'s commit message asserted "comparisons, formatting and **serialisation** are unchanged". Two of three were true. UW3 has a serialisation system that would have hit this. Fixed with `__reduce__`; test round-trips through all three.

### 3. Checked and clean

- **A custom-P install does not report spurious user overrides.** The ownership record is keyed by fully-qualified name while `preconditioner_settings` reads through the prefixed view; `_user_overridden_pc_options` returns `()` after a `set_custom_fmg` solve, so the two prefix conventions do line up. This was the specific worry after the F-1 regression, which was the same class of mistake.
- **The defaults arm is unchanged** at every step — the three routes still agree on every row but the deliberate rotated `svd`. That arm has now caught one regression in this PR (F-1's unqualified record) and is the reason the ownership work is trustworthy at all.
- **`describe()` degrades honestly** on a partial settings dict (`?` rather than a `KeyError`), and refuses to report before resolution.

### Residual risk we are accepting, stated plainly

- **`preconditioner_settings` returns strings**, because that is what the PETSc options database holds (`mg_levels_ksp_max_it` is `"4"`, not `4`). Tests assert against strings. Honest but slightly awkward; typing it would mean duplicating each key's type.
- **`_pc_resolved` is set even when the user-override latch early-returns.** Defensible — the resolution decision *was* made, it was "leave it alone" — but the report then describes options UW3 did not choose. It does label them as user-overridden.
- **The rotated route's option keys live under a per-solve prefix** (`rotfsN_`) that no user can address, so a user override cannot reach the rotated velocity block at all. Pre-existing, out of scope here, and worth its own issue.
- **`strategy="fast"` is not measured on the rotated route.** The variant is now *plumbed* there and asserted, but the fast/robust wall-clock comparison was only measured on standard/native. The regime argument should carry over; it is not verified.

### Scope

Six commits: bundle owner + #467 pickup; mechanism correction; F-1 ownership; strategy fill + rename; reporting; these fixes. Larger than "one owner for the bundle", and @lmoresi has signed off on keeping it whole on the grounds that the evolution is legible. Each commit after the first came from a measurement or a review finding rather than from scope drift.

Validation on the final state: `level_1 and tier_a` 526 passed with two pre-existing #470 failures; 79 passed across the MG/adaptivity/solve-report suites; parallel np2 and np4 17 passed each; no goldens moved.

PR #474, opened 2026-07-29
Update solver template DDT imports

ss2098:
This PR updates src/underworld3/systems/solver_template.py to use the current DDT class names exported by underworld3.systems.ddt.

The old names SemiLagrangian_DDt, Lagrangian_DDt, and Eulerian_DDt are no longer exported. The current public names are SemiLagrangian, Lagrangian, and Eulerian.

This updates the import and matching type hints only.

Validation:

- pixi run python src/underworld3/systems/solver_template.py

### DIFF
--- a/src/underworld3/systems/solver_template.py
+++ b/src/underworld3/systems/solver_template.py
@@ -9,7 +9,7 @@ import sympy
 from typing import Optional, Union, Callable
 import underworld3 as uw
 from underworld3.systems import SNES_Scalar, SNES_Vector, SNES_Stokes_SaddlePt
-from underworld3.systems.ddt import SemiLagrangian_DDt, Lagrangian_DDt, Eulerian_DDt
+from underworld3.systems.ddt import SemiLagrangian, Lagrangian, Eulerian
 from underworld3 import timing
 from underworld3.systems.solvers import expression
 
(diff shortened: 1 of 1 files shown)


ss2098:

MPI_Init_thread failed

This PR only updates solver_template.py DDT import/type-hint names. Local validation passed with:

pixi run python src/underworld3/systems/solver_template.py

lmoresi (approved):
# Adversarial review

We verified the rename is real and complete: `systems/ddt.py` exports `SemiLagrangian`, `Lagrangian`, `Eulerian`; no `*_DDt` names remain anywhere in the PR-head file. This fixes a live breakage — on current development, `import underworld3.systems.solver_template` fails with `cannot import name 'SemiLagrangian_DDt'`; the PR version imports and executes clean.

Attacks that failed: partial-rename hunt (0 `_DDt` remnants); deprecated-pattern sweep (0 hits); py_compile clean.

One optional nit: the newer `Symbolic` DDt class could join the type-hint Union — not required here.

The failing `test` check was the known runner-infrastructure flake (#461), rerun. Merging once CI is green. Thanks @ss2098 — this and #481/#482 are exactly the kind of API-currency fixes that rot silently.

Issue #473, opened 2026-07-29
yield_continuation: the documented cold-start guarantee does not hold, and the step control is effectively one-shot


lmoresi:
Two independent defects in `src/underworld3/systems/yield_continuation.py`, both found
while marching the Spiegelman notch. Filed together because a maintainer will want to look
at the march loop once.

## 1. The cold-start claim is false on a `Piecewise` yield stress

The docstring promises:

> Starting δ is deliberately **large**, where the power-mean soft-min stays bounded by the
> background viscosity even as ε̇ → 0. That is what makes the first (cold) solve well posed,
> and why no separate viscous pre-solve is needed.

Measured on the notch: the cold power-mean entry at `delta0 = 1` **fails at all four
regime points tested**. Every driver in that study needs an explicit viscous pre-solve
first, which is precisely what the docstring says is unnecessary.

Suspected cause (not yet isolated): the notch's yield stress is a `sympy.Piecewise` with a
deliberately unreachable τ_y in the weak inclusion, so the "bounded by the background
viscosity" argument does not hold pointwise across the material discontinuity. If that is
the cause then the claim is true only for a *single-material* yield law and the docstring
should say so.

`tests/test_1057` covers the claim on a simpler model, where it does hold — so the test
suite does not currently see this.

## 2. The adaptive step is one-shot in practice

```python
if nit <= max(2, budget // 5):
    step = max(step * step, 0.05)        # accelerate
elif nit >= 0.8 * budget:
    step = min(step ** 0.5, 0.95)        # decelerate
d *= step
```

With the default `step_maxit = 10` (and with 13, as used in the study) `budget // 5` is 2,
so **the accelerate branch requires a warm δ-step to converge in ≤ 2 nonlinear
iterations** — the threshold does not scale with the budget in the regime where the budget
is actually set. In the study runs the branch fired once at entry and never again, leaving
`step` pinned at ×1/4 for the rest of the march.

The deeper issue is that `step` has **no return path to the nominal `down`**: each branch
mutates it monotonically and only the *opposite* branch can move it back. So a single
early decision persists for the whole march. That matches the non-monotone stalls observed
on both sweep axes.

Suggested: make the accelerate threshold a fraction of the budget with a sane floor that
is reachable for a warm step, and relax `step` back toward `down` on an ordinary
(neither-branch) success so one early sample cannot set the pace of the entire march.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

[...]

Issue #472, opened 2026-07-29
shear_viscosity_min: the floor's rounding scale is tied to yield_mode (and the VEP path silently skips it)


lmoresi:
## What

`Constitutive_Model._apply_floor` (`src/underworld3/constitutive_models.py:1160`) picks the
rounding scale for a viscosity/yield-stress floor from `yield_mode`:

```python
rounding = 0 if getattr(self, "_yield_mode", "min") == "min" \
    else self._get_yield_softness() * floor
return uw.maths.smooth_max(value, floor, rounding)
```

Two consequences, and they are the two halves of the same design gap — the floor has no
smoothing parameter of its own.

**1. Under `yield_mode="min"` the floor is a hard `sympy.Max`.** That is correct for the
frozen/Picard tangent, but `consistent_jacobian=True` differentiates through the kink and
the solve dies immediately. Measured on the Spiegelman notch, isolated with a
floor-on/off × Newton/frozen 2×2 — the only failing cell is (floor on, Newton), at
`nl=0, DIVERGED_LINEAR_SOLVE`. So a user who sets a physically sensible viscosity floor
silently loses the consistent tangent.

**2. Under `yield_mode="softmin"` the rounding is `δ·floor`, so it vanishes as δ→0.**
The floor is smooth at the start of a yield homotopy and hard at the end — i.e. the
smoothing disappears exactly where the march is trying to land. δ is the *yield corner's*
regularisation parameter; the floor is a different corner and should not inherit it.

## Why it matters now

A **viscosity floor that is relaxed toward zero** is one of the classical continuation
routes for viscoplastic Stokes (it bounds the plastic viscosity from below and is then
removed — structurally the same idea as the Perzyna-style rate term ξ). That route cannot
be measured against the soft-min homotopies at all while (1) forces it onto the Picard
tangent and (2) couples its smoothing to the parameter being marched.

## Suggested shape of a fix

Give the floor its own rounding scale rather than deriving one from `yield_mode`:

- an explicit smoothing parameter on the floor (absolute, or relative to the floor with
  its own coefficient), defaulting to 0 so today's `min` behaviour is unchanged;
- `_apply_floor` uses that parameter regardless of `yield_mode`.

There is a related `TODO(DESIGN)` already in the same function (lines 1180-1186) about the
relative rounding collapsing for a tension cutoff at `floor = 0` — the two want settling
together, since both are "this corner has no length scale of its own".

[...]

PR #475, opened 2026-07-29
yield_anchor: which side of exact Min the soft-min yield law sits on

lmoresi:
## What this is

A soft-min rounds the yield corner by δ, and rounding a corner moves the whole curve — so
each soft-min family has one free constant. Until now it was fixed implicitly, and
differently in each family, with the consequence buried in the algebra: **both families
ended up sitting BELOW exact `Min` at and above the yield point.**

For a homotopy that is the wrong direction. The entry problem is supposed to be an easier
version of the sharp problem; a law that sits below it is a *weaker* problem, which is a
different thing. The sqrt family undershoots for overstress ratios $f < 2$ — a third
under-stressed at nominal yield — and the power mean degenerates to the p-norm
$(\eta_{ve}^{-s} + \eta_{pl}^{-s})^{-1/s}$, whose viscosity falls away toward zero as δ
grows.

`yield_anchor` names the choice and makes it explicit, for both families:

| `yield_anchor` | pinned exactly at | sqrt family | power mean |
|---|---|---|---|
| `"onset"` (default, historical) | $f \to 0$, the unyielded branch | undershoots for $f<2$ | the p-norm; $\eta \to 0$ as δ grows |
| `"yield"` | $f = 1$, the yield point | offset $\delta/2$; $\ge$ `Min` | the true power mean; $\ge$ `Min` |

The power-mean case is the neat one: it needs **no offset at all**. Averaging the two
terms — the `/2` — is exactly what makes it a generalised *mean* rather than a *norm*, and
a generalised mean returns the common value when its arguments are equal. So
$\eta = \eta_{ve} = \eta_{pl}$ at $f = 1$ falls out of the algebra.

## The cost, stated plainly

`"yield"` stiffens *unyielded* material: bounded by a factor 2 for the sqrt family, and by
$2^{\delta}$ for the power mean. At the power mean's own entry δ₀ = 1 that is a factor 2
on the entire background. Both decay to 1 as δ → 0, so the endpoint is unchanged, but the
entry problem is not merely "the sharp problem with a rounded corner" — the background is
genuinely stiffer along the path. That bound is also why the two families' entry δ are not
the same number, which is the second half of this change.

## δ is not one parameter

For the power mean the sharpness is $s = 1/(\delta + 0.001)$, so δ ≤ 1 and δ = 1 *is* the
harmonic mean. For the sqrt family δ is a percentage stress deviation and a generous entry
is O(10). A sqrt-family δ = 16 handed to the power mean asks for a law a factor 2^16 away
from `Min`. The entry now comes from the family (`_YIELD_HOMOTOPY_DELTA0`) rather than
from a caller carrying one number across both, and `yield_continuation` grows `smoother=`
and `anchor=` so a march can pick both — refusing them, rather than silently ignoring
them, when handed a ready-made control.

## What is NOT changed

**The default stays `"onset"`.** Flipping it changes results for every existing
`yield_mode="softmin"` user, and that is a maintainer decision rather than a side effect of
naming the knob. Left open deliberately — see the question below.

## Tests

34 passed. `test_1059_yield_anchor.py` is new and parametrised over both families; each
test that pins the new behaviour also asserts the OLD anchor **fails** it, so none of them
can pass against a no-op implementation. `test_1055` now sets the anchor explicitly where
it exercises the below-`Min` side, rather than relying on the default.

## Open question for the maintainer

Should the default flip to `"yield"`? The argument for: it is the correct direction for a
homotopy, and it is what any new user opting into `softmin` for continuation will want.
The argument against: it changes converged results (within tolerance, but not identically)
for anyone already using `softmin` as a plain regularisation rather than as a march, and
for the power mean it stiffens the background by up to 2^δ.


lmoresi:
## Adversarial review

Read against the diff, not the description. Four findings, none blocking; the two worth acting on before merge are R1 and R4.

**R1 — `yield_anchor` rebuilds the δ atom, so an already-built `YieldHomotopyControl` holds a stale handle.**
The setter nulls both `_yield_offset_expr` and `_yield_softness_expr` and calls `_reset()`. That is correct for the *value* — `_get_yield_softness` rebuilds from `self._yield_softness`, which the setter does not touch — but it creates a **new** `expression` object. `control.delta` is captured at construction, so

```python
control = cm._yield_homotopy_control()
cm.yield_anchor = "yield"          # rebuilds the atom
control.delta                      # now detached from the model
```

leaves the diagnostics handle pointing at an orphan. `set_delta` is safe (it goes through the `yield_softness` property, which re-fetches), and `_yield_homotopy_control` sets the anchor *before* it reads `delta=`, so the driver path is correct — and `yield_continuation` already refuses `anchor=` alongside a ready-made control, which closes the obvious footgun. What is left is the hand-rolled path. Cheapest honest fix is a line in the `YieldHomotopyControl.delta` docstring saying it is a snapshot; the better one is to make it a property that re-reads.

**R2 — the power mean's `2^δ` background stiffening deserves a cross-reference, not just a bound.**
The description and docstrings both state it, and the algebra checks out (as `eta_pl → ∞`, `softmin → a^-s/2`, so `η → η_ve · 2^(δ+0.001)`). But at the family's own entry `δ0 = 1` that is a **factor 2 on the entire unyielded domain** — which makes the entry problem a materially different problem, not a rounded version of the target. That is very likely relevant to #473's first defect (the cold power-mean entry at `delta0 = 1` failing at all four notch regime points). Worth a `See also` between them so the next person does not diagnose it twice.

**R3 — `_yield_offset_for_anchor` reads `self._yield_softness_expr` directly.**
It is only ever called from inside `_get_yield_softness`, on the line after that attribute is assigned, so it cannot currently see `None`. But the coupling is implicit and the function name does not suggest it. Taking δ as an argument would make it total.

**R4 — no test that the anchor survives a δ update.**
This is the coverage gap that matches the risk R1 describes. The setter *rebuilds* the offset atom from the anchor's formula, and the whole design rests on that atom continuing to track δ symbolically afterwards. Every test in `test_1059` sets the anchor and then evaluates; none does

```
set anchor → set δ → check the pin → set δ again → check the pin again
```

which is exactly the sequence a march performs. A no-op rebuild bug (offset frozen at its construction-time δ) would pass the current suite and fail every homotopy. Cheap to add and it guards the one mechanism that is genuinely subtle here.

**Checked and clean:**
- The `/2` does what it claims: at `a = b`, `((2a^-s)/2)^(-1/s) = a` exactly.
- Sides are right in both directions — the generalised mean `M_{-s} ≥ min`, the p-norm `≤ min`.
- The sqrt family's `δ/2` offset gives `g → 1/2` as `δ → ∞`, hence the stated factor-2 bound.
- Ordering in `_yield_homotopy_control` (smoother → anchor → read δ) is correct; reversing the first two would lose the power mean's δ bump.
- `test_1055`'s new explicit `anchor="onset"` lines are the right move — they stop those assertions from silently inverting if the default is ever flipped, which is precisely the open maintainer question.

lmoresi:
## Heads-up from the solver configuration audit: unpushed commits, and a collision with #471

Two coordination points found while auditing the solver configuration surface against `development` (report: `docs/reviews/2026-07/solver-configuration-reachability-audit.md` on branch `docs/solver-config-audit`).

### 1. Three commits are local only

`feature/regime-diagram` carries three commits that are not on origin:

```
954ff4f3 fix(solver): make Eisenstat-Walker switchable
5ad68b2f docs(solvers): the Stokes fieldsplit is two nested Krylov loops
f1555b3a fix(solver): make the Stokes solver settings reachable (#477, D-22/D2)
```

So this PR does not show the #477 fix. Auditing `development` or reading the PR list rediscovers a solved problem — which is exactly what happened here before we checked the local branch. Worth pushing so the fix is visible to anyone else working this surface.

Verified what it covers: `snes_max_it` and the two `fieldsplit_*_ksp_rtol` keys. It does **not** cover the MG bundle keys (`fieldsplit_velocity_mg_levels_ksp_type` / `_max_it` / `mg_coarse_pc_type`), which are clobbered at *build* time by `_apply_preconditioner_options` rather than at solve time — a different mechanism. That half is fixed in #471 (commit `5a390efe`).

### 2. This PR and #471 both edit the same stale-key list, and #471 subsumes it

This PR patches the hand-written GAMG stale-key clear in `_apply_preconditioner_options`:

```diff
-                        "mg_levels_pc_type", "mg_coarse_pc_type",
+                        "mg_levels_pc_type", "mg_levels_ksp_norm_type",
+                        "mg_coarse_pc_type",
```

#471 deletes that list entirely and **derives** it — every key a sibling bundle sets that this one does not — and the derived GAMG stale set already contains `mg_levels_ksp_norm_type`. So the fix here is correct and already covered by the derivation.

They will conflict textually. Whichever lands second should take #471's derivation rather than merging the two lists, otherwise the hand-written list comes back and so does the class of bug that produced this patch.

### 3. Adjacent findings routed here

Filed as #483: `Stokes` and `Stokes_Constrained` disagree about which options are reachable, in *opposite* directions (`ksp_rtol` settable on one and not the other; the fieldsplit rtols the reverse) — plus `snes_rtol`, which `_reassert_outer_tolerances()` deliberately re-writes, so it is owned by design but still looks like a settable knob. Both belong on this branch, next to the fix they are adjacent to.

lmoresi:
## Pushed the three commits that were local-only, and a heads-up on the conflicts

The #477 reachability fix and its two companions were committed but never pushed, so this PR did not show them and anyone auditing `development` rediscovered a solved problem — which is exactly what happened during the configuration audit (#485). Now visible:

```
f1555b3a fix(solver): make the Stokes solver settings reachable (#477, D-22/D2)
5ad68b2f docs(solvers): the Stokes fieldsplit is two nested Krylov loops
954ff4f3 fix(solver): make Eisenstat-Walker switchable
```

### Conflicts against the new `development` — six files, and one needs a decision not a resolution

`development` has moved a long way (#487, #485, #465, #471 all landed):

```
CONFLICT (add/add)  docs/developer/design/nonlinear-solver-homotopy-warmstart.md
CONFLICT (content)  src/underworld3/constitutive_models.py
CONFLICT (content)  src/underworld3/cython/petsc_generic_snes_solvers.pyx
CONFLICT (add/add)  src/underworld3/systems/yield_continuation.py
CONFLICT (add/add)  tests/test_1055_yield_smoother.py
CONFLICT (content)  tests/test_1058_solve_report.py
```

**Not resolved here, deliberately, for two reasons.** The `regime-diagram` worktree has uncommitted work in `constitutive_models.py` — one of the six conflicted files — so a merge from outside that session would collide with live edits. And two of the conflicts are decisions rather than merges:

1. **The stale-key list.** This PR patches the hand-written GAMG clear list in `_apply_preconditioner_options`:

   ```diff
   -                        "mg_levels_pc_type", "mg_coarse_pc_type",
   +                        "mg_levels_pc_type", "mg_levels_ksp_norm_type",
   +                        "mg_coarse_pc_type",
   ```

   #471 **deleted that list** and derives it instead — and the derived GAMG stale set already contains `mg_levels_ksp_norm_type`. So this patch is correct and already covered. Take #471's derivation; do **not** reinstate the list, or the bug class that produced this patch comes back.

2. **`_reassert_outer_tolerances` and the strategy setter.** #471 rewrote the same region: the MG bundles now come from `utilities/multigrid_options.py`, internal option writers go through `SolverBaseClass._push_managed_option` so a user's key is respected, and `solver.strategy` now selects a bundle *variant* rather than writing velocity-block options (its old writes were dead — measured). Anything in `f1555b3a` that writes a bundle key directly should be routed through `_push_managed_option`, or the bundle will treat it as a user override and back off.

Also relevant to this branch's own subject: the `snes_ksp_ew` switchability fix in `954ff4f3` addresses something the audit independently flagged (#483) — `ksp_rtol` is reachable on `Stokes` but not on `Stokes_Constrained`, and `snes_rtol` is owned-by-design but looks settable. Both belong here.

lmoresi:
# Adversarial review

Verified in an isolated worktree at the PR head (amr-dev env, `uw.__file__` checked). The core claims hold:

- Anchor math proven, not just tested: sqrt "yield" anchor ≥ Min for all f (equality at f=1, stiffening ≤ 2×); powermean `/2` exact at f=1, ceiling = geometric mean — matches the design doc and the test's `√0.1·1.001` ceiling.
- Tests are genuinely new coverage: all 7 test_1059 cases FAIL on plain development, with counter-assertions against the un-anchored form. Full run on the PR head: 14 passed, 30.8 s.
- Live solve: `anchor="yield"`, sqrt, δ=4 converges through JIT (reason 3); post-solve flip to onset recompiles and re-solves in the physically correct direction.
- No #437 solve-report interaction: retry/report funnel untouched; the difficulty probe caps max_it on the SNES object, not the options DB.
- Overlap vs #471: #471 is merged and in this PR's ancestry; the diff is strictly net-new. The solve-report off-by-one commit nets to zero against #444. `git merge-tree` clean.
- Issue cross-refs: #472 fixed opt-in (`viscosity_min_rounding`, default None = historical; VEP path honours the floor when set); #473 improved but not closed (step control still one-shot); #463 untouched.

Minor findings, tracked as a follow-up issue rather than blocking (each strictly narrower than pre-PR behavior or cosmetic):

1. `_snes_max_it_user` latch never clears: set 200 → delete → one default solve → stale 200 resurrects on the next. Probed.
2. Docstring attributes the Min side to the smoother family ("powermean … from below, sqrt … from above") — contradicts the PR's own thesis (side belongs to `yield_anchor`); `anchor` is forwardable via `homotopy_options` but missing from the documented keys.
3. `except Exception: return default` at the options read, no comment (Charter swallowed-exception rule).
4. `yield_anchor` setter rebuilds the δ atom; a pre-flip `YieldHomotopyControl` keeps the orphan in `control.delta` (diagnostics-only; `yield_continuation` refuses the combination).
5. `test_yield_anchor_keeps_onset_stress_exact` tests `anchor="yield"` — "onset" in the name is the other anchor.
6. `viscosity_min_rounding` setter accepts any value unvalidated.

Merging.

PR #479, opened 2026-07-29
Update adapt vs ref24 cost script for current APIs

ss2098:
This PR updates scripts/adapt_vs_ref24_cost.py so the cost comparison script runs with the current Underworld3 APIs.

Changes include:
- replace internal mesh._deform_mesh() calls with the public mesh.deform() API
- add fail-fast SNES convergence checks so the script does not continue after failed solves
- simplify the Stokes boundary condition so the cost diagnostic runs stably

Validation:
- pixi run python scripts/adapt_vs_ref24_cost.py

Example local output:
- res-24 plain step: 2.489 s
- res-16 plain step: 1.291 s
- a16s adaptation overhead: 2.562 s
- per-sim-time slowdown a16s / ref24: approximately 0.78x

### DIFF
--- a/scripts/adapt_vs_ref24_cost.py
+++ b/scripts/adapt_vs_ref24_cost.py
@@ -54,7 +54,7 @@ def build(res):
     st.penalty = 0.0
     ur = m.CoordinateSystem.unit_e_0
     st.add_essential_bc((0.0, 0.0), m.boundaries.Lower.name)
-    st.add_essential_bc((0.0, 0.0), m.boundaries.Upper.name)
+    st.add_rotated_freeslip_bc(0, "Upper")
     st.bodyforce = RA * (T.sym[0]
                          - (r_o - r) / (r_o - r_inner)) * ur
     ad = uw.systems.AdvDiffusionSLCN(m, u_Field=T, V_fn=v.sym,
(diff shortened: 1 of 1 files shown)


lmoresi (changes requested):
# Adversarial review

Same three findings as #476, and they bite harder here:

1. **MAJOR** — free-slip → no-slip on the upper boundary (in `build()`). This script's sole purpose is a cost comparison against a stored ref-24 baseline taken under free-slip with warm starts; the comparison is invalidated, not updated. Use `add_rotated_freeslip_bc(0, "Upper")`.
2. **MAJOR** — `zero_init_guess=True` in the timed loops: cold-start timings against a warm-start baseline.
3. MINOR — `deform()` double-counts field transfer in the remap timings (it runs `remesh_with_field_transfer` internally).

Requested changes as on #476. The API updates themselves verified correct.

ss2098:
Thanks , Professor @lmoresi . I restored the intended upper free-slip boundary using add_rotated_freeslip_bc(0, "Upper") and restored warm-start Stokes timing with zero_init_guess=False.

I also added a note that Mesh.deform() includes remesh_with_field_transfer internally, so the remap timing includes that transfer cost before the script overwrites fields.

Validation:
- python -m py_compile scripts/adapt_vs_ref24_cost.py
- pixi run python scripts/adapt_vs_ref24_cost.py

lmoresi (approved):
All three findings verified fixed in 5e1e1f9f: rotated free-slip restored on Upper with the correct value-first signature (line 57), `zero_init_guess=False` on the initial/warm/timed solves (76/82/89), and the deform double-transfer note (106-107). Smoke-ran to completion against current development. Thanks for the careful rework.

PR #481, opened 2026-07-29
Add input checks to stagnant lid comparison plot

ss2098:
This PR adds a preflight input check to scripts/stagnant_lid_adapt_compare_plot.py.

The script is a post-processing plot script and expects existing stagnant-lid checkpoint outputs. Previously, if the hard-coded checkpoint directories or files were missing, PETSc/HDF5 failed later with a low-level H5Fopen error.

This update checks for the required mesh and T checkpoint files before loading them and prints a clear message explaining which inputs are missing and that the user should run/copy the checkpoint outputs first or update U_DIR/A_DIR and U_STEM/A_STEM.

This does not change solver internals or checkpoint loading behavior when the required files exist.

Validation:
- python -m py_compile scripts/stagnant_lid_adapt_compare_plot.py
- pixi run python scripts/stagnant_lid_adapt_compare_plot.py

With missing local checkpoint files, the script now exits cleanly after listing the missing required inputs instead of failing with a PETSc/HDF5 traceback.

### DIFF
--- a/scripts/stagnant_lid_adapt_compare_plot.py
+++ b/scripts/stagnant_lid_adapt_compare_plot.py
@@ -10,9 +10,34 @@ import numpy as np
 import sympy
 import underworld3 as uw
 import underworld3.visualisation as vis
-import pyvista as pv
 
-pv.OFF_SCREEN = True
+
+def required_files(src_dir, stem):
+    return [
+        os.path.join(src_dir, f"{stem}.mesh.00000.h5"),
+        os.path.join(src_dir, f"{stem}.mesh.T_v2p1.00000.h5"),
+    ]
+
+
+def check_required_inputs(cases):
+    missing = []
+
+    for label, src_dir, stem in cases:
+        for filename in required_files(src_dir, stem):
+            if not os.path.exists(filename):
+                missing.append((label, filename))
+
+    if missing:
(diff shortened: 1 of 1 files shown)


ss2098:

MPI_Init_thread failed

This PR only adds preflight input checks to scripts/stagnant_lid_adapt_compare_plot.py.

Local validation passed with:

python -m py_compile scripts/stagnant_lid_adapt_compare_plot.py
pixi run python scripts/stagnant_lid_adapt_compare_plot.py

With missing local checkpoint inputs, the script now exits cleanly after listing the missing files instead of failing later with a PETSc/HDF5 traceback.

lmoresi (approved):
# Adversarial review

We checked the `required_files` pattern against the actual writer: it matches `read_timestep`'s legacy path `{stem}.mesh.{name}.{index:05}.h5` (`discretisation_mesh_variables.py:1197`) and the `Mesh(...)` load path exactly. Live run on a machine missing only the adapted-case outputs: exits 1 and flags precisely those files, nothing else. Deferring `import pyvista` behind the check is sensible (E402 nit only).

Attacks that failed: filename-mismatch attack (pattern equals the writer's); deprecated-pattern sweep (0 hits); py_compile clean.

The failing `test` check was the known runner flake (#461), rerun. Merging once green.

PR #482, opened 2026-07-29
Fix RegionalSphericalBox bounds helper

ss2098:
This PR fixes a NameError in RegionalSphericalBox().

RegionalSphericalBox() passes sphere_return_coords_to_bounds to Mesh(), but that helper was not defined in src/underworld3/meshing/geographic.py. This adds the local bounds helper used to keep coordinates inside the spherical shell bounds.

Validation:
- python -m py_compile src/underworld3/meshing/geographic.py
- pixi run python docs/examples/tutorials/geographic/Ex_Spherical_Meshes.py

The example now completes the regional spherical mesh section and gravity check without the previous NameError.

### DIFF
--- a/src/underworld3/meshing/geographic.py
+++ b/src/underworld3/meshing/geographic.py
@@ -359,6 +359,22 @@ def RegionalSphericalBox(
 
         return
 
+    def sphere_return_coords_to_bounds(coords):
+        Rsq = coords[:, 0] ** 2 + coords[:, 1] ** 2 + coords[:, 2] ** 2
+
+        outside = Rsq > radiusOuter**2
+        inside = Rsq < radiusInner**2
+
+        coords[outside, :] *= (
+            0.99 * radiusOuter / np.sqrt(Rsq[outside].reshape(-1, 1))
+        )
+        coords[inside, :] *= (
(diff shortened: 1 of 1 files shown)


lmoresi (approved):
# Adversarial review

Confirmed this fixes a real, live crash: development's `geographic.py:376` references `sphere_return_coords_to_bounds`, which is never defined — `RegionalSphericalBox(...)` raises `NameError` at construction. The helper is verbatim from `spherical.py:1272`; we probed the radial clamp numerically (out-of-range radii clamped into [r_i, r_o]; in-range points bit-identical). No behavior change for existing callers — they were all broken.

Two non-blocking follow-ups we will track separately:
- Only the radial bounds are enforced. A RegionalSphericalBox has four lateral faces; particles escaping through them are not returned. The shell meshes this helper was copied from have no lateral faces, so the copy is incomplete for this geometry.
- No test: existing coverage (`test_0800_optional_modules.py:187`) is gated on cartopy AND pyproj, which is exactly why this NameError shipped. The constructor needs only gmsh — an ungated construction test would have caught it.

Merging. Thanks @ss2098.

Issue #449, opened 2026-07-28
3D nested P1 MG prolongation has O(1) wrong entries (nested_prolongation_from_dms), invisible to its own test


lmoresi:
## Summary

The vertex-level nested MG prolongation recorded by `mesh.adapt` (landed in #433) produces **O(1) wrong values at 1–2 fine vertices per refinement pass in 3D**. The affected vertices lie on no coarse edge, which is exactly the set `tests/test_0753_nested_mg_prolongation.py::test_reproduces_an_arbitrary_coarse_field` skips — so it merged undetected.

## Root cause

`underworld3/utilities/nvb.py::nested_prolongation_from_dms` resolves in rounds:

1. exact match against coarse vertices and coarse edge midpoints;
2. **rounds 2+**: any still-unresolved fine vertex equal to the average of two already-resolved fine neighbours takes the average of their weights.

Round 2 is inference, not topology — the function's own comment says so ("being fine-mesh neighbours does not by itself make the segment an edge of the refinement tree").

In 3D the Maubach split bisects `(tuple[0], tuple[tag])` and creates edges from the new midpoint `m` to the two off-refedge vertices. Those edges run through the coarse tet's **interior**, so their midpoints are not at any coarse edge midpoint and are not matched in round 1. Round 2 then assigns them `1/2, 1/2` on a coarse edge that does not contain them.

## Evidence

Band-refined 3D box (`UnstructuredSimplexBox(cellSize=0.5, refinement=1)`, `adapt(metric, max_levels=2)`), random coarse nodal field, compared against an independent reference (containing cell by geometric point location, value by the closed-form simplex Lagrange basis — deliberately not `uw.function.evaluate`, cf #432):

| generation | disagreeing nodes | \|vertex transfer − truth\| | \|cell-based transfer − truth\| | lie on a coarse edge |
|---|---|---|---|---|
| 2 | 2 / 1004 | **1.64** | 5.3e-16 | 0 / 2 |
| 3 | 1 / 1608 | **1.47** | 0.0 | 0 / 1 |
| 4 | 1 / 2136 | **2.31** | 4.4e-16 | 0 / 1 |

Generations 0, 1 and 5 are identical between the two constructions. 2D is unaffected (identical at every generation).

The comparison does not assume which construction is right: both are measured against the same independent reference.

## Why the test missed it

[...]

PR #488, opened 2026-07-31
Adapt-on-top: closure-free edge_split engine, reconnection repair, and interface-pinned relaxation

lmoresi:
Three capabilities for locally-refined (adapt-on-top) meshes, plus the design note
recording what was measured. All parallel, all tested at np=2/3/4.

## What lands

**`mesh.adapt(engine="edge_split")`** — longest-edge refinement with **no conforming
closure**. Splitting an edge divides every incident cell at the same new vertex, so
there is no hanging node to repair and refinement cannot escape the marked region:
the band hugs the feature instead of a bounded halo around it. No new topology code
— it drives the compiled `uwnvb_bisect` transform that NVB already uses, so it
inherits star-forest propagation, co-partitioning, labels and coordinates.
Bit-confluent: identical mesh at np=1/2/3/4, verified to 56k cells.

Marking is on the cell **diameter**, not `(d! V)^(1/d)`. The volume proxy reported
the target met while the mesh was **3.2x coarser** across the feature.

**`mesh.adapt(..., repair=True)`** — a reconnection (Lawson flip) pass after each
generation. 2-D only; raises rather than silently doing nothing elsewhere.

The acceptance criterion is **not Delaunay**, although these are Lawson flips.
Delaunay maximises the *minimum* angle and says nothing about the maximum, while
the P1 interpolation bound depends on the maximum (Babuska-Aziz). Measured:
flipping a gmsh-generated mesh towards Delaunay *raised* the 99th-percentile
maximum angle from 126.8 to 129.3 degrees, because gmsh optimises element shape
rather than the empty-circle property. Since every UW3 mesh starts from gmsh, a
pass that can degrade one is unusable. Gating on the angle makes it monotone.

Worth it on a poor base — 99th-pct max angle 156 -> 115 degrees on an
aspect-ratio-4 grid, slivers below q=0.1 3.84% -> 0.00% on a non-Delaunay one. On
a clean gmsh base it moves 124.7 -> 120.5 and the error not at all.

Opt-in, because it gives up the one property `edge_split` has: which cavities may
be flipped depends on where the partitioner cut, so the repaired mesh is **not
partition-independent**. Conformity, orientation, volume, labels and the
star-forest stay exact at every rank count.

**`mesh.relax(pin_bands=...)`** — hold an interface while relaxing everything else.
Relaxation and interface-tracking refinement fight: the MMPDE mover optimises shape
against an equilateral reference and knows nothing about where the material
changes, so it slides the small cells refinement placed on an interface off it.
Measured on a step-edged fault, manufactured stress across the interface **+77%**.
Pinning the band leaves that unchanged to five decimal places while the mover still
reshapes the rest of the domain.

## Implementation notes worth a reviewer's attention

A flip is **not** expressible as a `DMPlexTransform` — a child's cone may only
reference its own parent's closure, and a flip's output cells use the other
parent's apex. So the DM is rebuilt. It is rebuilt **on the same point chart**: a
2-D flip adds and removes no points, so preserving the numbering lets the point
star-forest transfer verbatim, labels transfer by point id and coordinates transfer
unchanged. That removes the whole reconstruct-the-SF-by-matching-seam-coordinates
stage. Surgery on the source DM is impossible: `DMPlexSymmetrize` refuses to run on
a plex that already has supports and nothing outside `DMDestroy` frees them.

Exact predicates are a **static filter that declines when it cannot resolve a
sign**, not adaptive precision. Declining is always safe because a flip is an
optimisation, never a requirement.

## Tests

22 serial, 9 parallel at np=2/3/4. 149 adapt/relax/smoothing regression tests pass;
style gates clean.

The load-bearing ones are the controls: parallel confluence (three of the four
`edge_split` defects during development were invisible serially), volume
conservation rather than orientation (checking that new cells are positively
oriented is worthless when they were built anticlockwise by construction), and the
maximum-angle assertion, which is what caught the Delaunay criterion.

## Companion

Skills documentation is split out to #489 — it documents this API, so it should
merge **after** this.


lmoresi:
## Adversarial review

We reviewed this branch against itself before marking it ready. Two findings were
real enough to fix in `4b041f7c`; the rest are recorded so a reviewer does not have
to rediscover them, and so the ones we chose to live with are choices rather than
oversights.

### Fixed in this branch

**1. `_orient2d` invented a sign for collinear input.** The static filter reduces
to `0 >= 0` whenever both products vanish, and the code then returned `-1` — a
confident "clockwise". That fires for *any* axis-aligned collinear triple, not just
coincident points, which is an ordinary configuration on a structured mesh. The
caller declined the flip either way, so no mesh was ever corrupted; but a predicate
whose whole contract is "report a sign only when justified" was reporting one it
could not justify, and it is private precisely so the next caller can trust it.
Now returns `UNCERTAIN`, with a regression test over coincident, x-collinear and
y-collinear input plus the unambiguous cases.

**2. `pin_bands` had no parallel test.** Charter §11. It worked, but "works" was
unverified at np>1 for the case that matters: a pinned vertex that is a
star-forest *leaf*, owned by another rank, which a rank-local pin would let the
owner move. Added `ptest_0845`, asserting partition-independence of the pinned set
(compared by **coordinate**, not count — a shared vertex is held by every rank on
the seam and a count would double-count it and mask exactly this defect), that
shared pinned vertices do not move, and that the domain boundary stays pinned.
np=2 and np=3.

### Accepted, with reasons

**`repair=True` gives up bit-confluence.** This is the sharpest trade in the PR.
`edge_split` alone produces the same mesh at any rank count; repair does not,
because which cavities may be flipped depends on where the partitioner cut.
Everything else stays exact. We made it opt-in rather than papering over it. A
reviewer who thinks confluence is non-negotiable should say so — the alternative is
cross-rank cavity handling, which is a substantially larger job and which the
measured seam cost (0.9–3.5 % of repair sites at 56k cells, halving with each
halving of the target size) does not currently justify.

**Under a frozen seam the 99th-percentile angle recovers but the absolute maximum
does not** — 148° vs 123° serial on a fault case. A few worst cells sit on the seam
and are exactly the untouchable ones. Documented, not fixed.

**`repair`'s objective is misaligned with interface artefacts.** It gates on the
maximum angle, which is indifferent to where a material interface runs, and on a
step-edged fault it made the worst *leaking* cell worse (3.37 → 3.97) while
improving shape. A leak-aware gate would need the material field passed into the
flip predicate — a real API change, deliberately not smuggled in here.

**`adapt(repair=True, relax=True)` is a footgun.** The end-relax runs *after* the
last repair, unpinned, and that is the combination measured to spread the
interface artefact by 77 %. There is no warning. Arguably `adapt` should thread
`pin_bands` through, or refuse the combination; we did not want to guess the API
in this PR.

**Performance.** `flip_to_reduce_max_angle` rebuilds the DM once per sweep in pure
Python, and `label_interface_band` walks every cell's transitive closure once per
halo ring. Fine at the sizes tested, not obviously fine at production scale.
Neither is on a hot path today because both are opt-in.

**`edge_split._cells_on_edge` is still unused by the engine.** We fixed its 2-D
dimension bug (it applied the 3-D edge→face→cell walk in both dimensions and
returned an empty set, silently) because the repair work needs it, but nothing in
`src/` calls it yet.

### Where the numbers came from, and what we got wrong getting them

Several figures in the description are revisions of earlier, wrong ones. Recording
this because it bears on how much weight to put on the rest.

- The "flips remove every sliver" figure originally quoted was the **centroid**
  engine's row, not edge-split's.
- The ~20 % error gain first attributed to reconnection was a **placement** effect:
  the prototype flipped *inside* the refinement loop, so the arms had different
  point sets. Isolated properly, connectivity alone is worth ≤3 %.
- A reported "1–6 % of the leak escapes the fault" was a **percent-format bug** in
  our own reporting; the real figure is 27 % beyond d=0.03 and 85 % beyond d=0.01.
- We predicted a step-edged margin would converge at N^-0.5 and it measures
  N^-1.32; the scaling argument paired the right leak law with the wrong cost model.
- We predicted the max-angle repair pass would act as a discrete alignment operator
  for a step edge. It barely does — which is what turned the leak-aware gate from a
  nice-to-have into the identified next step.

The controls that caught these are in the tests: parallel confluence, volume
conservation rather than orientation, and asserting the quantity the method claims
to improve rather than a proxy. The idempotence check notably does **not** catch a
wrong objective — an inverted criterion is idempotent too, which is exactly how the
Delaunay criterion passed while degrading the mesh.

lmoresi:
# Adversarial review — live probes at head 4b041f7c (isolated worktree, own build)

Three findings block merge:

1. **CRITICAL** — `discretisation_mesh.py:6782` (`label_interface_band`): `if not pinned: raise ValueError` is rank-local, no allreduce. Confirmed live: np=4 with the surface confined to one corner → rank 2 raises "no cell is cut by distance == 0.0" while ranks 0/1/3 label fine; the subsequent `relax(pin_bands=[surface])` hits PETSc error 98 on rank 2 and hangs >300 s (killed). Any band missing one rank's subdomain kills `pin_bands` at np>1. The ptests never see it — their diagonal surface crosses every subdomain at np≤3. Fix: allreduce the pinned count; raise collectively, otherwise create the (possibly locally empty) label on every rank.
2. **MAJOR** — `_adapt_nested` edge_split branch (~L829-907): `level_dms.append(current_dm)` per engine pass with `n_pass = 8·dim·max_levels`. Measured on the same metric/max_levels: nvb tail 5 levels vs edge_split 8, the last two at 2902 vs 2922 cells (0.7% apart) — near-duplicate MG levels. Needs a level-per-h-doubling gate (or dedup) before wrapping intermediates.
3. **MAJOR (latent)** — same branch: `if centroids.shape[0]: M = eval_metric(...)` skips the metric evaluation on a zero-cell rank, but for a field/expression metric `eval_metric` wraps `global_evaluate` — collective at np>1 → deadlock on any empty rank. The nvb branch calls it unconditionally; this is the exact failure class ptest_0843's docstring claims fixed.

Minor: (4) `ptest_0845::test_pinned_set_is_partition_independent` is a tautology — `union` comes from allgather so `MAX(len(union)) == len(union)` always, and `SERIAL_PINNED_COORDS = None` is dead; the named property is never compared to a serial reference. (5) `label_interface_band` reuses an existing label without clearing — confirmed: offset=0 → 42 pins, then offset=0.15 same default name → 59 (union); repeated `relax(pin_bands=...)` in a time loop grows the frozen set monotonically. (6) `adapt(adapter="mmg", repair=True)` silently ignores `repair` (mmg early-return precedes the repair validation) — contradicts the PR's own refuse-don't-ignore rule. (7) `test_0843::test_conforming_and_diameter_target_met` is near-true-by-construction (`_refine_to` marks on the same `cell_diameters` it asserts on); the volume-proxy regression is never exercised through `mesh.adapt()`. (8) `len(tail) >= 3` is vacuous — base refinement=2 alone yields 3 coarse levels. (9) Doc nits: design-doc header still says "No src/ change yet"; not in any toctree; `edge_split.py` docstring says the flip pass is "not yet done" while `reconnect.py` ships in this PR; serial constants 104/412/7 are gmsh-version-coupled.

Attacks that failed: serial suites 18/18 in 25 s; ptest_0843 np=2 and np=3 pass against the serial 412-cell/7-pass reference; ptest_0844 np=2/np=3 4 passed each (shared-point cones untouched, chart/area invariant, solve on repaired mesh converges); ptest_0845 np=2 3 passed; no bare exception swallows (the four narrow SF-getGraph guards carry rationale comments); all `np.empty` buffers fully written; SF/label bookkeeping matches the proven C form and is pinned by the shared-cone postcondition test; test_0844's oracle is independent (arccos-degrees vs cosine gate) with real negative controls; test-merge into current development is clean.

Holding both this and #489 until findings 1–3 are fixed; 1 is a hang on the exact workflow the #489 skills then recommend.

lmoresi:
# Re-review of PR #488 — adapt-on-top: edge_split engine, reconnection repair, interface-pinned relaxation

Head reviewed: `6d494db1` (fetched as `pr488-review-head`). Previous review 2026-08-02 at `4b041f7c` gated on three findings. Worktree: `.claude/worktrees/r488-rereview`, branch `review/r488-ci-fixes` carries three candidate commits (NOT pushed).

## Verdict

**Still gated — but the distance to merge is four small commits, and we have written them.** Of the three 2026-08-02 blockers, one (MG level inflation) is genuinely fixed, one (the collective raise) is **still live and reproduced hanging at np=4**, and one (empty-rank eval_metric) is untouched. On top of that we found and fixed a **new parallel correctness defect** in the same subsystem: the pinned band is partition-dependent, and at np=4 the mover moves "pinned" vertices by 4e-3. The two CI failures are diagnosed and fixed; neither is what it looked like.

## The 2026-08-02 blockers

### 1. CRITICAL — rank-local raise in `label_interface_band`: NOT CLEARED at head; fixed in candidate commit `90cfc578`

Reproduced live at head: np=4, surface confined to one rank's corner, `mesh.relax(pin_bands=[surface])`. Rank 2 raised the rank-local `ValueError("no cell is cut…")` while its peers entered the collective mover; the job hung the full 300 s until mpirun's timeout killed it. Same behaviour we gated on. Commit 6949d48e ("collective error paths") collectivised `line_cut.py` thoroughly but never touched this raise (`discretisation_mesh.py:6837` at head).

Fix (committed): band emptiness is an `allreduce`; only a GLOBALLY empty band raises, on every rank; the label is created on every rank including locally-empty ones. Verified np=4: corner band relaxes and all ranks pass a barrier; `offset=5.0` raises `ValueError` on all four ranks and passes the barrier.

### 2. MAJOR — per-pass MG level inflation: CLEARED

Commit `e7c32d47` (one level per doubling, `mg_coarsening_ratio`, composed transfers). Measured on an edge_split adapt (base 0.25/refinement 1, band metric to h=0.015): **5 levels, adjacent 5th-percentile-h ratios 2.00 / 2.05 / 2.00 / 2.31** — no near-duplicate pair (the gated run had 8 levels with the last two 0.7 % apart). The updated contract tests pass: `test_0836` + `test_0840` = 25 passed.

### 3. LATENT — empty-rank `eval_metric` collective skip: NOT REMEDIATED (still latent)

Code-verified at head: the guard pattern survives at three sites in `_adapt_nested` (`discretisation_mesh.py` ~8009 nvb-native `if cur_h.size:`, ~8094 edge_split `if centroids.shape[0]:`, ~8247 sbr — which `break`s out of the level loop entirely). At np>1 with a field/expression metric, `eval_metric` is `global_evaluate` (swarm-migration, collective); a cell-less rank skips it while its peers enter.

Probing `global_evaluate` directly with an empty point set on one rank deadlocked — but so did our non-empty control with rank-distinct off-rank points, so that standalone probe cannot cleanly attribute the hang and we do not lean on it. The real path (np=4 `mesh.adapt(field_metric, max_levels=2)`, every rank owning cells) works. The finding stays LATENT exactly as rated in August: it needs a rank with zero cells. We did not fix it; it should be a tracked follow-up (the sbr-path `break` is the worst of the three — that rank leaves the loop for good).

## NEW finding (this review)

### 4. MAJOR — the pinned band is partition-DEPENDENT; pinned vertices move at np=4. Fixed in candidate commit `217b0a5c`

`label_interface_band`'s docstring claims "every rank labels its own copy of a shared vertex identically and the result does not depend on the partition". False: the straddle test and the halo ring both walk rank-LOCAL cells, and cells are partitioned disjointly, so a shared vertex whose cut (or ring) cell lives on the neighbour rank is pinned there but not on its OWNER. The owner moves it; the neighbour's pinned copy follows through the SF.

Measured (np=4, the `ptest_0845` fixture itself): pinned leaves at (0.835, 0.490) and (0.690, 0.542) moved **4.2e-3 and 1.9e-3**. np=2/3 pass on partition luck — and the ptest header documents np=2 and 3 only, stopping one rank short of the case that fails. Three test defects hid it:

* `test_pinned_set_is_partition_independent` is vacuous — it allgathers the union on every rank, then compares the union's size to the MAX over ranks of that same size. Identical by construction; cannot fail.
* `_pinned_indices` calls `getStratumIS(1)` unguarded — segfault on an empty DMLabel (#291), which is what a band-less rank now holds.
* `moved[idx].max()` on the band-less rank's empty `idx` raises rank-locally and desyncs the collective asserts (observed as a 400 s timeout).

Fix (committed): the pinned set is synchronised across ranks by rounded coordinate after the core band and after each halo ring (`uw.mpi.size == 1` short-circuits); the two ptest helpers guarded. `ptest_0845` now passes at np=2, 3 AND 4; serial `test_0845` unchanged (4 passed); the corner-band probe still clean. The vacuous first test is left as-is — flagging it here; replacing it with a genuine cross-np fingerprint is follow-up test work, and the moved-vertex test now covers the defect it missed.

## The two CI failures (both fixed in candidate commit `03cc18a7`)

### `test_0753…[3d]` — the #449 guard is a gmsh lottery, and the bug is NOT fixed

CI reports "3-D interior-vertex rows are now exact — appears FIXED". It is not: at head, on macOS gmsh 4.15.1, the defect is measurably live — **1 wrong row of 842 interior rows** — and the `TODO(BUG)` in `nvb.nested_prolongation` stands; no branch commit touches the 3-D parentage (only `e7c32d47` touches `nvb.py`, and only to add the TODO). Whether any interior vertex lands in the defective closure-cascade configuration depends on the gmsh-generated base mesh; CI's Linux PyPI wheel (open pin `>=4.13,<5`) produces none, so the positive `assert wrong` guard tripped — the exact fragility its own docstring attributed to strict xfail while reproducing it. Fix: the 3-D branch xfails (non-strict) when the defect manifests, passes quietly when the mesh doesn't exercise it; #449 remains tracked at the source. Do NOT flip to assert-exactness as the message instructs — that fails on every gmsh build that does produce the bad row (any current macOS dev box).

### `test_0842` FMG-vs-GAMG — a tolerance calibrated to the old hierarchy, plus a comparison that was vacuous all along

CI: `err/nrm = 1.2959e-5 / 18.715 = 6.9e-7` vs the asserted `1e-8`. Not an accuracy defect: `ksp_rtol=1e-8` is enforced in the PRECONDITIONED norm (norm type 1, measured), so the constant between declared reduction and nodal error depends on the PC — and `e7c32d47` legitimately changed the PC (2 recorded coarse levels where the per-generation contract gave 3). Locally the same solve gives 2.7e-10; the 1e-8 nodal bound was a calibration to the old hierarchy's margin, never a property of the method. New bound 1e-6 with the reasoning inline — a lost Dirichlet label or wrong transfer, the failures this test exists for, are O(1).

Worse, found on the way: **the "gamg" arm ran `pc_type=mg`** — both arms identical to 16 digits. `auto_inject_custom_mg` (pre-existing on development, not a branch regression) clobbers an explicit `preconditioner="gamg"` with the custom-P PCMG on any adapt child, so the FMG-vs-GAMG parity assertion compared FMG to itself, and users could not opt out of the pickup. Fixed in `custom_mg.py`: an explicit gamg choice now wins over the opportunistic pickup. The comparison is real again — **fmg 6 its vs gamg 21 its**, parity assert holds; note the true gamg arm's err/nrm is 9.3e-9, inside the old 1e-8 bound by only 7 % even locally, confirming the bound was a margin lottery. `test_0842` passes serial and np=2.

## Delta review of the new capability commits (not covered 2026-08-02)

* **Collectivity**: `line_cut.py` is exemplary — every data-dependent raise sits behind an allreduce, with comments naming the rank-local trap. `edge_split.py` has no raises; its SF reconcile handles the unpopulated-SF case symmetrically. `reconnect.py`'s loops vote collectively ("a rank with nothing to flip/delete still has to vote"); one latent rank-local invariant guard at `reconnect.py:781` ("a shared point was deleted") — defensive, fires only on an internal contract violation, would hang peers if it ever fires. Acceptable as an assert-class guard; noting it.
* **Silent fallbacks**: the `except` sites in the new modules are documented and bounded (unpopulated-SF early-outs; `snap_frac` sweep recording per-tolerance failures as data). Nothing swallowed silently.
* **Visualisation**: raises are symmetric argument validation; the empty-stratum segfault case on ranks owning no part of a surface is explicitly guarded.
* **Serial tests**: `test_0843 + test_0844_line_cut + test_0844_reconnect_repair + test_0845 + test_0846` = **66 passed** in 52 s.
* **Parallel ptests** (documented sizes and one past them): `ptest_0843` np=3: 2 passed; np=4: 2 passed. `ptest_0844_line_cut` np=4: 16 passed; np=3: 16 passed. `ptest_0844_reconnect` np=3: 7 passed; np=4: 7 passed. `ptest_0845` np=2/3: 3 passed — and np=4 segfaulted, then hung, then failed genuinely (finding 4 above); passes at 2/3/4 with the candidate commits.

## Gate

**UNFINISHED — full serial `pytest -m "level_1 and tier_a"` did not produce a verified summary.** Four attempts were made; the machine was saturated the whole time by three other sessions' long PETSc jobs (fault-split-node, regime-diagram ×2, plus a concurrent r500-review pytest gate), which slowed a 10-minute suite past our command timeouts (one attempt sampled mid-run was healthily grinding a Stokes SNES solve at 99 % CPU after 50 CPU-minutes — working, not hung). One completed summary was captured — `555 passed, 17 skipped, 1355 deselected, 1 xfailed in 601.83s`, zero failures — but its warnings section cites the (since-deleted) r500-review worktree's site-packages and the wrapper exited 1 despite zero failures, so we cannot cleanly attribute it to our worktree and do not claim it. Treat the full gate as TO BE RE-RUN on a quiet machine before merge.

What we DID verify green in this worktree, per-file, with our three candidate commits applied (pass counts, all exit 0):

* `test_0753` 19 passed + 1 xfailed (the [3d] #449 xfail, locally manifest)
* `test_0842` 2 passed serial, 2 passed np=2
* `test_0836` + `test_0840` (level-count contract) 25 passed
* new capability serial files (0843, 0844×2, 0845, 0846) 66 passed
* MG stakeholders of the custom_mg change (`test_1014`, `test_1016`, `test_1017`) 21 passed
* parallel: ptest_0843 np=3/4 (2+2), ptest_0844_line_cut np=3/4 (16+16), ptest_0844_reconnect np=3/4 (7+7), ptest_0845 np=2/3/4 (3 each)

## MERGE BLOCKERS

1. **Take (or re-derive) the collective-emptiness fix for `label_interface_band`** (`90cfc578`). Without it, `relax(pin_bands=…)` hangs at np=4 on any surface that misses one rank — reproduced at head.
2. **Take the partition-independence fix for the pinned band** (`217b0a5c`), including the ptest guards. Without it, "pinned" vertices move (4.2e-3 measured) whenever a band vertex's owner rank does not see the cut cell — silent wrong answers, worse than the hang.
3. **Take the CI fixes** (`03cc18a7`): the 0753[3d] non-strict xfail and the 0842 bound + gamg-respect fix. CI cannot go green without them, and #502 inherits both failures.
4. **Re-run the full tier-A gate on a quiet machine** — our per-file coverage is green (see Gate) but the whole-suite number is unverified; the machine was contended by three other sessions throughout this review.

## Cleared

* MG level inflation (finding 2, 2026-08-02): fixed by `e7c32d47`/`32443ee5`; measured one level per doubling, no near-duplicates.
* The new capability code's collectivity discipline is genuinely good — `line_cut.py` in particular should be the template.

## Left open (tracked follow-ups, not blockers)

* Empty-rank `eval_metric` collective skip (finding 3, 2026-08-02) — still latent at three sites; the sbr-path `break` is the worst variant.
* `reconnect.py:781` rank-local invariant guard.
* `test_pinned_set_is_partition_independent` is vacuous and should be rewritten as a cross-np fingerprint.
* `global_evaluate` hung in a standalone probe with rank-distinct off-domain points (control included) — predates this branch, deserves its own scrutiny.
* nvb 3-D interior-vertex prolongation (#449) remains live: 1/842 rows wrong at head on gmsh 4.15.1.

lmoresi:
Response push 304d2d06..469b5e53 — all four merge blockers from the re-review are resolved, plus development merged in and one latent defect the composition exposed:

1. **Collective emptiness for `label_interface_band`** (462eb122): band emptiness is an allreduce; only a globally empty band raises, on every rank. The np=4 corner-band hang reproduced at the previous head is gone (relaxes clean; `offset=5.0` raises on all four ranks and passes a barrier).
2. **Partition-independent pinned band** (debeec42): the pinned set is synchronised by rounded coordinate after the core band and each halo ring; the two ptest helpers guarded (#291-class empty-label segfault, empty-`max()` desync). ptest_0845 now passes at np=2/3/4 — previously pinned vertices moved 4.2e-3 at np=4.
3. **The two CI failures** (304d2d06): test_0753[3d] is a non-strict xfail — the #449 defect is still live (1/842 rows on macOS gmsh 4.15.1) and the "appears FIXED" CI verdict was a gmsh-mesh lottery; test_0842's bound story continues below.
4. **Development merged** (7cbb5984, zero textual conflicts) — and the semantic composition was attacked per house rule, which found two real problems the clean merge hid:
   - **ebe314cc**: the branch's explicit-gamg-wins gate was merge-planted inside #471's new `build_transfers()` (a 2-tuple contract) with its old bare `return`, so an explicit gamg choice CRASHED at the unpack instead of opting out. The gate now returns `(None, None)`, also stands down when `_pc_user_override` is latched, and the unpack is None-guarded everywhere.
   - **469b5e53**: with #471's bundle (gmres/4 smoothers — a non-stationary preconditioner) under the scalar solver's plain left-preconditioned gmres outer, the recurrence norm fell to 1e-11 while the TRUE residual stalled at 2.4e-6, insensitive to ksp_rtol. `_ensure_flexible_outer` now upgrades only the framework-default outer gmres → fgmres, only when the effective smoother is Krylov; measured err/nrm 1e-6 → **1e-12**. test_0842's final form asserts each arm's actual PC type (the FMG-vs-itself comparison can never return) and a 1e-7 bound that catches the rtol-insensitive stall signature with margin.

Gate on the merged head: **579 passed, 0 failed, 17 skipped, 2 xfailed** (both pre-existing) in 11:08, machine quiet. Follow-ups tracked: #512 (empty-rank eval_metric skip, reconnect:781 guard, vacuous partition test), #513 (pre-existing global_evaluate deadlock), #449 updated with the lottery status.

lmoresi:
Final landing wave (2d6d1208 + merge 379d2fcc):

- **The Linux CI segfault is explained and disarmed.** It was issue #492 detonating, not reconnect: test_0842's second MeshVariable creation (after the fmg solve) rebuilds child.dm and destroys the old DM at refcount 1, leaving the coarse/fine links dangling — deterministic segfault two files downstream on Linux, allocator-lucky on macOS (Guard Malloc clean because the page is genuinely freed and reused, not guard-hit). Both variables are now created before any solve, with a comment naming the trap; the underlying bug stays open on #492 with the probe evidence. The place-surface commits are exonerated (their head passed the same Linux batch).
- `scripts/test.sh` now exports `PYTHONFAULTHANDLER=1` — the next hard crash prints a stack instead of "core dumped".
- **Development re-merged at head**: #503's Null_Boundary removal — the blanket-label fixture now builds its own blanket label (self-contained, control assert preserved); #515's bundle-owned fgmres pairing fully covers what our `_ensure_flexible_outer` did (verified on the scalar custom-P route: fmg arm runs pc=mg with ksp=fgmres, err/nrm 7.5e-10), so our helper is removed outright — zero duplicate mechanism. The explicit-gamg gate survived semantically (real gamg, 23 its, err/nrm 2.0e-9, no crash).

Verification: exact CI batch (`tests/test_08*py`, one process) 527 passed / 0 failed; full tier-a gate 579 passed / 0 failed (8:41); test_0842 serial + np2; MG stakeholders 39 passed; adapt/reconnect serial 66 passed.

lmoresi:
Status: this PR's core content landed on development via #510 (merge 5e1a9030) while the final CI round here was in flight, including that branch's own fixes for the 0753 tripwire (skip-based), the vertex-blanket fixture, and a deterministic plotter close. However, development as merged still LACKS five fixes from the re-review that exist only on this branch — most seriously the collective band-emptiness fix (`label_interface_band`'s rank-local raise = the measured np=4 hang) and the partition-independent pinned band (pinned vertices moved 4.2e-3 at np=4). Those are being extracted into a focused PR onto current development: collective emptiness, pinned-band partition independence + ptest guards, the explicit-gamg gate + None discipline, the test_0842 final form with the #492 disarm, and PYTHONFAULTHANDLER in the CI harness.

The two place-surface commits are parked on a separate branch (`feature/place-surface` locally; also still on this PR's head) — they are exonerated of the CI segfault (their head passed the Linux batch) and should go up as their own reviewed PR.

This PR closes as superseded once the fixes PR merges.

lmoresi:
Superseded end-to-end: the base capability landed via #510/#511 (the untangling merges), the place-surface work landed reconciled via #511 and its successors, and the five adversarial-re-review fixes this branch uniquely carried landed via #530 (bc53cfc4). Everything this PR contained is now on development. The review record lives here: the 2026-08-02 gating review, the 2026-08-09 re-review (three blockers + the partition-dependent pinned band), and the #492 CI-segfault forensics.

Issue #478, opened 2026-07-29
Geometric multigrid is unreachable for every single-field solver, even when explicitly requested (the #276 lockout)


lmoresi:
`preconditioner="auto"` / `"fmg"` is documented as giving geometric multigrid whenever the mesh carries a refinement hierarchy. For the Stokes velocity block it does. For every **single-field** solver — `Poisson`, `Darcy`, `AdvDiffusionSLCN`, `Projection`, `SNES_Vector` — it never does, at any refinement depth, even when asked for explicitly.

## Measured, off the live PC after setup

| solver | mesh | `preconditioner=` | live PC |
|---|---|---|---|
| Stokes | no hierarchy | auto | `gamg` |
| Stokes | refinement=1 | auto | **`mg(2)`** |
| Stokes | refinement=2 | auto | **`mg(3)`** |
| Poisson | no hierarchy | auto | `gamg` |
| Poisson | refinement=1 | auto | `gamg` |
| Poisson | refinement=1 | **fmg (explicit)** | **`gamg`** + warning |
| Poisson | refinement=2 | **fmg (explicit)** | **`gamg`** + warning |

Repro: `~/+Simulations/rotated_pmat_audit/fmg_gate_audit.py`. It reads the live PC rather than the options database, because the gate leaves no trace in the options — the bundle is simply never written.

## Cause

`SolverBaseClass._apply_preconditioner_options`:

```python
if want_fmg and prefix == "":     # prefix "" == single-field
    want_fmg = False
```

This is the #276 lockout, and it is not a bug: PETSc's `DMCreateInjection` fails at solve time with err62 (`Could not locate matching functional for injection`) on curved shells and some flat high-degree cases, so rather than let users hit that, the native route is closed for single-field solvers. #276 was closed on that basis, with `custom_mg.set_custom_fmg()` as the documented way around it.

## Why it is worth reopening as a routing question

The workaround is real but it is opt-in, per-solver, and requires the user to build a coarse-mesh tail by hand — for a mesh that **already carries one** as `mesh.dm_hierarchy`. So the situation is: the hierarchy exists, geometric MG would work on it, the machinery to use it exists, and the user gets algebraic multigrid unless they know to go and wire it up themselves. On a scalar Poisson over a refined box, custom-P geometric MG converges in 1–3 iterations against GAMG's 13 (measured in `custom_mg`'s own module docstring).

[...]

PR #489, opened 2026-07-31
skills: adapt engines, band pinning, FMG on adapt children — brought current; plasticity-solvers rewritten (#454)

### LINKED ISSUE #454: plasticity-solvers skill teaches the retired yield-homotopy doctrine — rewrite around Picard entry + grid sequencing, do not restore the API
The `plasticity-solvers` skill shipped in `.claude/skills/` presents `cm.enable_yield_homotopy()` as **"the recommended default for any hard-Min solve"** (SKILL.md:17, :21). That method does not exist:

```
$ git grep -c "enable_yield_homotopy" development -- src/
  (no match)
$ git grep -c "enable_yield_homotopy" feature/regime-diagram -- src/
  (no match)
```

Its headline code example raises `AttributeError`. The skill references it in six places (:17, :21, :42, :82, :102, :112, :123), including solver-option advice (`snes_linesearch_type="basic"`, an `snes_max_it` floor of 100) that describes machinery no longer present.

### Why this bites

The two skills disagree and the stale one is the one you land on. `nonlinear-solver` is correct and explicitly says so:

> (This supersedes the `enable_yield_homotopy()` in-SNES ramp still described in `plasticity-solvers`; that path is retained only as a dead-experiment record — use the multi-solve continuation above.)

But `plasticity-solvers` carries no such warning at its own top, and its description ("Reach for THIS first when a Drucker-Prager / yield-stress Stokes solve stalls...") is what routes you there. So a session that follows the descriptions gets the dead API and the retired in-SNES ramp strategy — which is not merely absent but **proven harmful**: `nonlinear-solver` records that ramping δ inside a single SNES solve diverges after ~2 iterations and grinds for ~2 hours even on the correct config.

The live entry point is:

```python
report = stokes.solve(homotopy=True)      # smooth mode + tangent + march
report["settled_delta"]
```

### Also stale in the same file

- It teaches only the sqrt soft-min law. The homotopy path now hardcodes the **power-mean** (`_yield_homotopy_control()` sets `yield_smoother="powermean"`), so the δ semantics the skill describes are not the ones the driver uses — δ≤1 for the power mean, where δ=1 is the harmonic mean, versus the sqrt family's δ up to 64. Following the skill's δ intuition on the live path gives silently wrong schedules.
- No mention of `powermean` anywhere in the file.

### Suggested fix

Either fold `plasticity-solvers` into `nonlinear-solver`, or reduce it to the yield-law maths and the tangent-per-model table (which are still correct and are what `nonlinear-solver` defers to it for), and delete the `enable_yield_homotopy` recipe entirely rather than leaving it as a "dead-experiment record" in the file a reader is routed to first.

Related: while auditing this I also noticed `constitutive_models.py:1044` (the `yield_smoother` property docstring) still states the power-mean sharpness is `s = 1/δ`, where the code at :1014 uses `s = 1/(δ + 0.001)`. Minor, but it is the number a reader would use to pick δ0.

lmoresi:
# PR #489 body draft (updated for the combined skills tidy)

**Title suggestion:** skills: parallel adapt engines, band pinning, FMG on adapt children — brought current, plus the #454 plasticity-solvers rewrite

Fixes #454.

## What this PR now carries

Two layers on the same branch, so one PR lands a consistent skill set:

### 1. The original three-file content (unchanged in substance)

- **adaptive-meshing**: the "PIN THE INTERFACE when you relax" section
  (`relax(pin_bands=...)`, `label_interface_band`, the +77 % / 1343→965
  straddling-cell measurements). Verified accurate against current development,
  merged as-is — #530's collective/partition-independent pinning only made it
  more true.
- **adapt-on-top-faults**: the engines section (`nvb` vs `edge_split`, `repair=`),
  band pinning, and band sizing (leak = −2 Cov(η, ε̇)).
- **nonlinear-solver**: the "FMG on an ADAPT-ON-TOP child" section (auto pickup of
  the mesh-owned custom-P tail, the #424 rbf cliff, repair vs the any-degree
  transfer).

### 2. Currency fixes (development moved since the branch)

Each correction, with its one-line rationale:

- **adapt-on-top-faults — rotated free-slip gotcha row + section**: the row said
  FMG under rotated free-slip is "FUNDAMENTAL, not a quick fix"; #467 fixed it —
  the rotated path now resolves hierarchies through the same
  `custom_mg.build_transfers` rule and picks up an adapt child's tail
  automatically (`rotated_bc._build_rotated_custom_Pl`). The plain-refined-mesh
  case still needs `set_custom_fmg`, and the skill now says exactly that.
- **adapt-on-top-faults — interface-alignment pointer**: the band-sizing section
  named "align the interface with element boundaries" as an exact fix with no way
  to do it; `place_sheet` / `place_thin_volume` / `remove_embedded`
  (`utilities/place_surface.py`, #517–#526) now exist, so it points at them.
- **nonlinear-solver — rotated free-slip claim**: "none of this is automatic
  under rotated free-slip" was actively steering users away from a path that now
  works (#467); corrected.
- **nonlinear-solver — MG level per generation → per DOUBLING of h**: #515 made
  `_adapt_nested` subsample the tail (`mg_coarsening_ratio=2.0` default;
  per-pass levels measured 2.3–7.3× slower). The "don't hand it a base-only
  tail" advice stands.
- **nonlinear-solver — the `refinement>=1` overstatement**: with `refinement=0`
  the tail is `[base] + intermediate doubling levels`, so the hierarchy starts at
  the base — deeper uniform tails help, but "no coarse grid" was not true.
- **nonlinear-solver — three capabilities that landed since the branch**:
  `preconditioner="gamg"` is respected on adapt children (#530 — previously
  silently clobbered, so the skill's own GAMG-probe advice was unfollowable);
  single-field `preconditioner="fmg"` over `mesh.dm_hierarchy` (#478/#534); and
  `solver.pc_fallbacks` as the observability hook for every degradation,
  including the #424 barycentric→rbf retry (#534).

### 3. The #454 rewrite of plasticity-solvers

The skill headlined `cm.enable_yield_homotopy()` — an API that does not exist in
`src/` — and taught the in-SNES δ-ramp doctrine in ~8 places. Per the issue's
rescope: the doctrine rested on a unit-scaling error, and the corrected-scaling
evidence runs against it (the δ-march never succeeded where a direct hard-Min
solve failed; direct is 4–5× faster where both work; ruling — regularise the
problem, not the solver).

Rewritten around what the campaigns measured:

- **Newton with the automatic Picard entry** (#441): a cold consistent-Newton
  start takes one Picard step automatically; Picard is an entry requirement, not
  an accelerator (front-loading it measured pure waste where Newton works).
- **Newton-first, rescue on failure OR stagnation** with `picard=N` /
  `consistent_jacobian="continuation"` — admissible uselessness never trips a
  FAIL reason, so failure-only rescue is not enough.
- **Grid sequencing** as the validated warm start for hard problems (coarse solve
  → linear-exact RBF transfer → fine; hand-rolled, since `-snes_grid_sequence`
  does not work on UW3 meshes).
- **The #475 substrate** (`yield_mode` / `yield_smoother` / `yield_anchor`, δ as
  a recompile-free constant) presented as a modelling choice, with the
  multi-solve-only discipline for any δ-march (never ramp inside one SNES solve).
- **`yield_continuation` described honestly** per open #473: the documented
  cold-start guarantee does not hold on a `Piecewise` yield stress, and the step
  control is effectively one-shot. The latched `snes_max_it` wart (#490) is not
  taught as a feature.
- **Floors post-#475**: `viscosity_min_rounding` keeps `shear_viscosity_min`
  differentiable at δ = 0, where the default rounding vanishes.
- Kept: the per-model tangent table (VEP-indefinite ruling), the
  confirm-you-are-running-Newton check, the SNESFAS ruling, the Picard-folklore
  footnote.

Sanity: `enable_yield_homotopy` — zero matches across all skill files;
`Null_Boundary` — zero matches.


lmoresi:
# Adversarial review

Every command/API/flag the skills teach exists in #488's diff or on development — `engine="edge_split"`, `repair=True` refusal semantics, `relax(pin_bands=[fault] / [(fault, 0.02)], pin_halo=)`, `label_interface_band(surface, offset, halo)`, `child._custom_mg_coarse_meshes` auto-pickup, `influence_function(profile="step")`, the custom_mg barycentric→rbf retry warning text, `set_custom_fmg(..., field_id=0)`, the `Mesh(dm, boundaries=...)` gotcha. No invented parameters, no stale-API teaching.

Two doc-level overclaims, both tied to #488 defects:
- "raises rather than silently doing nothing otherwise" is false for the `adapter="mmg", repair=True` corner (#488 finding 6).
- The skill teaches `child.relax(pin_bands=[fault])` for parallel fault models with no caveat, while #488's `label_interface_band` hangs whenever a rank's subdomain misses the band (#488 finding 1). Carry the warning until that is fixed.

Merge strictly after #488 (else the skills document nonexistent features). Holding with it.

PR #493, opened 2026-08-03
Stokes uu_G3 transposed tangent + TI un-frozen Picard: fix issue #457 at source

### LINKED ISSUE #457: TransverseIsotropicFlowModel: consistent (Newton) tangent inconsistent with the residual when anisotropy is active

lmoresi:
Fixes #457 at source. Two independent defects, both hidden by major-symmetric tangents until the transverse-isotropic Newton tangent — the first in-repo tangent without major symmetry — exposed them.

## 1. The Stokes velocity-block Jacobian was the major transpose of the true tangent

`sympy.derive_by_array` is dx-first, but `SNES_Stokes_SaddlePt` translated it with `permutedims((0,2,1,3))` — correct only under the dx-last convention that `petsc-jacobian-layout.md` wrongly recorded. The assembled `uu_G3` therefore held `dF1[gc,dg]/dL[fc,df]` in PETSc's `[fc,gc,df,dg]` slot. Frozen constitutive tensors and isotropic Newton terms are major-symmetric, so nothing in the suite could see it; the 2026-04-21 layout audit signed the Stokes path off against the mis-stated convention at the same time SNES_Vector was being fixed for the identical bug.

All `uu`, `up`, and natural-BC (including Nitsche G2/G3) blocks now use the explicit-index loops SNES_Vector and SNES_MultiComponent already use. The block-constraint boundary `G0` keeps its form — shape-protected, and `r(n⊗n)` is symmetric.

## 2. The TI c-tensor builds silently un-froze the Picard tangent

`TransverseIsotropicFlowModel._build_c_tensor` baked the unwrapped `.sym` contents of η₀/η₁ into the tensor, exposing their strain-rate dependence to `sympy.diff` — so "Picard" for TI was the (transposed) Newton derivative, never a frozen-coefficient tangent. It now bakes the wrapped Parameter atoms exactly like the isotropic path. `TransverseIsotropicVEPFlowModel` gets the same discipline plus a persistent `_eta1_yield_eff` UWexpression container (the `ViscoPlasticFlowModel` pattern) so the yield-limited weak-plane viscosity freezes under Picard and is recovered by `_jacobian_unwrap` under Newton.

## Measured (unpatched, this branch, on the #457 reproduction)

| quantity | before | after |
|---|---|---|
| Picard velocity-block asymmetry | 1.089e-01 | 4.0e-17 |
| Newton `-snes_test_jacobian` rel | 1.09e-01 | 1.3e-4 (FD-limited; isotropic control 1.06e-4) |
| Newton convergence | Picard pace / stall | 9 iterations |

## Tests

`tests/test_1066_stokes_jacobian_layout.py`, validated both ways (all three fail on the pre-fix build at their primary assertions — 0.313 rel FD mismatch, velocity-gradient dependence throughout — and pass on this branch):
1. Assembled `uu_G3` vs a central-difference oracle built from the residual flux, with a built-in guard that the oracle tangent is genuinely major-asymmetric (the test cannot pass by symmetry).
2. TI Picard `uu_G3` has zero velocity-gradient dependence (frozen) and exact major symmetry; Newton positive control.
3. TI-VEP c-tensor coefficient freezing with a Newton-unwrap positive control.

Full `level_1 and tier_a` gate: **554 passed, 0 failed**.

Docs: `petsc-jacobian-layout.md` corrected — dx-first convention, a warning recording the earlier mis-audit, the Stokes migration, and a ban on `permutedims` for new Jacobian code.

Out of scope, noted for the record: #463 (TI-VEP `viscosity` property discards the yield-limited η₁ — orthogonal property-level defect) and the isotropic VEP `_build_c_tensor`'s own `.sym` use (its flux path runs through `stress()`, not this tensor, and its tangents are symmetric either way; it should be tidied with #463).

### DIFF
--- a/docs/developer/subsystems/petsc-jacobian-layout.md
+++ b/docs/developer/subsystems/petsc-jacobian-layout.md
@@ -61,33 +61,38 @@ pairs like-kind indices.
 
 ## Sympy's native convention
 
-`sympy.derive_by_array(F, x)` returns an array with **`F`'s indices
-first, then `x`'s**:
+```{warning}
+An earlier revision of this page stated the opposite convention
+("F's indices first, then x's") and audited the Stokes `permutedims`
+path as correct on that basis. That was wrong — it is how the issue
+#457 transposed-tangent bug survived the 2026-04-21 audit.
+```
+
+`sympy.derive_by_array(F, x)` is **dx-FIRST**: the returned array has
+**`x`'s indices first, then `F`'s** (see the sympy documentation for
+`derive_by_array`; verified numerically against finite differences in
+the #457 investigation):
 
 ```python
 G3 = sympy.derive_by_array(F1, L)
 #    F1 shape (Nc, dim), L shape (Nc, dim)
 #  → G3 shape (Nc, dim, Nc, dim)
-#  → G3[i, j, k, l] = ∂F1[i, j] / ∂L[k, l]
-#  → index order [fc, df, gc, dg]
+#  → G3[k, l, i, j] = ∂F1[i, j] / ∂L[k, l]
+#  → index order [gc, dg, fc, df]
 ```
 
-This is the same mathematical object PETSc wants, but with the **two
-middle axes swapped** relative to PETSc's flat layout. Translating
-sympy → PETSc therefore needs the permutation **`(0, 2, 1, 3)`** — swap
-axes 1 and 2 — before flattening.
+Translating sympy → PETSc's `[fc, gc, df, dg]` therefore needs the
+permutation **`(2, 0, 3, 1)`** before flattening. The historically-used
+`(0, 2, 1, 3)` yields `[gc, fc, dg, df]` — the **major transpose**
+`∂F1[gc, dg]/∂L[fc, df]` in the `[fc, gc, df, dg]` slot. That is
+identical to the true tangent whenever the tangent has major symmetry
+(any frozen major-symmetric `C`, the isotropic Newton term
+`η'·(ε̂⊗ε̂)`, linear transverse isotropy) and silently wrong when it
+does not (the transverse-isotropic Newton term is `B⊗a` with `B ∦ a` —
+issue #457).
 
-For `g1` (3D, sympy order `[fc, gc, df]`) and `g2` (3D, sympy order
-`[fc, df, gc]`):
-
-| Array | Sympy shape after `derive_by_array` | Permutation | Final 2D shape |
-|-------|--------------------------------------|-------------|----------------|
-| `g0`  | `(Nc, Nc)` — already `[fc, gc]`       | none         | `(Nc, Nc)`     |
-| `g1`  | `(Nc, Nc, dim)` — `[fc, gc, df]`      | none         | `(Nc·Nc, dim)` |
-| `g2`  | `(Nc, dim, Nc)` — `[fc, df, gc]`      | `(0, 2, 1)`  | `(Nc·Nc, dim)` |
-| `g3`  | `(Nc, dim, Nc, dim)` — `[fc, df, gc, dg]` | `(0, 2, 1, 3)` | `(Nc·Nc, dim·dim)` |
-
-The row-major flatten of the 2D form matches PETSc's flat index exactly.
+Do not reach for `permutedims` at all — use the explicit construction
+below.
 
 ## Alternative: explicit construction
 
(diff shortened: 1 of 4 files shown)


lmoresi:
# Adversarial review — live probes at the PR head (isolated worktree, own build)

We attempted to break both fixes independently. Verdict: clean, two nits (response commit below).

**Findings**
1. MINOR — `petsc-jacobian-layout.md` claimed "no remaining in-repo `derive_by_array + permutedims` Jacobian path" while the block-constrained boundary `uu_G0` still used it (`cbc_permutation`, content `r·(n⊗n)` — symmetric and shape-protected, so numerically harmless). Fixed in response commit b0cc2405 by converting that site to the explicit loops; the claim is now literally true (0 live `permutedims` in the file). Retest: layout + constrained free-slip + constrained SolCx + Stokes, 25 passed.
2. MINOR — two G1 loops named the trial-side derivative index `df` where the layout doc uses `dg`. Cosmetic; renamed in response commit b0cc2405.
3. NOT-PR — `tests/parallel/test_1063_constrained_freeslip_parallel.py` fails at np=2 on BOTH pre-fix development and the PR build with byte-identical iso values (0.59828... vs golden 0.61945...; TI differs between builds only at 6e-13). Pre-existing golden/environment mismatch — filed as #495.

**Attacks that failed** (independently measured)
- 3D TI Newton, tilted director (1,2,3)/√14: symbolic FD-oracle on `uu_G3` rel 6.7e-11 (the transposed form scores 1.63e-1); PETSc `-snes_test_jacobian` 3D TI 7.9e-08 vs isotropic 3D control 6.9e-08 — same order.
- 2D reproduction: pre-fix `-snes_test_jacobian` 0.16–0.20 → PR build 5.5e-07 (sharper than the PR body's 1.3e-4).
- sympy convention independently re-verified dx-FIRST — the doc correction and transpose diagnosis stand.
- Flatten orders of the reshaped `up_*`/Nitsche blocks verified against the fe.c flat-index formula; no in-repo solver puts p/grad-p in F0, so the old wrong-ordered-but-zero blocks never fired; `pu_*` blocks are correct because Nc_f=1 makes dx-first coincide.
- Navier-Stokes: F0's u-dependence is diagonal (ρc₀/Δt·I) — the old transpose was invisible there; no in-repo asymmetric G0; test_0610 passes. A user-supplied asymmetric u-dependent body force previously got a transposed G0 — this fix improves that case; converged answers unchanged.
- TI-VEP container: hybrid mode keeps the container out of the ETD tensor; per-instance `_uw_id` (no cross-model collision); yield_stress updates flow through without a stale tangent; `is_constant_expr` = False (the JIT will not freeze it as a constant); `.units` behavior identical to the established `_plastic_eff_viscosity` pattern.
- Test vacuity: the FD oracle is lambdify + central differences (no shared `sympy.diff` machinery); the asymmetry guard has 16x margin (0.163 vs 0.01); 3/3 pass on the PR build, 3/3 fail at their primary assertions on pre-fix development (uu_G3 rel 3.13e-01).
- Charter sweep clean; np=2: test_1010 6 passed; Nitsche/rotated/constrained 38 passed; VEP regression 8 passed; full level_1+tier_a gate on the PR build: 531 passed, 0 failed.

Issue #501, opened 2026-08-05
Analytic normal for add_rotated_freeslip_bc fails on any mesh after the first (unwrap canonicalises coordinates to the first mesh's frame)


lmoresi:
In `rotated_bc._boundary_velocity_nodes`, an analytic (sympy) normal is unwrapped and then validated with

```python
stray = set().union(*[c.free_symbols for c in comps]) - set(solver.mesh.X)
```

`unwrap(...)` canonicalises coordinate BaseScalars onto the FIRST mesh's coordinate frame (multi-mesh symbol identity), so on any mesh created after the first, the unwrapped `x` is not `mesh.X[0]` by identity even though it prints identically. The stray check then raises

```
ValueError: analytic normal for boundary '...' contains symbols ['N.x', 'N.y'] that are not mesh coordinates
```

for a normal correctly written in that mesh's own `mesh.X`.

Reproducer (second mesh in one session):

```python
import sympy, underworld3 as uw
from underworld3.function.expressions import unwrap
m1 = uw.meshing.UnstructuredSimplexBox(cellSize=0.2)
m2 = uw.meshing.UnstructuredSimplexBox(cellSize=0.15)
x, y = m2.X
u = sympy.sympify(unwrap(x - 0.5, keep_constants=False, return_self=False))
print(u.free_symbols - set(m2.X))   # {N.x}  — same object as m1's x
```

Fix (already applied to the fault-side twin `fault_contact._compiled_normal_override` on `feature/fault-split-node`, where the check was mirrored from rotated_bc): re-tag the unwrapped symbols by NAME onto the solver's own `mesh.X` before the stray check and the lambdify:

```python
by_name = {str(xk): xk for xk in solver.mesh.X}
retag = {s: by_name[str(s)] for c in comps for s in c.free_symbols
         if str(s) in by_name and s is not by_name[str(s)]}
if retag:
    comps = [c.subs(retag) for c in comps]
```

Genuine stray symbols (names not matching any coordinate) are still caught. Note `lambdify` was already coincidentally correct when the check passed, because it binds by printed name.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
[...]

Issue #494, opened 2026-08-03
cut_along_lines with several lines in one call returns incomplete chains, silently


lmoresi:
## Summary

`cut_along_lines(dm, [line_a, line_b], ...)` returns a mesh in which the surfaces
are **not complete chains of element edges** — there are consecutive pairs of
on-surface vertices with no edge between them. It does not refuse; the returned
mesh looks fine and the labelled facet count looks plausible.

Cutting the same two lines **sequentially**, via two `add_conforming_surface`
calls, is exact.

## Measurement

Box fault, two parallel flanks separated by `W = h` (so an edge can reach both),
graded mesh, `snap_frac=0.30`. Counting collinear vertices and the edges between
them by geometry (a complete open chain on `V` collinear vertices has `V-1`
edges):

| path | flank | vertices | edges | gaps |
|---|---|---|---|---|
| both lines, ONE `cut_along_lines` call | A | 306 | 294 | **11** |
| | B | 310 | 295 | **14** |
| SEQUENTIAL `add_conforming_surface` x2 | A | 318 | 317 | **0** |
| | B | 319 | 318 | **0** |

The gap count is not small, and the affected mesh passes the other checks:
conforming (no facet with more than two cells), Euler characteristic 1, nothing
inverted, and no vertex lying in another edge's interior. So the usual validity
tests do not catch it.

Reproducer: `~/+Simulations/mesh_reconnection_study/chain_gap_check.py` (builds
both paths on identical geometry and prints the table above).

## Why it matters beyond this case

`_fault_collect_polylines(surface)` returns **several polylines for a branched
fault**, and `add_conforming_surface` passes them to `cut_along_lines` in a single
call. So a fault NETWORK goes through exactly this path. The parallel network
tests assert junction geometry and partition-independence but not chain
completeness, so they would not see this.

## Suspected cause (not confirmed)

`_crossing_parameters` iterates over the segments of ALL lines together and keeps
one `t` per edge (`t = np.where(hit, tk, t)`), so a later line overwrites an
earlier one's crossing on the same edge. `multiply_crossed` is meant to catch
that and did NOT fire here, so the mechanism needs checking rather than assuming.

## Suggested fix direction

Either process lines one at a time internally (matching what sequential
`add_conforming_surface` does and is known to be exact), or make the multi-line
path detect and refuse the configuration rather than returning a broken chain.

Either way the chain-completeness property is worth asserting in the test suite
[...]

Issue #492, opened 2026-08-03
Creating a MeshVariable destroys the previous mesh.dm; a held handle segfaults (use-after-free)


lmoresi:
## Summary

Adding a `MeshVariable` replaces `mesh.dm` with a new object **and destroys the
previous DM**, while Python may still hold a reference to it. Any use of the old
handle is a use-after-free and segfaults (SIGSEGV, exit 139).

This makes a natural pattern unsafe:

```python
dm = mesh.dm            # capture
v = uw.discretisation.MeshVariable("v", mesh, mesh.dim, degree=2)
p = uw.discretisation.MeshVariable("p", mesh, 1, degree=1)
dm.getDimension()       # SIGSEGV
```

There is no warning and no error — the process dies. It is easy to mistake for a
fault in whatever ran in between (in my case a Stokes solve with a P0 pressure,
which was entirely innocent).

## Minimal reproducer

42 cells, no solver, no adaptation:

```python
import underworld3 as uw

mesh = uw.meshing.UnstructuredSimplexBox(
    minCoords=(0.0, 0.0), maxCoords=(1.0, 1.0), cellSize=0.3,
    regular=False, qdegree=2)
held = mesh.dm
print("held:", held.getHeightStratum(0)[1])          # 42

v = uw.discretisation.MeshVariable("v1", mesh, mesh.dim, degree=2)
print("after v:", mesh.dm is held, held.getDimension())   # True 2  -- fine

p = uw.discretisation.MeshVariable("p1", mesh, 1, degree=1)
print("after p:", mesh.dm is held)                   # False
print(held.getDimension())                           # SIGSEGV
```

Output:

```
held a handle; cells = 42
  after nothing                            mesh.dm is held -> True
    held.getDimension() = 2
  after creating a P2 vector variable      mesh.dm is held -> True
    held.getDimension() = 2
  after creating a P1 scalar variable      mesh.dm is held -> False
    the held handle is now a DIFFERENT object; using it...
<segmentation fault>
```

Note the trigger is the **second** variable — the first reuses the DM, the one
that changes the field layout rebuilds it.

## Why this is a defect rather than a documentation matter

petsc4py objects are reference counted. Replacing `mesh.dm` is reasonable; what
is not is destroying the old DM while a Python reference to it is alive. The old
handle should stay valid (merely stale) until the last reference goes away, so
the failure mode is at worst a wrong answer from an out-of-date object, never a
crash.

Two possible fixes, in preference order:

[...]

PR #502, opened 2026-08-06
Split-node faults: zero-thickness fault contacts in 2-D and 3-D, parallel, with interface constitutive laws

lmoresi:
## What this is

Zero-thickness split-node faults for UW3, in 2-D and 3-D, as a user-facing
capability. A fault becomes a genuine velocity discontinuity in the mesh:
the labelled facet chain (2-D) or triangulated patch (3-D) is split by
duplicating its interior nodes, a strong no-opening constraint ties the
coincident pairs together, and interface constitutive laws set the
tangential condition. There is no thin weak band and no viscosity contrast
for the solver to fight: the conditioning benchmark measures 10 outer
Schur iterations against 147 for the equivalent one-element weak inclusion
at 1e-4 contrast on the same mesh.

## The user surface

```python
child = mesh.add_fault(("Fault", points))        # cut + split, one call
stokes.add_fault_bc(0, boundary="Fault")         # frictionless
stokes.solve()                                    # ordinary solve
```

- `Mesh.add_fault` places tips and kinks onto mesh vertices, cuts a
  conforming chain, splits it, and records the coincident DOF pairing.
  Lists of faults build offset networks in one call.
- `add_fault_bc(conds, boundary)`: `0` = frictionless, `eta_f` = viscous
  interface; `add_coulomb_fault_bc` and `add_rate_state_fault_bc` for the
  friction family. Laws are sympy expressions in canonical symbols
  (slip rate, effective normal stress, state); consistent tangents come
  from `sympy.diff`, never hand-coded. The effective normal stress is
  SIGNED and reaction-fed; each law clamps its own strength (a cohesive
  law keeps strength into mild tension, a cohesionless one does not).
- `normal=` on every law variant supplies a smooth fault normal — an
  analytic sympy expression, or `"trace"` to build it from the fault's own
  polyline. This matters for sampled curved traces: the default
  facet-averaged normal zig-zags at sampling kinks and produces slip
  notches and traction sawteeth that grow under refinement; the smooth
  normal removes them (measured 7-17x) and restores h-convergence.
- Slip, leak and normal traction are read through the DOF pairing
  (`fault_slip`, `fault_pair_jumps`, `fault_normal_traction`) — the sides
  are geometrically coincident, so coordinate queries cannot see them.
- 3-D meshing support: `BoxInternalPatch` embeds a planar interior patch,
  now with `patch_cellSize` / `grading_distance` to grade resolution from
  the fault outward.

## Parallel

- 2-D: a fault may cross a partition seam through a pinned crossing
  vertex, manufactured automatically; a fault along a seam refuses
  loudly and collectively. Swept at np 2-5. A seven-fault network splits
  and solves at np 1-8 with per-fault slip rank-independent to ~1e-5.
- 3-D: `split_fault` redistributes first — the patch's cell star plus one
  growth layer moves to the rank that already owns most of it (shell
  partitioner, SF-propagated marking), everything else keeps the balanced
  partition. Splits at np 2-8 with pair topology identical to serial;
  measured imbalance 1.8x at np 8 (the thin star, not the refined band).
  Several 3-D faults on one parallel mesh are refused (pairing migration
  through the redistribution is future work).
- Checkpointing: rebuilt-chart meshes now install the point SF on the
  coordinate DM as well (`_install_point_sf`) — previously a parallel
  HDF5 save of any rebuilt mesh wrote shared vertices as owned on every
  rank and the file would not reload. The run-parallel / render-serial
  workflow (solve at any np, `write_timestep`, read checkpoints serially
  for analysis) is exercised end to end by the King study scripts.

## Validation

- 2-D crack against the elliptical slip oracle; 3-D penny crack
  converging to the (8/3 pi)(d tau/eta) sqrt(a^2-r^2) oracle from below;
  leak at machine zero (1e-17 or exactly 0) in every benchmark.
- Law family: viscous compliance monotone from free to welded; Coulomb
  stick/slide switching with reaction-fed normal stress (2-D and 3-D);
  rate-state smoke tests with exact ageing updates.
- Tests: `test_0845` (split topology), `test_0846` (contact physics),
  `test_0847` (user API, analytic/trace normals with negative control),
  `test_0848` (3-D), plus parallel `ptest_0845` / `ptest_0848`.
- Documentation: user guide (`docs/advanced/split-node-faults.md`),
  method and benchmark write-up
  (`docs/developer/design/SPLIT_NODE_FAULT_METHOD_2026-08.md`), a
  fault-mechanics teaching page with nine committed figures and three
  animations, and the deployment design doc.

## Known limitations (all refuse loudly)

Closed-loop faults; daylighting (faults reaching the domain boundary);
junctions sharing vertices; multiple 3-D faults on one parallel mesh.
Split meshes carry no geometric-MG tail (coarse levels do not contain
the fault) — solvers take algebraic-multigrid defaults. The bilateral
no-opening constraint holds a fault shut where it would physically open;
the tell is a tensile fault normal traction, documented in the guide.

## Coordination

This merge brings `reconnect.py` / `line_cut.py` (the cut and rebuild
machinery) to development for the first time. The placed-surface branch
also modifies `reconnect.py`; per the agreed protocol, whichever branch
merges second reconciles — this PR merging first means the
placed-surface session rebases onto development and folds
`_install_point_sf` into its rebuild paths.

An adversarial review is posted as a comment below, per our review
practice.

## Since the review: the network toolkit and the placed-sheet integration

Every follow-up listed at review time is now IN this PR, plus the week's
measured programme:

- **`uw.meshing.FaultNetwork`** — the user-facing network toolkit, both
  dimensions through one API: seniority hierarchy (junior severed by
  senior), automatic junction preparation to the offset form
  (`prepare_fault_network` in 2-D; the planar-patch preparer
  `prepare_fault_surfaces` in 3-D — plane-plane intersection segments,
  in-plane ligament trims), graded meshing, contact on every piece, the
  damage-zone glue (`damage_yield`: ONE dial, strength and xi-style rate
  regularisation together, sharp Piecewise plugs — discs in 2-D, tubes
  about the junction segment in 3-D), per-piece slip diagnostics.
- **Rulings encoded as defaults, each measured** (study:
  `~/+Simulations/fault_junction_rheology`): junction gaps of 1-2
  elements transmit slip within ~7% of a continuous fault at flat solver
  cost and are resolution-converged at 2 elements; zone stress is
  proportional to the dial down to ~100x contrast (a compact plug never
  triggers thin-inclusion conditioning); the glue is ISOTROPIC
  deliberately — an oriented (TI) fabric cannot relax the
  corner-turning deformation a junction must accommodate; the stateless
  viscous-plug baseline matches all damage variants on instantaneous
  slip partition, so damage earns its keep on MEMORY only.
- **`normal="surface"` / `FaultSurface` through the 3-D path**, and
  `BoxInternalPatch` embeds LISTS of disjoint patches.
- **The native placed-sheet route** (`build(mesher="place")`, opt-in):
  non-cumulative from a static base, with the rim triangulator
  (interior grid, centreline points for ligament strips, edge flips —
  no all-rim triangles, the split's precondition). The composed
  parallel chain place -> split_fault -> contact at np=2 converges at
  serial speed with the serial answer (ptest_0852; the SOLVE is the
  gate, because two operator-health defects this week passed every
  topological check). `"embed"` remains the network default pending the
  open place-on-graded-mesh operator-health item.
- **Fixes landed en route**: the rotated-solve auxiliary-vector attach
  (any state field in a rheology segfaulted through the direct entry;
  regression-tested), and the two placed-sheet defects found by this
  branch's acceptance and fixed on that stream (#518's cell handedness,
  #520/#521 interface cones).

Known limitations, stated in the docs: planar convex patches in 3-D
(non-planar preparer designed, not built); parallel MULTI-fault
splitting (single faults are parallel-validated end to end); the
place-on-adapted-base solve-cost item above.


lmoresi:
## Adversarial review

We reviewed this branch as adversaries before asking anyone else to. Six
findings, ordered by how much they should worry a user.

**1. Nodal `read_timestep` on a split mesh is silently wrong, and nothing
refuses it.** Coordinate-based reload cannot distinguish coincident pair
nodes — that is the same reason `fault_slip` reads through the DOF
pairing. Our checkpoints restrict themselves to P0 cell fields by
convention (documented in the King study and the writer's docstring), but
a user who `read_timestep`s a P2 velocity on a split mesh gets
side-mixed values with no error. The API should refuse or warn when the
mesh carries `_fault_point_pairs` and the variable has nodal DOFs.
Follow-up issue material; not fixed in this PR.

**2. The held-shut regime has no runtime tell.** A zero-strength fault
under tensile normal stress would physically open; the bilateral
constraint glues it and the solution is unphysical. The tell (tensile
`fault_normal_traction`) is documented in the guide and the law
docstrings, but nothing warns at solve time. A cheap post-solve check on
law-carrying faults would close this.

**3. Welded-probe slip shows a small rank dependence.** In the
seven-fault network sweep, slipping faults agree across np = 1/2/4/8 to
~1e-5, but one welded probe's peak slip reads 0.00080 at np = 4 vs
0.00083 elsewhere — 4% at the probes' noise floor. Harmless at teaching
scale; worth understanding before quantitative probe use on large
parallel runs.

**4. The redistribution target keeps its full far-field share.** The
star-owning rank measured 1.8x mean load at np = 8 (36k cells). The
imbalance is bounded by the star, not the refined band, but it grows
with patch area fraction; a rebalance pass that sheds far cells from the
target rank is the obvious v2.

**5. The unwrap coordinate-retag fix exists only on the fault path.**
`add_rotated_freeslip_bc`'s analytic normal still fails on any mesh
after the first in a session (#501), while `add_fault_bc(normal=...)`
now works — twin code paths behave differently for the same input until
#501 is fixed on development.

**6. Diff size.** 90 files, +16.6k lines, of which roughly a third is
committed figure assets (PNGs/GIFs/NPZ caches for the docs pages). The
reviewable source surface is: `utilities/fault_split.py`,
`utilities/fault_contact.py`, `utilities/reconnect.py`,
`utilities/line_cut.py`, `utilities/rotated_bc.py` (pair blocks),
`cython/petsc_generic_snes_solvers.pyx` (`add_fault_bc` + solve
dispatch), `meshing/cartesian.py` (`BoxInternalPatch`), and
`discretisation_mesh.py` (`add_fault`). We suggest reviewing those eight
files and treating the figures as docs payload.

**7. The Surface idiom is plumbed but not practised, and crossings are a
release blocker for generic imports.** `add_fault` accepts
`uw.meshing.Surface` (tested), but every shipped example uses the
`(name, points)` shorthand, and in 3-D `FaultSurface` is not consumed by
the split path at all. Any shared vertex between faults refuses; real
imported trace sets cross and abut. The PR body records the agreed
follow-ups (Surface-first idiom, FaultSurface wiring, normals from the
surface objects' own geometry, importer with offset-junction
auto-conversion, ligament-sensitivity measurement before a true junction
design). We flag it here so the merge is not mistaken for
import-anything readiness.

None of these are correctness defects in what the PR claims; 1 and 2 are
the sharp edges a user can cut themselves on today.

lmoresi:
Two of the pre-release follow-ups have now landed on the branch (and in this PR): the FaultSurface object is the 3-D input idiom end to end (rim embed, stored on the mesh, `normal="surface"` from its own face normals), and `uw.meshing.prepare_fault_network` auto-converts crossing/abutting/near-miss traces to the offset-junction form with angle-corrected pull-backs.

The ligament approximation is now measured, not asserted: at a 35-degree T-junction with both faults slipping, the branch's peak slip is 1.01x / 0.96x / 0.79x its isolated value for ligaments of 1 / 2 / 4 local cell sizes, and at fixed ligament 0.04 the entire branch slip profile is identical at h = 0.02 and h = 0.01. The offset junction is a converged physical answer for a given plug size; keep ligaments at 1-2 local cells. A true shared-vertex junction design is therefore an accuracy refinement, staged, not a correctness gap.

lmoresi:
# Adversarial review: PR #502 — split-node faults (fault delta over #488)

Scope: the 53-commit fault delta `pr488-head..pr502-head` (65623c2f..547879c3),
reviewed and run on a clean worktree at the PR head (r502-review,
`pr502-review-head` = 547879c3), built into its own amr-dev env. The
adapt/reconnect layer below (#488) is out of scope; line_cut.py has zero delta
vs the #488 head, and the non-fault files in the raw diff (constitutive_models,
free_surface, multigrid_options, ...) are development merges, not fault work.

## Verdict

The core mechanism is sound and well-tested: the split is a clean rebuild with
one collective refusal point, the pair blocks in Q are exactly consistent with
the interface assembler (mean/√2 on Plus rows, jump/√2 on Minus rows, slip =
Minus rows 1..dim−1; residual √2·M·τ, tangent 2·M·dτ/dV — the algebra closes),
the laws are single-source symbolic with sympy-derived tangents, and serial +
3-D parallel measurements all pass. **One measured merge-blocker**: the
advertised 2-D parallel seam-crossing capability fails its own shipped ptest at
np=2 and skips at np=4 — on this machine the crossing success path runs at NO
rank count. One confirmed moderate bug in the multi-fault diagnostics.

## Measurements

- Serial: `pytest tests/test_0845..0848 -q` → **27 passed, 111.2 s**
  (a concurrent r488 pytest session was running; expect contention inflation).
- `ptest_0845` np=2: **1 FAILED, 3 passed** (the failure is
  `test_a_clean_crossing_splits_and_slips`, identical on both ranks, clean
  exit — no hang). np=4: **3 passed, 1 skipped** (crossing refused:
  "fault vertex sits on the partition seam with both its facets on this
  rank").
- `ptest_0848` (3-D) np=2 and np=4: **PASS, 9/9 checks each, ~7 s** — the
  fault-aware redistribution path is healthy; no hang, well inside the 900 s
  guard.
- Scientific spot-check (serial, `add_fault` + `add_fault_bc(0)` + plain
  `solve()`, h=1/24, far-field shear): converged, outer Newton its = 1,
  outer KSP its = [1], velocity sub-KSP last-apply 60, pressure 15 — no
  repeating-cap signature; **leak 6.9e-18** (machine zero); peak slip 0.1575
  vs crack value Δτ·a/η = 0.2062 (inside the finite-box band the tests
  assert); slip one-signed, fore-aft symmetry defect 0.209 of peak (tip
  nodes dominate; the suite's ellipse RMS < 0.15 criterion passes).
- Negative controls: η_f = 5e4 welds the fault — peak slip 2.05e-5, a
  **7700× reduction**, leak still 1.4e-18. The Coulomb stick control
  (strength > driving shear → creep ~V0) and the reaction-fed σ_n sign/value
  recovery are asserted in test_0846 and pass.
- Law probe: `CoulombFaultLaw(0.6, "reaction", 1e-5)` on a mixed
  compressive/tensile node vector — the `Max(σ,0)` clamp lambdifies
  **elementwise** (tensile node → τ = 0 without contaminating its
  neighbours; no `amax` reduction trap), signs odd in V, tangent is
  `sympy.diff` of the same expression object the residual lambdifies.

## MERGE-BLOCKERS

1. **2-D parallel seam crossing does not work as shipped** (fault_split.py,
   `_pull_seam_vertices_onto_crossings` + the crossing rules in
   `_fault_chain`). At np=2 the shipped ptest geometry
   ((0.15,0.35)→(0.85,0.65), h=1/16 box) dies with
   `ValueError: the labelled facets form a closed loop or several fragments`
   — not even the designed refusal. Standalone diagnosis: after the cut,
   rank 0 holds 7 fault edges in 2 fragments, rank 1 holds 13 in 3
   fragments (the whole fault is ~14 facets — pieces are duplicated as
   shared seam edges), with fragment tips at three distinct shared vertices
   (0.277, 0.405), (0.366, 0.442), (0.530, 0.513). Mechanism: the cut snaps
   the nearest vertices onto the fault line — including partition-seam
   vertices — so wherever the ragged seam approaches the line, the seam is
   dragged ONTO the fault; the crossing design's premise (one crossing
   vertex, both flanking facets rank-local) then cannot hold, and the
   single-crossing machinery has no answer. At np=4 the same geometry is
   refused ("touches the seam without crossing it") and the test skips. Net:
   the capability the guide sells as "handled automatically by add_fault"
   (docs/advanced/split-node-faults.md, Parallel bullet) and the
   03055633/c8693579 commit messages claim at np=1..8 is, as measured on the
   PR head, exercised at no rank count. Options: (a) give 2-D the same
   fault-aware redistribution 3-D already has (`_redistribute_fault_interior`
   gathers a thin star, cost is bounded — this subsumes the whole crossing
   machinery and its special cases), or (b) keep the crossing machinery but
   make the multi-contact case a proper COLLECTIVE refusal with an honest
   message (the fragments ValueError misdiagnoses and escapes the ptest's
   tolerated-refusal list), and soften the guide's claim. We recommend (a):
   the crossing code (crossings walk, `_sf_sum` census, dup_new SF keying,
   third-rank corner rule, seam-vertex pulling) is the most intricate code
   in the delta defending a capability that measurement says is fragile.

2. **Multi-fault diagnostics read the wrong faults** (fault_contact.py,
   `_InterfaceAssembler.__init__` walks every registered LAW fault plus the
   `include` fault). Measured: two-fault network with viscous laws on both,
   `fault_normal_traction(solver, "A", info)` returns **26 nodes (A ∪ B)
   where fault A has 13** — fault B's nodes are interleaved into A's
   traction profile as bogus arc-length samples. `update_fault_state` has
   the same flaw: its slip vector and its θ store update span all
   law-carrying faults, so the returned convergence monitor (median θ|V|/Dc)
   mixes faults, and foreign point keys are written into this fault's
   `by_point` store (dormant, but a landmine for point-id collisions).
   Single-fault runs — every current test — are unaffected, which is why
   the suite is green. Fix is local: the assembler already records
   `fault_of` per row; filter on it in both consumers (or build the
   assembler with ONLY the included fault when include is given).

## Findings (non-blocking, severity order)

3. Rank-asymmetric raises inside the collective rotation build. In
   `build_rotation` the pair-straddles-ranks `RuntimeError` and in
   `_fault_pair_nodes` the pairing/closure-mismatch `RuntimeError` are
   raised rank-locally while peers continue into Q assembly and the Newton
   loop's collectives — a true occurrence deadlocks np>2 instead of
   aborting. Both are "impossible by the seam rules", but blocker 1 shows
   the seam rules are exactly where the surprises live. Same class: an
   analytic normal that vanishes at a node (`_compile_normal_spec` /
   override evaluation) raises only on ranks owning fault nodes. Cheap
   hardening: funnel these through the same allgather-verdict pattern
   fault_split already uses.

4. Crossing-vertex frames are one-sided per rank. At a 2-D seam crossing
   each rank accumulates the pair normal from ITS one local facet, so the
   two ranks assemble interface terms for the SAME global slip row in
   slightly different frames (and the owner's Q block uses its one-sided
   normal) — parallel ≠ serial for any kinked/curved trace at the crossing
   vertex. Straight faults and analytic/trace normal overrides are immune.
   Moot while blocker 1 stands; worth a comment or an SF-completed normal
   when crossings return.

5. Docs contradiction: split-node-faults.md "in 2-D a fault may cross a
   partition seam ... handled automatically by add_fault" — contradicted by
   the np=2/np=4 measurements above. Everything else we spot-read (law
   table, signed-σ_n clamp semantics, would-open detection via the sign of
   `fault_normal_traction`, η_f = η_band/w, tips-stay-welded) matches the
   code and the measurements.

6. Dead state: `_fault_interface_viscosity = {}` initialised in the pyx
   solver setup, never written or read (the laws live in
   `solver._fault_interface_laws`). Drop it or use it.

7. Minor: `solver.add_fault_bc` exposes only frictionless/viscous; Coulomb
   and rate-state stay module functions (documented as such — fine, but the
   asymmetry will surprise); `prepare_fault_network` prints via bare
   `print` (per-rank spam if ever called in parallel); test_0846:239 has a
   dead first `sigma_bg` assignment immediately overwritten.

## Contract checks that PASSED scrutiny

- **Freezing/tangent contract (A)**: interface laws are plain-sympy
  expressions in three module symbols, lambdified once — they never enter
  UWexpression/JIT, and `SymbolicFaultLaw.__init__` REFUSES any stray symbol,
  so a law cannot bake `.sym` or smuggle velocity dependence by
  construction. Residual and tangent lambdify the same expression object;
  the tangent is `sympy.diff(expr, slip_rate)` — no drift possible. The
  bulk operator's Picard/Newton freezing is untouched (the interface tangent
  is a separate Mat added to a COPY of the ptap'd operator each iterate,
  with the null space re-attached). The one lagged quantity (reaction-fed
  σ_n) is lagged deliberately, at iteration starts only, with line-search
  trials evaluated against frozen σ — internally consistent.
- **Rate-state θ**: interpolated as exp(N·ln θ) in BOTH residual and
  tangent (the house rule), θ kept positive under P2 undershoot; the
  ageing update is the exact expm1 integral with a smooth V→0 limit.
- **Signed normal stress**: `update_normal_stress` feeds −σ_nn (positive
  compression) and each law clamps its OWN strength (`Max` inside the
  sympy law; lowers elementwise — measured). Sign convention verified
  end-to-end by test_0846's background-σ_nn recovery.
- **#494**: `add_fault` cuts SEQUENTIALLY — one polyline per
  `cut_along_lines` call — so the multi-line silent-incomplete-chain trap
  is avoided; networks re-split with prior pairings carried through
  `point_map` (verbatim-id trap explicitly handled and asserted).
- **#501**: `_compile_normal_spec` re-tags unwrapped coordinate BaseScalars
  by NAME onto the child mesh before the stray check and lambdify, with a
  comment citing the issue; the negative-control test (foreign symbols
  refused, averaged normal as control) passes. `normal="trace"` (2-D) and
  `"surface"`/FaultSurface (3-D) routes are guarded with clear errors.
- **#492**: no dm handle is held across a MeshVariable creation anywhere in
  the delta — `solver.dm`/`mesh.dm` are re-read per call; the split
  pipeline finishes all DMPlex surgery before constructing the child Mesh.
- **getLabel/empty-IS traps**: every stratum access in the delta is behind
  `hasLabel(name) and getLabel(name).getStratumSize(v) > 0` (2-D split,
  3-D split, redistribution, pair-node walk, assembler) — the null-wrapper
  and null-IS aborts are systematically avoided.
- **Collectivity of refusals**: both splitters gather every verdict at ONE
  allgather before raising, seam verdicts outrank fragment symptoms, and
  ptest_0845/0848 assert same-error-on-every-rank; measured np=2 the
  refusal contract held (identical error both ranks, clean exit, no hang).
- **pyx delta (C)**: +73 lines are dispatch/guards/docstrings only —
  `add_fault_bc`, the `_fault_contact_faults` gate into the rotated path,
  guard()/estimate_difficulty() refusals. No kernel/tensor code, so the
  #493 explicit-loop rule is not in play. `add_fault_bc` composes with
  `add_rotated_freeslip_bc` through the same `build_rotation` (wall blocks
  and pair blocks in one Q; sharing a node is refused explicitly), and the
  fault path reuses the rotated loop verbatim via `solve_with_fault`.
- **Schur-health claim (D)**: we did not reproduce the 10-vs-147 benchmark;
  the frictionless solve here is demonstrably cheap and healthy (outer
  KSP 1, vel sub-KSP 60, pres 15, no capped-count repetition), consistent
  with the claim's direction.

## Rebase notes (08f86036 supersession)

- The stack carries 08f86036: per-field essential-value insertion in
  `_finalize_rotated_solution` (rotated_bc.py ~452-457) via the cython shim
  `petsc_dm_insert_boundary_values` (petsc_discretisation.pyx:314), plus
  test_1018 additions. Development's #500 (d39dbb72) replaced this with
  delegation to `_scatter_global_to_fields` (full DM) because the sub-DM
  insertion segfaults when a datum references a MeshVariable. Known task,
  not a finding.
- Interaction surface: **none of the five fault commits touching
  rotated_bc.py (6b7caac9, ca222a29, a532dbbe, ba78da45, 5a85682b) modify
  `_finalize_rotated_solution` itself.** Their changes live in
  `build_rotation` (pair blocks: 6b7caac9 2-D, 5a85682b 3-D), the Newton
  loop (interface residual/tangent hooks, reaction stash ordering,
  `update_normal_stress` Picard lag: ca222a29, a532dbbe, ba78da45), and the
  diagnostics. Every fault solve reaches the finalizer only through the
  single shared call at the loop exit (rotated_bc.py:997). The rebase can
  therefore take development's finalizer wholesale; drop the shim and the
  08f86036 test additions in favour of development's. Fault-specific
  re-verification afterwards: the split mesh's coincident pair DOFs are
  ordinary unknowns (the no-opening constraint lives in zeroRowsColumns,
  not in DS essential values), so the full-DM scatter should be
  layout-neutral — re-run tests 0845-0848 + ptest_0848 np=2 to confirm.

lmoresi:
Response commits 1d487319..c0d64cda — both blockers fixed, plus findings 3 and 6:

1. **Blocker 1, full route (a)**: 2-D now redistributes first, exactly like 3-D — `_redistribute_fault_interior` is dimension-generic and `split_fault` invokes it whenever the chain touches a partition seam (collective allreduce verdict); `add_fault` redistributes ONCE for a whole network, which also keeps prior pairings valid. The seam-crossing machinery is retired (crossings fan-walk, `_sf_sum` census, third-rank corner rule, seam-vertex pulling, the keyed-SF `dup_new` extension); the low-level splitter now refuses ANY shared chain vertex, checked first, so a straddling fault gets the honest seam verdict rather than the fragment misdiagnosis the review measured. The crossing ptest asserts SUCCESS: np=2 and np=4 all pass (previously died / skipped), np=3 passes the user-facing test. Rank-independence on the shipped crossing geometry: peak slip agrees to ~1e-8 across np=1/2/4 (leak 1.4–2.8e-17). 3-D unaffected: ptest_0848 9/9 at np=2 and np=4 with the shared helper. Docs Parallel bullet rewritten for redistribute-first in both dimensions.
2. **Blocker 2**: `_InterfaceAssembler` restricts to the `include` faults when given (the solve path, which needs all faults, is untouched). Regression test `test_multi_fault_diagnostics_are_per_fault` — negative control confirmed (44 rows where fault A has 23 with the fix reverted; exact-count pass with it).
3. **Finding 3**: rank-local raises in the fault rotation build now funnel through an allgather-verdict `_collective_raise` (straddle, shares-a-node, closure-mismatch, vanishing-normal) — deadlock class retired. The straddle check is kept: checks that "cannot fire" are exactly where blocker 1 lived.
4. **Finding 6**: dead `_fault_interface_viscosity` dropped (verified sole reference repo-wide).

Serial suite 28/28 (the new regression test included). Remaining known work before merge: this branch is stacked on #488 (which owns the two failing CI tests) and carries the superseded 08f86036 form of the #500 fix — both resolved at the rebase once #488 lands.

lmoresi:
CI diagnosis for the current head (run of 2026-08-10, `4 failed, 573 passed`): all four failures are in the new `test_0851_*` files, and three are the environment-geometry lottery this repo keeps meeting (cf. the #449 tripwire and the 0753 story on #488):

1. `test_0851_fault_network_3d` ×3 — CI's Linux gmsh builds a different base geometry than the dev box, so 'Main' and 'Cross' land 0.02233 apart, under the ligament thresholds (0.045 / 0.08), and the v1 preparer's genuine-X-crossings-only guard raises `NotImplementedError`; the `test_embed_refuses_crossing_patches` variant instead reaches tetgen and dies with a raw `PLC Error: a segment and a facet intersect`. The tests encode dev-box geometry as constants. Suggested shape: derive the patch separation from the *constructed* patches (measure, then assert the preparer's verdict matches the measured class), or pin the geometry analytically so no gmsh discretisation enters the ligament computation. The raw PLC escape is also worth catching at the preparer level — a refusal should be the library's message, not tetgen's.
2. `test_0851_std_reduction_method::test_mesh_simple_array_view_std` — `petsc4py.PETSc.Error: error code 79` (PETSC_ERR_MISSING_FACTOR class) on CI only; likely an environment-sensitive solver default rather than the reduction itself. Needs its own look — if the test doesn't need a solve, decouple it from one.

Per the shared-file coordination, leaving these to this branch's owning session — flagging so the next CI round is aimed. The `petsc_dm_insert_boundary_values` shim in `petsc_discretisation.pyx` also survived the merge with no callers (development's #500 replaced it); it should be dropped rather than shipped dead.

lmoresi:
Correction to our 2026-08-10 CI diagnosis above: the `test_0851_*` failures were NOT the geometry/gmsh lottery we suggested. PR #533 has the true root cause: `_line_clip_interval`'s parallel-edge branch had an INVERTED inside test, reachable only when the projection denominator is exactly 0.0 — CI's OpenBLAS produces exact zeros for axis-aligned plane bases, macOS Accelerate leaves ~1e-17 jitter and dodged into the sign-correct division path. The dev boxes passed by luck; CI was right. The error-79 failure was a cascade (the escaped PLC exception left gmsh initialized; the next mesh build in the same batch wrote an invalid .msh). Both fixed in #533 with a platform-deterministic regression test. The residual follow-up for this branch's session: raw PLC errors can still escape for coplanar/touching patch sets — the preparer should own that refusal (TODO(BUG) placed at the site).

Issue #162, opened 2026-04-30
[BUG]  The natural boundary condition is not working on the internal boundary


NengLu:
The natural boundary condition is not working on the internal boundary.

Here is a simple test: uniform viscosity, 0 body force, traction boundary condition on internal boundary, no slip on other boundaries.

### uw3
```python
import underworld3 as uw
import numpy as np
import sympy
xmin, xmax = 0, 1
ymin, yint,ymax = 0,0.5,1
xres, yres = 32,32
yresa,yresb = 16,16
mesh = mesh = uw.meshing.BoxInternalBoundary(elementRes=(xres,yres),zelementRes=(yresa,yresb),minCoords=(xmin,ymin),maxCoords=(xmax, ymax),zintCoord=yint,degree=1,qdegree=2)
v = uw.discretisation.MeshVariable("V", mesh, mesh.dim, degree=2,continuous=True)
p = uw.discretisation.MeshVariable("P", mesh, 1, degree=1,continuous=True)
stokes = uw.systems.Stokes(mesh, velocityField=v, pressureField=p)
stokes.constitutive_model = uw.constitutive_models.ViscousFlowModel
stokes.bodyforce = sympy.Matrix([0., 0.])
stokes.constitutive_model.Parameters.shear_viscosity_0 = 1.0
stokes.saddle_preconditioner = 1.0 / stokes.constitutive_model.Parameters.shear_viscosity_0
for wall in ("Left", "Right", "Bottom", "Top"):
    stokes.add_essential_bc((0.0, 0.0), wall)
stokes.add_natural_bc(sympy.Matrix([1.0, 0.0]), "Internal")
stokes.tolerance = 1.0e-6
stokes.petsc_options["ksp_rtol"] = 1.0e-6
stokes.petsc_options["ksp_atol"] = 1.0e-6
stokes.petsc_options["snes_converged_reason"] = None
stokes.petsc_options["snes_monitor_short"] = None
stokes.solve(zero_init_guess=False)
print("u just below y=0.5:", uw.function.evaluate(v,np.array([0.5,0.499])))
print("u just above y=0.5:", uw.function.evaluate(v,np.array([0.5,0.501])))

```
The solver did not converge.

### same case using FEniCS
```python
from dolfin import *
import numpy as np
mesh = UnitSquareMesh(32, 32)
OUTER     = 1
INTERFACE = 2
boundaries = MeshFunction("size_t", mesh, mesh.topology().dim() - 1, 0)
[...]

PR #503, opened 2026-08-06
Remove the Null_Boundary fake natural BC and the label it was built on

lmoresi:
## What this removes

Two things, in two commits, plus a test.

- `866ae55d` — the fake natural BC. Whenever a solver had any natural BC, it
  manufactured one more on `Null_Boundary`, commented as "a workaround for some
  problem in the PETSc machinery where we need a surface integral term somewhere
  on every process if we have a contribution from anywhere".
- `feb86401` — the label itself. `Null_Boundary`, value 666, was added to every
  UW3 mesh at construction.
- `2aec7923` — `tests/parallel/test_0768_unowned_boundary_natural_bc_mpi.py`,
  which pins the case the workaround claimed to need.

## The measured basis

**Value 666 marks no facet.** Measured after a solve on a unit box: 198 points,
all vertices, zero edges. A boundary residual attached to a value that marks no
facet assembles over nothing, so the fake term was identically zero. The source
comment said as much ("integrates nothing over a bunch of points") without
following the thought through.

**The DS was already rank-identical.** Natural BCs are registered against the
consolidated `UW_Boundaries` label unconditionally on every rank —
`PetscDSAddBoundary_UW` runs in a loop with no rank-local skip. So the hack could
not have been fixing a rank-*consistency* problem. The one thing it could still
have supplied was a locally non-empty stratum on every rank, and that is what the
probes were built to test.

**Answers do not move.** A long flat box (4.0 x 0.25) makes the partitioner cut
along x, so the short end faces land on one rank each and everyone else owns
none of them. With a natural BC on such a face, results are bit-identical np=1
through np=8 with up to 7 of 8 ranks owning no facet of it: scalar Poisson
against the P2-exact `T = x²` at `tolerance = 1e-10`, a `BdIntegral` sweep
(byte-for-byte identical, `diff` clean), and Stokes traction and Nitsche free-slip
for the vector path the fake condition `(0,)*u.shape[1]` was actually shaped for.

**Old checkpoints still read.** A file written before the removal loads at np=1
and np=3, area exactly `1.0000000000`, keeping its own `Null_Boundary` label and
enum member from the file. Old data keeps its labels; new meshes simply stop
manufacturing one. That is also why the sentinel skips stay — a reloaded mesh
still carries `Null_Boundary` in its boundaries enum, as do several caller enums
in `docs/examples` and `test_1012`.

## What the test coverage actually is

Please do not read the parallel-suite pass count as evidence for this change. Of
the parallel tests run, exactly **one existing file** creates a natural BC at all
(`test_0770_submesh_extract_mpi`), and in it every rank owns part of both natural
boundaries — `Internal` [26,26] at np=2 and [12,14,13,13] at np=4. Ranks owning
none: zero. `test_1064_rotated_freeslip_parallel` uses a strong rotated
constraint, which never populates `natural_bcs`, so it never entered the removed
branch. Those tests are regression cover that nothing else broke. They say
nothing about this.

The new file is the coverage. Owned facets of `Right`, ranks owning none: 1 of 2,
2 of 3, 3 of 4, 5 of 6, **7 of 8**. It carries a vacuity guard, because without
one a partitioner change would make every other test in it pass for no reason.
Both its negative controls were run and fire: dropping the scalar flux term moves
the relative L2 error from 3.6e-11 to 8.010e-03, and dropping the vector traction
makes the two Stokes solves agree to every digit.

It is a tripwire on current behaviour, not a proof that the PETSc problem the
workaround was aimed at is gone. Removing a failed fix does not establish that
the thing it failed to fix has gone away.

## Hazard left in place, deliberately

Beside the DS registration loop sits

```
bc_is = bc_label.getStratumIS(value)   # petsc_generic_snes_solvers.pyx:3284
```

computed and never read. That dead line invites `if bc_is is None: continue` as
an apparent optimisation, which would reintroduce precisely the failure the
workaround claimed to guard against — and, before this PR, with every test still
green. It is also a latent issue-#291 segfault that never fires only because it
is never dereferenced. Out of scope here; flagged so it is not lost.

## Where this came from

Found while landing the mesh DELETE primitive. `Null_Boundary` marks every vertex
of every mesh and `UW_Boundaries` re-packs it, so reading labelled *points* as
material interfaces refused 1114 of 1114 mesh-repair candidates, silently. That
fix (read edges, not points) is independent and already landed; this is the
upstream cleanup.

## Verification

- serial: 93 passed, including `test_1012_stokesImportedDMPlex` and the smoothing
  tests that skip the sentinel;
- parallel np=2 and np=4: 23 passed each, including
  `test_0766_box_internal_boundary_mpi` — the issue-#162 case, where losing these
  labels once silenced `BoxInternalBoundary`'s natural BCs;
- Nitsche paths, which also triggered the hack and were missing from the first
  sweep: `test_1065_nitsche_local_h`, `test_0641_wave_c_api_shims`, 48 passed;
- new file: 4 passed at np=2 and np=4, skips cleanly in serial.

Full logs and probe scripts under `~/+Simulations/null_boundary_bc_hack/`.

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -1961,17 +1961,6 @@ class SolverBaseClass(uw_object):
             if hasattr(self, "_constant_nullspace_obj"):
                 self._constant_nullspace_obj = None
 
-        # This is a workaround for some problem in the PETSc machinery
-        # where we need a surface integral term somewhere on every process
-        # if we have a contribution from anywhere. We add a fake one here
-        # which just integrates nothing over a bunch of points. It's enough
-        # to let the rest of the machinery work.
-
-        if len(self.natural_bcs) > 0:
-            if not any(bc.boundary == "Null_Boundary" for bc in self.natural_bcs):
-                bc = (0,)*self.Unknowns.u.shape[1]
-                self.add_natural_bc(bc, "Null_Boundary")
-
         if verbose:
             uw.pprint("Build pointwise functions")
         self._setup_pointwise_functions(verbose, debug=debug, debug_name=debug_name)
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -884,13 +884,24 @@ class Mesh(Stateful, uw_object):
     ):
         """Extend the boundary enum and rebuild the DM's boundary labels.
 
-        Patches the user-supplied boundary enum with the members every UW
-        mesh needs (``Null_Boundary`` — every vertex, value 666 — and
-        ``All_Boundaries``, value 1001), records the boundary / region
-        metadata attributes on the mesh, rebuilds named boundary labels for
-        wrapped DMPlex imports that only expose stacked Gmsh label sets,
-        and builds the ``Null_Boundary`` and stacked ``UW_Boundaries``
-        labels used by the boundary-condition machinery.
+        Patches the user-supplied boundary enum with ``All_Boundaries``
+        (value 1001, every exterior facet, populated by ``markBoundaryFaces``),
+        records the boundary / region metadata attributes on the mesh, rebuilds
+        named boundary labels for wrapped DMPlex imports that only expose
+        stacked Gmsh label sets, and builds the stacked ``UW_Boundaries`` label
+        used by the boundary-condition machinery.
+
+        ``Null_Boundary`` (value 666, every VERTEX of the mesh) used to be
+        added here as well. Nothing needs it: it marks no facet, so it
+        integrates nothing, and its one functional consumer — a fake natural BC
+        the solver manufactured to guarantee "a surface integral term on every
+        process" — has been removed and measured to change no answer at any
+        rank count. What it did do was mislead: any pass that reads *labelled
+        points* to find material interfaces saw every vertex of the mesh
+        labelled, which silently refused 1114 of 1114 mesh-repair candidates.
+        A boundaries enum supplied by a caller may still declare it, and old
+        checkpoints still carry the value, so the sentinel skips downstream
+        remain.
         """
         ## Patch up the boundaries to include the additional
         ## definitions that we do / might need. Note: the
@@ -900,7 +911,6 @@ class Mesh(Stateful, uw_object):
         if boundaries is None:
 
             class replacement_boundaries(Enum):
-                Null_Boundary = 666
                 All_Boundaries = 1001
 
             boundaries = replacement_boundaries
@@ -908,7 +918,6 @@ class Mesh(Stateful, uw_object):
 
             @extend_enum([boundaries])
             class replacement_boundaries(Enum):
-                Null_Boundary = 666
                 All_Boundaries = 1001
 
             boundaries = replacement_boundaries
@@ -947,20 +956,6 @@ class Mesh(Stateful, uw_object):
                     if self.dm.getLabel(stacked_label_name):
                         uw.adaptivity._dm_unstack_bcs(self.dm, self.boundaries, stacked_label_name)
                         break
-
-        # Null_Boundary marks EVERY vertex with the reserved value 666 —
-        # the catch-all stratum used when a condition applies mesh-wide.
-        # Note: getStratumIS(0) on the "depth" label fetches the depth-0
-        # points, i.e. VERTICES (the old name "all_edges" was wrong).
-        depth_label = self.dm.getLabel("depth")
-        self.dm.createLabel("Null_Boundary")
-        null_boundary_label = self.dm.getLabel("Null_Boundary")
-        if depth_label and null_boundary_label:
-            vertex_stratum_is = depth_label.getStratumIS(0)
-            if vertex_stratum_is:
-                null_boundary_label.setStratumIS(
-                    boundaries.Null_Boundary.value, vertex_stratum_is
-                )
 
         ## --- UW_Boundaries label
         if self.boundaries is not None:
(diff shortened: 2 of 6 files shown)


lmoresi:
# Adversarial review — `bugfix/null-boundary-hack` (`866ae55d`, `feb86401`)

Reviewed: removal of the fake `Null_Boundary` natural BC, and of the label.

## Findings

**1. The parallel-suite evidence was overstated.** We reported "23 parallel
tests pass" as evidence for the removal. Of the parallel tests run, only
`test_0770_submesh_extract_mpi` creates a natural BC
(`stokes.add_natural_bc(...)`, lines 178-179).
`test_1064_rotated_freeslip_parallel` uses `add_rotated_freeslip_bc`, a strong
constraint that never populates `natural_bcs`, so it never entered the removed
branch (`if len(self.natural_bcs) > 0`). Those tests are regression cover that
nothing else broke — not evidence about this change. Real parallel natural-BC
coverage in the suite: **one file**.

**2. Nitsche was an unexamined surface.** `add_nitsche_bc` appends directly to
`self.natural_bcs` (`petsc_generic_snes_solvers.pyx:4221` vector, `:6314`
Stokes saddle-point), so every Nitsche solve triggered the hack. Two
hack-triggering test files were in neither the baseline nor the after run:
`test_1065_nitsche_local_h.py`, `test_0641_wave_c_api_shims.py`. Now run: 48
passed.

**3. The vector case had no probe.** The fake condition was built as
`(0,)*self.Unknowns.u.shape[1]` — shaped for a vector unknown — and every probe
was scalar Poisson. Added `stokes_nitsche_unowned_box.py`. Stokes traction on a
face 7 of 8 ranks own no facet of: vrms `2.582785783291e-01` (np=1) →
`...783278e-01` (np=8), 12 figures. Nitsche free-slip at the bundle default
tolerance: converged, vrms `2.582981693232e-01` (np=1) vs `2.582998224075e-01`
(np=4).

**4. First version of that probe was true by construction.** Uniform gravity in
a closed incompressible box is hydrostatic — vrms came out `1.25e-13`, and
np-independence of zero is not evidence for anything. Re-posed as lid-driven,
with a `-uw_mode none` liveness control: `none` vrms `2.582851221413e-01` vs
`natural` `2.582785783291e-01`, so the boundary term is live.

**5. `DIVERGED_LINE_SEARCH` in the Nitsche arm is ours, not the code's.**
At `solver.tolerance = 1e-10` Nitsche fails the line search (reason −6) at every
rank count. Attributed by rebuilding at `22aae82c` (pre-removal): identical
reason −6, and np=1 vrms bit-identical at `2.583006748159e-01`. It is a
tolerance artefact — at the bundle default it converges (reason 3) at np=1 and
np=4. Not a defect, not attributable to this change.

**6. Our own absence assertion was wrong.** `test_no_null_boundary_label_is_manufactured`
first asserted `dm.getLabel("Null_Boundary") is None` and failed against a DM
that genuinely does not have the label: `getLabel` returns a non-`None` DMLabel
wrapper with handle 0, so `is None` reports every absent label as present. Now
asserted against `[dm.getLabelName(i) for i in range(dm.getNumLabels())]`. Same
family as #291.

**7. The justification is not pinned by any test.** Two different claims:

* *the removal happened* — pinned. Re-adding the fake BC fails `test_1001`
  (`natural_bcs == ["Top"]`); re-adding the label fails `test_0001`.
* *the removal was safe* — i.e. natural BCs assemble correctly when a rank owns
  no facet of the natural boundary — pinned by **nothing in `tests/`**.

The suite has one parallel test that creates a natural BC, `test_0770`, and in
it every rank owns part of both natural boundaries: `Internal` [26,26] at np=2
and [12,14,13,13] at np=4; `Lower` [13,13] and [6,5,6,9]. Ranks owning none:
**zero**. The discriminating condition is exercised by no test in the repo,
before or after this change. Our evidence for it (np=1..8, up to 7 of 8 ranks
owning nothing, bit-identical) lives in `~/+Simulations`, which CI does not run.

The live hazard: the DS loop registers every natural BC on every rank
unconditionally, and `bc_is = bc_label.getStratumIS(value)` at `:3284` sits
beside it computed and never read. That dead line invites
`if bc_is is None: continue` as an apparent optimisation, which would
reintroduce exactly the failure the workaround claimed to guard against, with
every test still green.

## Attacks that failed

* **Label completion.** Value 666 measured AFTER a solve: 198 points, all
  vertices, 0 facets. PETSc never completes it into anything integrable.
* **Accumulation across rebuilds.** Guard held; no duplicates before or after.
* **Issue #162** (losing these labels once silenced `BoxInternalBoundary`'s
  natural BCs): `test_0766_box_internal_boundary_mpi` passes np=2 and np=4.
* **Caller enums declaring `Null_Boundary = 666`** (`test_1012`, several
  `docs/examples`): pass. `extend_enum` merges the caller's member; the sentinel
  skips were kept for exactly this.
* **Old checkpoints.** A pre-removal file reads at np=1 and np=3, area exactly
  `1.0000000000`, keeping its own label and enum member from the file.
* **Scalar unowned-boundary case.** Bit-identical to every digit, np=1..8.
* **`BdIntegral` sweep.** Byte-for-byte identical (`diff` clean).
* **Dead code.** `bc_is = bc_label.getStratumIS(value)` at `:3284` is computed
  and never read — a latent #291 segfault that never fires because it is not
  dereferenced. Left alone; out of scope.

## Merge conditions

* Land the unowned-boundary case as a parallel test
  (`tests/parallel/test_0768_unowned_boundary_integral_mpi.py`), scalar at
  minimum, so finding 7 is closed.
* State in the PR body that parallel natural-BC coverage rests on one existing
  test file plus the new one — not on the 23-test count.

lmoresi:
-

lmoresi:
The dead `bc_is` line flagged in the PR body is now tracked as #506. It turned out to be four sites, not one — `SNES_Scalar` (twice), `SNES_Vector`, `SNES_MultiComponent` — all assigned and never read. Still out of scope here; the test this PR lands is what would catch a regression if someone "optimised" them into a rank-local skip.

PR #510, opened 2026-08-08
Mesh surgery base: conforming surfaces, reconnection repair, fault zones (2-D)

lmoresi:
The 23-commit mesh-surgery stack that both active fault streams already build on, landed as the shared base so that neither carries it as private ancestry.

## What this is

The branch point (6d494db1) common to `feature/mesh-reconnection` and `feature/fault-split-node`. Landing it subtracts identical content from both feature branches and turns their eventual PRs into reviews of their own work only.

Contents, all 2-D:

- `utilities/line_cut.py` — conforming surfaces cut into an existing mesh (`Mesh.add_conforming_surface`), serial and parallel, partition-independent; snapping with quality veto; `pull_vertex_onto` for tips and junctions.
- `utilities/reconnect.py` — the repair passes: `flip_to_reduce_max_angle` (Lawson flips gated on the maximum angle, Babuska-Aziz), `remove_vertices` (the delete primitive, flip-first ordering measured), chart-preserving and chart-compacting rebuilds with one-exchange star-forest renumbering.
- `utilities/edge_split.py` engine integration and the one-MG-level-per-doubling rule (a cut is not an MG level).
- `Mesh.cells_supporting` — the fault zone as the support of the labelled facets.
- Tests: `test_0843`, `test_0844_line_cut` (37), `test_0844_reconnect_repair`, `test_0845`, parallel suites `ptest_0844_*` (np=2/3 validated).
- Docs: `docs/developer/subsystems/conforming-surfaces-and-fault-zones.md` and design notes.

## What this deliberately excludes

The two placement commits (`b907c83f`, `f4490493`) follow separately with the reconciled `reconnect.py`, because `feature/fault-split-node` modified that file independently after this branch point and the reconciliation belongs in its own reviewable PR.

## Verification

- Dry-run merge against `development`: zero conflicts.
- The stack has been exercised for weeks as the ancestry of both fault streams; the serial and parallel suites above pass on the worktree environments.

Coordination ledger: Planning hub, liaison item (2026-08-09 response) — landing sequence PR-A (this) → PR-B (placement + reconciled reconnect) → PR-C (fault-split-node).


lmoresi (commented):
Adversarial review (authoring session; findings, not ceremony).

**What we tried to break, and could not:** the stack is the common ancestry of both active fault branches and has been exercised for ~3 weeks under them; the serial suites (0843/0844/0845, 82 tests) and the parallel suites (np=2/3, 23 each) pass on the reconciliation branch stacked on this one. The dry-run merge into development is conflict-free.

**Findings that ride along, stated so the merge is informed:**

1. **Issue #494 lands with this.** `cut_along_lines` with multiple lines in ONE call can return incomplete chains without raising. Known, tracked, and the workaround (cut sequentially — each surface its own call) is what every caller in the repo does. Not a regression; it predates this PR's branch point and the fix is scheduled with the placement path (#511's successor work).
2. **`repair=True` gives up partition-independence** (which cells may be touched depends on the seam). Documented prominently in `reconnect.py` and opt-in; the default path is bit-confluent at any rank count.
3. **Semantic staleness risk vs post-#503 development:** this base was written when `Null_Boundary` still existed; development has since removed it. We checked the code paths: `_interface_edges` / `_BOOKKEEPING_LABELS` read labels generically and nothing depends on `Null_Boundary`'s presence — its removal makes the documented vertex-noise trap moot rather than the guard wrong. The docstrings describing that trap are now history rather than description; accurate either way. CI's merged-tree run is the authoritative check here.
4. **An essential BC on an embedded surface under geometric MG is unsound** (coarse levels carry no surface DOFs) — documented with a warning in `add_conforming_surface`; the material-contrast use case this exists for is unaffected.

Recommend merge once CI is green; #511 stacks on it.

PR #511, opened 2026-08-08
Placed surfaces (2-D serial) + the reconciled reconnect.py

lmoresi:
Stacked on #510; retarget to `development` when it lands. Second step of the untangling ledger (planning hub, liaison item, 2026-08-09).

## What this adds

`utilities/place_surface.py` — embed a surface by placing its own points, deleting the mesh vertices in the way, and retriangulating the cavity. Unlike the cut it handles a fault **ending inside the mesh** (the tip is an angle-interpolated fan, no width floor) and a surface **finer than the local h** (`spacing` is a parameter). 2-D, serial; parallel is refused with the reason (placement adds points, and the chart-expansion star-forest rebuild does not exist).

Guards learned the hard way, all tested: cells owning an interior labelled facet are held out of the cavity (protecting a vertex does not protect a facet — both cells of a facet can be cleared while every corner is protected, silently destroying an earlier surface); every placement re-reads each earlier surface's facet count off the **result** mesh; the walk carries a quality floor because collinear wall triangles pass every positivity test at area 1e-15.

Honest capability note, in the docs: for closely spaced surfaces the **cut is currently more capable** (accepts 1.0 h separation vs placement's 1.5 h); placement's limit is implementational (one cavity, one surface) where the cut's is inherent.

## The reconciliation

Both fault streams evolved `utilities/reconnect.py` after the shared branch point — this stream renamed `rebuild_without_vertices` → `rebuild_cavities` (it now adds placed vertices as well as deleting), the contact stream added `_install_point_sf` (a rebuilt star-forest must also be installed on the coordinate DM or parallel HDF5 saves are unloadable serially; their measurement). This PR folds their hunks in verbatim (from `c680e010`/`c8693579`/`1d487319`), and routes `rebuild_with_cones`' verbatim SF handoff through the same helper since it carries the same trap. After this lands, `reconnect.py` changes only via `development` (ledger rule 1), and `feature/fault-split-node`'s merge diff on it is the one-line rename adaptation plus three doc references.

Test renumbered `0848` → `0853` (`0848` is claimed by the contact stream's pushed `test_0848_fault_split_3d.py`; registry in the ledger).

## Verification

- 82 tests: `test_0853_place_surface` (21), `test_0844_line_cut` (37), `test_0844_reconnect_repair`, `test_0843`, `test_0845` — all pass.
- Parallel: `ptest_0844_reconnect_parallel` + `ptest_0844_line_cut_parallel` at np=2 and np=3 (23 each) — pass.
- Style gate clean. Negative control: stubbing `place_along_lines` to a no-op fails the suite.


lmoresi (commented):

**Where we attacked it:**

1. **The silent-corruption class.** The reason this PR exists in this shape: protecting interface *vertices* does not protect interface *facets* — both cells of a labelled facet can be cleared while every corner is protected, and an earlier surface loses a link with nothing raised. The guard holds those cells, requires a victim's whole star to be clearable, and re-reads every earlier surface's facet count **off the result mesh**. The regression test reproduces the original corruption (trunk 21→20 facets) when the guard is reverted.
2. **Vacuous-test check.** Stubbing `place_along_lines` to a no-op fails the suite (negative control run before commit). The close-pair test asserts counts off the result, not summed bookkeeping — the earlier version of that test asserted an identity a partial corruption still satisfies, and the file now documents that mistake so it is not re-made.
3. **The reconciliation.** We diffed the reconciled `reconnect.py` against `feature/fault-split-node`'s copy hunk by hunk: the residual is exactly the `rebuild_cavities` rename + `placed_coords` machinery, i.e. their file + our delta. Their `_install_point_sf` hunks are verbatim (from c680e010/c8693579/1d487319). One deliberate extension: `rebuild_with_cones`' verbatim SF handoff goes through the same helper — it creates its coordinate DM before the SF is installed, so it carries the identical trap; leaving one of three sites on the old pattern is how the next divergence starts.

**Honest limitations, in the docs rather than discovered later:** serial only (parallel refuses with the reason — placement adds points and the chart-expansion SF rebuild does not exist); for closely spaced surfaces the CUT is currently more capable (1.0 h vs 1.5 h separation, measured, both surfaces verified intact); junctions refuse loudly — no junction capability is claimed.

**Efficiency note:** the per-placement passes are O(cells) Python loops and `_interface_facet_counts` scans all labels; fine at study scale (~10⁴ cells, <1 s), and the right time to optimise is when the 3-D generalisation forces the data structures anyway.

Recommend merge after #510; the contact stream's adaptation is the one-line rename + two→three return values, recorded in the planning ledger.

PR #515, opened 2026-08-09
Geometric-MG bundle owns its outer-KSP pairing: fgmres, applied to the live KSP

### LINKED ISSUE #514: Geometric-MG bundle pairs a gmres smoother with a non-flexible outer KSP: converges by recurrence while the true residual stalls
Found by PR #510's CI: `test_0842_nvb_3d_parallel_adapt::test_poisson_fmg_on_3d_child_matches_gamg` fails on the merged tree (base + development) with true relative error 1.0e-6 against the 1e-8 gate, while passing on either side alone.

## Mechanism, measured

Since #471 the geometric-MG bundle's default smoother is `gmres/4` ("robust"). A Krylov smoother makes the PCMG preconditioner **vary between applications**, and the outer KSP on the scalar path is plain `gmres`, whose recurrence assumes a fixed preconditioner. The signature is exact: preconditioned residual drops nine orders and the KSP reports CONVERGED_RTOL at 3 iterations, while the **true** residual stalls at 1.3e-6 from iteration 2.

One-knob probes on the failing tree (3-D adapt child, 5198 cells, Poisson T=z):

| configuration | outer | its | true rel. error |
|---|---|---|---|
| bundle as shipped (gmres smoother, gmres outer) | gmres | 3 | **1.0e-6** |
| `mg_levels_ksp_type=richardson` (linear PC) | gmres | 5 | 4.6e-10 |
| bundle as shipped + `ksp_type=fgmres` | fgmres | 3 | 7.5e-10 |

Either a linear smoother or a flexible outer restores honest convergence; the shipped pairing is the only failing combination.

## Why development's own CI is green

The variation of a gmres smoother per application depends on the hierarchy quality; on current development hierarchies the recurrence drift stays below the 1e-8 test gate. PR #510's one-level-per-doubling hierarchies (subsampled, transfers spanning more refinement) push the drift to 1e-6. The pairing is unsound everywhere; the exposure threshold is what differs — same class as the fieldsplit ruling that made `fgmres` mandatory over Krylov sub-solves.

## Proposed fix

`multigrid_options`: when the bundle's smoother variant resolves to a Krylov type, also set the outer KSP to `fgmres` under the same user-ownership latch #471 introduced (a user-owned `ksp_type` is respected and warned about). The `fast` (richardson) variant needs no outer change. Regression: a unit assertion on the bundle pairing plus `test_0842` as the integration canary — it bites on any hierarchy the drift exceeds the gate on.

lmoresi:
Fixes #514. First car of the landing train recorded in the planning-hub liaison ledger: this → #510 → #511.

## The defect

Since #471 the geometric-MG bundle's default smoother is `gmres/4`. A Krylov smoother makes the preconditioner **vary between applications**, and a plain-gmres outer's recurrence assumes it fixed. Measured on a 3-D adapt child (PR #510's CI, reproduced locally): the KSP reported CONVERGED_RTOL at 3 iterations while the **true** residual stalled at 1.3e-6 — one hundred times the gate. The same failure was previously found and fixed *locally*, twice: the Stokes velocity sub-KSP (the bundle's own comment records the reason) and the free-surface solver ("false-converges … TRUE residual blew up", `systems/solvers.py`). Piecemeal fixes of one pairing rule are the drift #471 exists to end.

## The fix, three parts

1. `ksp_type = fgmres` joins the geometric bundle, both smoother variants — richardson does not need flexibility, but every variant must own the same key set (the stale-key derivation requires it), and fgmres judges convergence on the true residual, which is never wrong.
2. The three solver-default writers of `ksp_type` now go through `_push_managed_option`, so the bundle may upgrade the **default** but never a user's explicit choice (`petsc_options["ksp_type"] = ...` still wins, per the #471 ownership latch).
3. The top-level install path applies the resolved type to the **live KSP**: the database write alone is provably inert there — the KSP consumed its options long before injection — and the new live test caught exactly that inert first version of this fix. The fieldsplit velocity sub-KSP needs no live application (its `setFromOptions` runs at the parent's `PCSetUp`, after the write); the rotated path already writes fgmres unconditionally on its per-solve prefix.

## Verification

- New unit test: the pairing invariant on every smoother variant, plus GAMG's stale list clearing the key.
- New live test: adapt-child scalar solve comes out `pc=mg` + `ksp=fgmres` on the live objects.
- On development + this fix: `test_1021` (20), `test_0842` (4), rotated parallel `test_1064` at np=2 (9), Poisson/Stokes ND smoke (9) — all pass.
- On the composed tree with PR #510's hierarchies — the shape that exposes the drift: `test_0842` passes at its 1e-8 gate (was 1.0e-6 broken; 2.7e-10 fixed), 40 tests green across 0842/1021/0753.
- Honest scope note: on hierarchies where one FMG application converges the cycle, the two pairings are measurably indistinguishable (3.3e-8 either way on the 2-D fixture); `test_0842` is the discriminating gate.

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -3076,7 +3076,7 @@ class SNES_Scalar(SolverBaseClass):
         self._pc_option_prefix = ""
 
         self.petsc_options["snes_type"] = "newtonls"
-        self.petsc_options["ksp_type"] = "gmres"
+        self._push_managed_option("ksp_type", "gmres")
         self._push_managed_option("pc_type", "gamg")
         self._push_managed_option("pc_gamg_type", "agg")
         self._push_managed_option("pc_gamg_repartition", True)
@@ -3989,7 +3989,7 @@ class SNES_Vector(SolverBaseClass):
         # Here we can set some defaults for this set of KSP / SNES solvers
         self.petsc_options["snes_type"] = "newtonls"
         self.petsc_options["ksp_rtol"] = 1.0e-3
-        self.petsc_options["ksp_type"] = "gmres"
+        self._push_managed_option("ksp_type", "gmres")
         self._push_managed_option("pc_type", "gamg")
         self._push_managed_option("pc_gamg_type", "agg")
         self._push_managed_option("pc_gamg_repartition", True)
@@ -5001,7 +5001,7 @@ class SNES_MultiComponent(SolverBaseClass):
         # PETSc options — mirror SNES_Vector defaults
         self.petsc_options["snes_type"] = "newtonls"
         self.petsc_options["ksp_rtol"] = 1.0e-3
-        self.petsc_options["ksp_type"] = "gmres"
+        self._push_managed_option("ksp_type", "gmres")
         self._push_managed_option("pc_type", "gamg")
         self._push_managed_option("pc_gamg_type", "agg")
         self._push_managed_option("pc_gamg_repartition", True)
--- a/src/underworld3/utilities/custom_mg.py
+++ b/src/underworld3/utilities/custom_mg.py
@@ -595,7 +595,8 @@ def _assert_no_zero_columns_serial(P_csr, level):
             f"operator would be singular.")
 
 
-def _configure_pcmg(pc, Ps, coarse="redundant", smoother="robust", owned=None):
+def _configure_pcmg(pc, Ps, coarse="redundant", smoother="robust", owned=None,
+                    ksp=None):
     """Reconfigure ``pc`` as a fresh PCMG (FMG F-cycle) driven by the supplied
     reduced->reduced prolongations ``Ps``, Galerkin RAP for coarse operators.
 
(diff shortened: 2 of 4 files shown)


lmoresi (commented):

**Where we attacked it:**

1. **The inert-fix trap — caught by our own test's first run.** The obvious fix (put `ksp_type` in the bundle) does nothing on the standard path: the top-level KSP consumes its options before injection, so the DB write never takes. The live-object application in `_configure_pcmg(ksp=...)` is the actual fix; the live test asserts on `ksp.getType()`, not the database, for exactly this reason.
2. **Ownership latch semantics.** A user's `petsc_options["ksp_type"] = "cg"` survives: the latch sees an unrecorded key and the bundle backs off; the live application then applies the *user's* value (a no-op setType). The three converted default writers are internal writers of a now-bundle key, which #471's own doctrine says must be recorded — this was a pre-existing violation of that rule, not a new pattern.
3. **Negative control on the new live test.** On the 2-D fixture the broken and fixed pairings are indistinguishable (3.3e-8 both, one FMG application — measured before setting the gate), so the error assert is a sanity bound only and says so; the discriminating gate is `test_0842` on the 3-D subsampled hierarchy, which goes 1.0e-6 → 2.7e-10 with this fix and is part of PR #510's CI.
4. **Blast radius.** `fast`/richardson variant: stationary smoother, fgmres merely judges on the true residual — sound, small extra memory. Fieldsplit velocity sub-KSP: bundle value equals current practice (fgmres, per the code's own comment). Rotated path: writes fgmres unconditionally already, untouched. GAMG fallback: `ksp_type` is in its stale list, so falling back clears our key and returns the default — verified by the bundle unit test.

Recommend merge; #510's re-run against post-merge development is the integration proof.

PR #517, opened 2026-08-09
place_sheet: the 3-D placed surface (serial), with a gated gmsh cavity fill

lmoresi:
The 3-D single-cut surface on an existing mesh — the capability the fault streams were waiting on. A triangulated sheet's points become mesh vertices and every sheet triangle a labelled interior face, **with the rim free inside the mesh**: the case a cut cannot represent and a two-sub-volume construction cannot fake.

## Mechanism

Carve (clearance + crossed tets, every victim's whole star), fill, sew on a compacted vertex set, carry every label **topologically** (vertex numbering through the rebuild is defined, so faces and edges transfer by vertex tuple — coordinates are never consulted). The fill is delegated to gmsh — a cavity shell has no total order to walk, and constrained tetrahedralisation with Steiner insertion is what a mesh generator is for — and **gated, not trusted**: both constraint surfaces bit-identical, every sheet triangle an interior face, conformity / Euler / degeneracy / wall-label preservation checked per call.

Measured basis: `sheet_cavity_spike.py` (study dir), 6/6 across tilt, resolution, clearance and **sub-h sheet spacing** — gmsh inserting its own Steiner points without touching either constraint, which retires the plan's constrained-tetrahedralisation STOP-gate.

## Guards, ported from the 2-D lessons

Cells owning an interior labelled face are held out of the cavity; a sheet needing them is refused; every earlier surface's interior-face count is **re-read off the result mesh** (a partial corruption satisfies any identity summed over per-placement bookkeeping — learned the hard way in 2-D).

## Scope

Serial by design: the parallel route is gather-first (redistribute so the sheet's star is rank-interior — the contact stream's measured policy — then this operation *is* the rank-local step, and no placed point is ever shared). `reconnect.py` untouched; the chart-surgery generalisation belongs to the parallel stage. Test number 0854 claimed in the coordination ledger.

## Handoff

The deliverable contract for `split_along_label_3d` (fault-split-node): a labelled facet patch — every sheet triangle a face under one `(name, value)`, interior with two cells, an open manifold with single rim edges — asserted as `test_0854` (5 tests). Full surgery regression: 77 tests green (0844 x2, 0853, 0854); style gate clean.

### DIFF
--- a/src/underworld3/utilities/place_surface.py
+++ b/src/underworld3/utilities/place_surface.py
@@ -79,10 +79,22 @@ of a plateau on which the walk has nothing left to order by.
 
 Scope
 -----
-Two dimensions, serial. Placement ADDS points, and the chart-expansion rebuild a
-shared point would need does not exist yet, so a parallel call is refused rather
-than silently returning a mesh whose star-forest is wrong. See
+Serial. Placement ADDS points, and the chart-expansion rebuild a shared point
+would need does not exist yet, so a parallel call is refused rather than
+silently returning a mesh whose star-forest is wrong. The parallel route is
+gather-first — redistribute so the fault's star is rank-interior, operate
+locally, renumber — at which point no placed point is ever shared. See
 :func:`~underworld3.utilities.reconnect.rebuild_cavities`.
+
+Both dimensions, by different fills. The 2-D curve (:func:`place_along_lines`)
+fills its cavity with the arc-length walk. The 3-D sheet (:func:`place_sheet`)
+delegates the fill to gmsh — a cavity shell is a 2-D surface with no total
+order to walk, and constrained tetrahedralisation with Steiner insertion is
+exactly what a mesh generator is for. The delegation is GATED, not trusted:
+both constraint surfaces are checked bit-identical, the sheet's triangles are
+checked present as interior faces, and conformity / Euler / volume are checked
+on every call (measured basis:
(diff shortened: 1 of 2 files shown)


lmoresi (commented):

**Where we attacked it:** (1) The fill's trust boundary — gmsh output is gated per call (constraint surfaces bit-identical, sheet triangles all interior faces, conformity/Euler/degeneracy/wall counts); the inert-DB-write and misattributed-crash lessons from #515/#510 are why nothing here trusts a library return without reading the result back. (2) The silent-corruption class from 2-D — both cells of a labelled facet cleared while every corner is protected — is guarded with faces standing where edges stood, and the too-close-sheet test accepts *either* refusal path (held-cell or shell-manifold) while asserting the first sheet is intact afterwards, since which guard trips first is geometry-dependent. (3) Label transfer is topological (vertex tuples through a defined renumbering), never coordinate matching — the property the 2-D rebuilds were built on. (4) Serial-only refusal states the gather-first rationale so the parallel stage has its design recorded at the refusal site.

**Known limits, stated rather than discovered later:** one sheet per cavity (close pairs are one thin volume — the next work item); no cavity-growth retry loop in 3-D yet (a wedged fill raises with guidance rather than growing, unlike 2-D); `test_0854` asserts the handoff contract shape only — the wire-up to `split_along_label_3d` runs on the fault branch, per the ledger.

Recommend merge (squash — single-purpose branch, no shared ancestry).

PR #518, opened 2026-08-09
place_sheet runs in parallel: gather-first, one rebuild for every rank count

lmoresi:
Closes the last refusal on the 3-D placed surface, per the maintainer ruling that "planned, ready" parallel gaps fossilize. Serial and parallel share **one mechanism** — no fork for the serial suite to diverge from.

## Mechanism

**Gather-first.** The sheet's region is marked as an SF-reconciled function of the mesh (per-vertex h reduced over the star-forest, so no rank's verdict depends on the partition), gathered onto one rank by a shell partitioner (mask-driven port of the contact stream's fault-aware redistribution, `c8693579`/`1d487319`, recorded in the ledger — the label-driven original is measured at np=2–8), and the serial carve-and-fill runs there as the rank-local step.

**The rebuild** is the uninterpolated-cells + `DMPlexInterpolate` pattern (their 3-D splitter's move, same ledger entry): only cell→vertex cones are wired by hand, PETSc derives every face/edge/orientation, and old faces and edges are recovered by joining surviving vertex tuples — labels transfer topologically, no tet orientation hand-encoded, ever. The gathered region is rank-interior, so nothing the surgery touches is shared: the leaf set is unchanged and `_rebuild_point_sf` renumbers in one exchange. **Partition-independent by construction** — the fill sees the identical cavity whatever the incoming partition.

## Two parallel defects found by the sweep, fixed, regression-tested

1. **np≥4 hang**: the gather can empty a rank, and `cell_diameters` raises on one — a rank-local death while peers entered the next collective (pytest-timeout couldn't even kill it). Empty ranks now contribute the identity of every reduction; the refusal ptest at np=4/5 regression-tests it and simultaneously proves every refusal collective (all ranks raise the identical error, or none does).
2. **Seam faces masquerading as walls**: a partition-seam face also has local support 1; wall = support 1 **and unshared**, vertex mask OR-reconciled.

## Verification

- `ptest_0854` at np=2, 3, 4, 5 (version-stable gates: facet count ≡ input triangles, volume conserved, global Euler 1 from owned counts, walls preserved as owned sums, info dict identical on every rank).
- Serial: 87 tests across 0843/0844×2/0853/0854/0845, unchanged and green on the unified path; 2-D parallel suites at np=2/3 green; style gate clean.
- Test number 0854 (parallel file) already claimed in the ledger.

### DIFF
--- a/src/underworld3/utilities/place_surface.py
+++ b/src/underworld3/utilities/place_surface.py
@@ -102,6 +102,7 @@ interface, so a later placement will not delete a vertex out of an earlier one.
 """
 
 import numpy as np
+from mpi4py import MPI
 from petsc4py import PETSc
 
 import underworld3 as uw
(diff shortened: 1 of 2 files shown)


lmoresi (commented):

**Where we attacked it:** (1) The happy path is not evidence — the np sweep went to 5 ranks and the refusal path is a first-class test, which is what caught the np≥4 hang (gather empties a rank; `cell_diameters` raises on one; the rank died inside a collective chain and even pytest-timeout could not kill the survivors — diagnosed from the partial dump's petsc garbage-cleanup MPI error). Empty ranks now contribute the identity of every reduction, and the refusal ptest doubles as the collectivity proof: all ranks raise the identical message or none does. (2) Seam-vs-wall: local support 1 misclassifies partition seams as domain walls; the test is support 1 AND unshared, OR-reconciled. (3) Partition-dependence: per-vertex h is SF-reduced before any marking decision, so which vertices are victims is a function of the mesh, never the partition; the gathered cavity is therefore identical from any incoming distribution, and the ptest asserts the info dict identical on every rank. (4) No serial/parallel fork: one rebuild path, so 87 serial tests exercise the machinery the parallel path runs.

**Taken, with attribution (ledger rule 2):** the shell-partition gather + SF-propagate (fault-split-node c8693579/1d487319, mask-driven port) and the uninterp+DMPlexInterpolate rebuild with getFullJoin recovery (their 3-D splitter) — the move that means no tet cone orientation is ever hand-encoded.

**Known limits, stated:** load concentrates on the surgery rank for the duration of the placement (bounded by the fault region; per-rank fault patches are the recorded escalation); `edge_split.cell_diameters` still raises on an empty rank — guarded at our call sites, latent for other parallel callers (flagged in the ledger rather than drive-by-fixed, Charter S9); gather-mask margin is +2 h with a loud collective assert if it ever under-reaches.

Recommend squash-merge (solo branch).

lmoresi:
We reviewed this against the fault-contact socket (place -> split_fault -> contact -> solve) and pushed fixes for what we found (8ff3659b). Verdict: with those fixes, the PR meets the contract and is ready to merge (as a MERGE COMMIT, per the shared-ancestry rule from #510).

**Finding 1 (correctness, fixed): the sewn mesh had mixed cell handedness.** The parallel rewrite dropped the serial path's signed-volume flip; DMPlexInterpolate derives face cones but never normalises cell vertex order, and gmsh's fill tets arrive opposite-handed to the kept cells. Evidence: 4131 vs 443 signed volumes on the sewn mesh; the first Stokes solve (plain, no fault — the isolation control) never converged while every abs()-based volume gate stayed green; the same mesh with cells re-oriented solves in 22 s. Fix: made cells oriented against the kept cells' own sign at wiring, plus a global orientation census gate — abs() volume checks are structurally blind to this defect class.

**Finding 2 (gates, added): PETSc's DMPlex validity battery as the final pass** (maintainer request). Full set in serial — check_geometry is exactly the instrument that catches Finding 1 on day one. In parallel only the partition-safe subset (symmetry, skeleton, point-SF): we measured check_geometry false-positive on every rank of a PLAIN distributed UnstructuredSimplexBox, so it cannot gate distributed results on this stack. Gate failures are allgathered before raising, per this module's own collective-refusal discipline.

**Finding 3 (contract, our side of the socket): sheet triangulations must have no all-rim triangles** — the splitter refuses them (both copies of such a face would carry the same vertex triple), and structured grids produce them at corners. Conditioning by centroid-fanning creates sub-h facets that give the contact a residual floor (measured: rel 1.1e-2 stall persisting under LU, so not solver conditioning); conditioning by flipping the corner-quad diagonals is clean. Recorded on the fault-network side; noted here because sheet providers will hit it.

**Acceptance on the fixed build**: place -> split -> frictionless contact converges in 18 s (faster than the gmsh-embed control, 22 s), no-opening leak 8e-18, sensible slip taper; serial suites 26 green; ptest_0854 green at np=2 and np=4. One honest gap for the ledger: the parallel gates here are topological — a parallel FE acceptance through the splitter waits on the contact stream's pairing-through-redistribution work, which is ours, not this PR's.

lmoresi:
Follow-up from the composed acceptance (place -> split -> contact, and its isolation ladder): there is a SECOND parallel defect, real and reproducible, in the rebuilt mesh's higher-strata consistency. We are handing this one back with the evidence rather than fixing it, since it sits in the rebuild's design.

**Symptom**: plain Stokes (no fault machinery at all) on the placed mesh at np=2 takes 3594 s; the identical mesh in serial solves in 22 s; the plain BASE box at np=2 solves in 13 s. The solve converges eventually — the operator is assembled wrong, not singular.

**What it is NOT** (each eliminated by measurement): not the contact machinery (plain Stokes grinds); not element quality (radius-ratio distribution matches the base mesh); not cell handedness (the orientation census gate passes); not missing/duplicated points (global owned v/e/f/c identical serial vs np=2: 1138/6444/9881/4574); not vertex SF corruption (coordinate drift exactly 0); not the partition (a FRESH DMPlexDistribute of the placed mesh — canonical SF, healthy seam counts — still takes 2212 s); not the solver config (base box np=2 is fast).

**What the evidence points at**: DMPlexCheckFaces FAILS on the rebuilt mesh at np=4 (rank 3: face closure 1462/1456/1418 vs cell face 1456/1462/1418 — same vertex set, swapped order = a cone-orientation mismatch), while the plain distributed base box passes CheckFaces cleanly. We must retract part of our earlier gate scoping: only check_geometry false-positives on healthy distributed meshes; check_faces failing on the REBUILT mesh is a true positive. Inconsistent face/edge cone orientations across the point SF disorder the P2 edge/face DOF numbering during assembly — which produces exactly this signature, and which serial cannot see (no seam) and redistribution cannot heal (it faithfully migrates the local cones).

**Where to look**: the collective uninterpolate -> DMPlexInterpolate rebuild interpolates each rank's chart locally; nothing afterwards reconciles the leaf faces'/edges' cone order against their roots. The 2-D module never met this because its surgery output is consumed rank-locally; the FE assembly across the seam is what demands root/leaf cone agreement. Suggested acceptance for the fix: CheckFaces green on the rebuilt mesh at np=2..5, and plain Stokes on a placed mesh at np=2 within ~2x of the base box (ours: 13 s).

**Merge posture**: this does not block the serial capability (fully validated, 18 s acceptance) or the parallel topological layer (gates green np=2-5). It DOES block parallel FE work on placed meshes. Our recommendation: merge, with this defect recorded as the ledger item gating "parallel placed-mesh FE", or hold for the cone-reconciliation fix — maintainer's call.

lmoresi:
The parallel higher-strata defect from the review dossier above is now tracked as #520 (with reproducer and acceptance criteria) — it survives #519 unchanged, measured on development at eb8d2d30, and gates parallel FE on both placed sheets and thin volumes. Serial capability unaffected.

PR #521, opened 2026-08-10
Fix #520: seam cone reconciliation + the recovered orientation fix — parallel FE on placed meshes works

### LINKED ISSUE #520: place_sheet parallel rebuild: face/edge cone orientations unreconciled across the point SF — parallel FE on placed meshes assembles a wrong operator
Split out of the PR #518 review (comments there, 2026-08-10) so it cannot be lost: the composed acceptance found a parallel-only defect in the gather-first rebuild that blocks FE work on placed meshes at np>1. It survives #519 unchanged (measured on development at eb8d2d30), and #519's thin-volume embedding shares the same rebuild skeleton, so its parallel FE path inherits it.

## Symptom

Plain Stokes — no fault machinery — on a placed-sheet mesh:

| configuration | time |
|---|---|
| base box, serial | 22 s |
| base box, np=2 | 13 s |
| placed mesh, serial | 22 s |
| **placed mesh, np=2** | **3594 s** (converges — wrong, not singular) |
| placed mesh, np=2, after a FRESH DMPlexDistribute | 2212 s |

## Eliminated by measurement

Element quality (radius-ratio distribution matches base); cell handedness (the #518 orientation census gate passes); duplicated/missing points (global owned v/e/f/c identical serial vs np=2: 1138/6444/9881/4574); vertex SF corruption (coordinate drift 0.0 exactly); the partition (fresh canonical redistribute does not cure it); the solver configuration (base box np=2 is fast); the fault contact machinery (plain Stokes grinds).

## The true signal

`DMPlexCheckFaces` FAILS on the rebuilt mesh in parallel — np=2: both ranks (measured on development eb8d2d30); np=4: e.g. rank 3 with `face closure: 1462 1456 1418` vs `cell face: 1456 1462 1418` — same vertex set, swapped order: a cone-orientation mismatch. The plain distributed base box passes CheckFaces cleanly, so this is not a checker false-positive (unlike check_geometry, which does false-positive on healthy distributed meshes on this stack — see the gate scoping in 8ff3659b).

Mechanism: the collective uninterpolate -> DMPlexInterpolate rebuild interpolates each rank's chart locally, and nothing afterwards reconciles leaf faces'/edges' cone order against their roots. Inconsistent cone orientations across the SF disorder the P2 edge/face DOF numbering in cross-seam assembly: the operator is assembled wrong. Serial cannot see it (no seam); redistribution faithfully migrates the bad cones.

## Reproducer (np=2, ~2 min to the CheckFaces verdict)

```python
import numpy as np
import underworld3 as uw
from underworld3.utilities.place_surface import place_sheet
from petsc4py import PETSc

base = uw.meshing.UnstructuredSimplexBox(
    minCoords=(0, 0, 0), maxCoords=(1, 1, 1), cellSize=0.11,
    regular=False, qdegree=2)
u = np.array([1.0, 0.0, 0.25]); u /= np.linalg.norm(u)
v = np.array([0.0, 1.0, 0.0]); s = np.linspace(-0.2, 0.2, 5)
pts = np.array([np.array([0.5, 0.5, 0.5]) + a*u + b*v
                for a in s for b in s])
tris, n = [], 5
rim = lambda k: (k // n in (0, n-1)) or (k % n in (0, n-1))
for i in range(n-1):
    for j in range(n-1):
        a, b = i*n+j, i*n+j+1
        c, d = (i+1)*n+j, (i+1)*n+j+1
        t1, t2 = (a, b, d), (a, d, c)
        if all(map(rim, t1)) or all(map(rim, t2)):
            t1, t2 = (a, b, c), (b, d, c)   # no all-rim triangles
        tris += [t1, t2]
bounds = base._boundaries_with("Rupture")
placed, info = place_sheet(base.dm, pts, np.array(tris, dtype=np.int64),
                           label="Rupture",
                           label_value=bounds["Rupture"].value)
chk = placed.clone(); chk.setOptionsPrefix("cf_")
o = PETSc.Options(); o["cf_dm_plex_check_faces"] = ""
ok = 1
try:
    chk.setFromOptions()
except PETSc.Error:
    ok = 0
del o["cf_dm_plex_check_faces"]
oks = uw.mpi.comm.allgather(ok)
uw.pprint("CheckFaces: " + ("OK" if all(oks) else f"FAIL {oks}"))
# then wrap in Mesh and run plain Stokes to see the grind
```

## Acceptance

1. `DMPlexCheckFaces` green on the rebuilt mesh at np=2..5 (options route; it is a valid oracle here — the base mesh passes it).
2. Plain Stokes on a placed mesh at np=2 within ~2x of the base box (13 s on the review machine).
3. Same two checks for a `place_thin_volume` embedding (shared skeleton).

Serial capability is unaffected and fully validated (place -> split -> contact converges in 18 s, leak 8e-18). This issue gates parallel FE on placed/thin-volume meshes only.

lmoresi:
Closes #520 — and en route recovers an orphaned review fix without which even SERIAL FE on placed meshes was wrong.

## What #520's investigation actually found

Two stacked defects, one of them older and worse than the issue knew:

**1. The #518 review's finding-1 fix never landed.** The contact session pushed `8ff3659b` (orient fill cells to the kept convention + PETSc validity gates) to a review side branch; #518 was squash-merged without it and the branch deleted — the fix was orphaned. Consequence: every placed-sheet mesh since #517 had mixed cell handedness (measured again: 4131 cells one way, 443 the other — the review's exact numbers), and FE assembly on such a mesh is wrong **in serial**: a P2 Poisson with a LINEAR exact solution came back with errors of 27 under exact LU, violating the maximum principle. Nobody saw it because the only FE consumer (the fault-contact stream) re-rebuilds the plex through its own splitter, and #520's serial table row only recorded time. `8ff3659b` is cherry-picked here verbatim, authorship preserved.

**2. The #520 defect proper: interface cones unreconciled across the point SF.** The rebuild interpolated each rank's chart with NO star-forest attached and installed one afterwards — so nothing ever reconciled a leaf face/edge cone against its root. The fix is ordering: `_attach_uninterp_vertex_sf` gives the uninterpolated plex its vertex SF (one broadcast of the new numbering over the old SF — the leaf set is unchanged under gather-first), and `DMPlexInterpolate` then runs its distributed path — creating faces/edges, orienting interface cones consistently, and extending the SF itself. The after-the-fact `_rebuild_point_sf` call was the defect and is gone; the interpolated SF is mirrored onto the coordinate DM (the parallel-checkpoint property).

## Acceptance (issue #520's criteria)

1. **DMPlexCheckFaces green at np=2..5** on the placed mesh (was failing at every rank count) — and now an ALWAYS-ON internal gate: the recovered validity battery moves into `_validity_and_orientation_gates()`, shared by `place_sheet` and `place_thin_volume`, with `check_faces` promoted from serial-only to every rank count (the plain distributed box passes it cleanly; it is exactly this issue's oracle). `check_geometry` stays serial-only (measured false-positives on healthy distributed meshes).
2. **Plain Stokes on the placed mesh at np=2: 12.1 s vs base box 7.6 s** (1.6x, bound was ~2x; the defect measured 3594 s).
3. **Same checks for the thin volume**: shared skeleton, shared fix; P2 quadratic reproduced through the embedded zone to 9e-16 at np=1,2,4.

## The new oracle class

P2 Poisson must reproduce x²+y²+z² to solver precision on the sewn meshes — added to test_0854 and test_0855. This is the end-to-end assembly check that was missing when the handedness defect shipped: conformity, Euler, exact volume and every abs()-based gate are all structurally blind to a wrong operator; a contained exact solution cannot be faked. Serial suites 38 green; ptest_0854+0855 green at np=2..5 with the check_faces gate firing inside every placement call.

### DIFF
--- a/src/underworld3/utilities/place_surface.py
+++ b/src/underworld3/utilities/place_surface.py
@@ -1572,6 +1572,35 @@ def _rebuild_sewn_3d(dm, drop_cell_ids, victim_ids, made_cells, placed):
     nc_new = int(keep_cell.sum()) + len(made_cells)
     nv_new = n_surv + len(placed)
 
+    # Orient every made cell to the KEPT cells' handedness before wiring.
+    # DMPlexInterpolate derives face cones and orientations from the
+    # cell-vertex cones but does NOT normalise the cells' own vertex
+    # order; the fill's tets arrive in gmsh's convention, which is
+    # opposite to the plex closure convention the kept cells carry, and
+    # a mixed-handedness mesh assembles negative Jacobians (measured:
+    # the first Stokes solve on the sewn mesh never converged, while
+    # every abs()-based volume gate stayed green). The serial rewrite
+    # dropped the old explicit flip; this restores it, against the kept
+    # cells' own sign so the convention is read off the mesh, not assumed.
+    made_cells = np.asarray(made_cells, dtype=np.int64).reshape(-1, 4)
+    if len(made_cells) and len(kept):
+        X_old = _coords(dm)[:nv_old]
+
+        def signed6(P):
+            return float(np.dot(np.cross(P[1] - P[0], P[2] - P[0]),
+                                P[3] - P[0]))
+
+        ref_sign = np.sign(signed6(X_old[kept[0]]))
+
+        def xyz_of(x):
+            return placed[-int(x) - 1] if x < 0 else X_old[int(x)]
+
+        made_cells = made_cells.copy()
+        for j, tet in enumerate(made_cells):
+            P = np.array([xyz_of(v) for v in tet])
+            if np.sign(signed6(P)) != ref_sign:
+                made_cells[j] = tet[[0, 1, 3, 2]]
+
     def v_uninterp(x):
         if x < 0:
             return nc_new + n_surv + (-int(x) - 1)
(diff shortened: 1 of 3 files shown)


lmoresi:
Adversarial review (self, posted per project practice). The verification ladder, and what each rung eliminated:

**The probe was wrong before the mesh was.** Our first Stokes probe silently swallowed BC failures (try/except around add_dirichlet_bc) and timed unconverged solves; a second grep-filtered its own diagnostics away. Both violations of our own probe hygiene, both caught by their numbers making no sense. The probe that finally cut through was the exactness oracle: P2 contains quadratics, so a correct assembly RETURNS them — no baselines, no timing, no tolerance debate.

**Attribution was measured, not assumed.** The serial wrongness reproduced bit-identically on landed eb8d2d30 with the working tree stashed — so it predates this branch and is not our SF change. The fresh-plex control (createFromCellList from the same cells+coords: identical error) placed the defect in the DATA, and the signed census then produced the review's own numbers, 4131/443 — which is what sent us looking for the orphaned commit rather than a new mechanism.

**The recovered fix is theirs, verbatim.** 8ff3659b cherry-picked with authorship preserved; our additions are the refactor into a shared gate (both primitives), the check_faces promotion to all rank counts, and the FE oracles. Credit and the process lesson go together: a review-fix pushed to a side branch after a squash-merge has no path into the trunk — the ledger now says review fixes land as PRs, not branch pushes.

**Residual risks, stated:** (1) DMPlexInterpolate's distributed path is now load-bearing for cone reconciliation — we verify its output (check_faces, every call, every rank count) rather than trusting it. (2) The FE oracle uses diffusivity 1 everywhere; a coefficient-jump oracle across the zone skin would additionally pin the label geometry — deferred, the labels are join-recovered and count-gated. (3) The parallel FE proof in CI rides on the in-call gates plus serial exactness tests; a full parallel solve test was weighed and dropped to keep the suites lean — the gate that found this defect now fires inside every parallel placement test.

PR #523, opened 2026-08-11
2-D placement goes parallel: one rebuild for every dimension and rank count

lmoresi:
Second half of the lifecycle ruling (no redistribution, ever), first instalment: the 2-D placement family joins the 3-D one on the gather-first machinery.

- **One rebuild everywhere.** `_rebuild_sewn` is now dimension-generic (cone size dim+1; the handedness flip reads signed area in 2-D; join recovery over edges alone) and both 2-D serial paths route through it — `reconnect.rebuild_cavities` leaves the placement family. The handedness census generalises with it (signed 2×area; abs() is blind to inversion in any dimension).
- **`place_along_lines` runs distributed**: mark by distance → gather the region to one rank → carve and fill there → collective rebuild (vertex SF attached before interpolate, the #520 property) → label by the chain's new ids → global gates (area conserved, inversion, the validity battery with check_faces at every rank count, held surfaces compared per rank — the mesh outside the region is untouched, so per-rank equality is exactly the right gate).
- **Scope stated, not fudged**: interior surfaces — the fault case. A wall-touching end needs the end-settling machinery (wall-vertex slides, facet splits) whose collective form is not built; it refuses collectively with the reason, and the serial path keeps the full feature set. Wall proximity is judged against TRUE wall segments (support 1 AND unshared — a seam edge also has local support 1; the 3-D lesson one dimension down).

Measured: same facet count as serial and P2 FE exactness (x²+y² to 1e-11) at np=1..4. ptest_0853 added (FE oracle + collective wall refusal, since a rank-local raise is a hang at np≥3); all serial suites (42) and the 3-D ptests green.

Next instalment on this line: the 2-D ribbon (`place_thin_volume`) and 2-D removal on the same skeleton, then the distributed add–solve–remove–add composition test.

### DIFF
--- a/src/underworld3/utilities/place_surface.py
+++ b/src/underworld3/utilities/place_surface.py
@@ -59,12 +59,13 @@ ordinary free-end embed.
 
 Scope
 -----
-Serial. Placement ADDS points, and the chart-expansion rebuild a shared point
-would need does not exist yet, so a parallel call is refused rather than
-silently returning a mesh whose star-forest is wrong. The parallel route is
-gather-first — redistribute so the fault's star is rank-interior, operate
-locally, renumber — at which point no placed point is ever shared. See
-:func:`~underworld3.utilities.reconnect.rebuild_cavities`.
+Serial and parallel, gather-first (the fault's star is redistributed onto
+one rank, the surgery runs there, every rank rebuilds collectively; no
+placed point is ever shared). One parallel restriction in 2-D: a surface
+whose END reaches the domain wall carries the end-settling machinery
+(wall-vertex slides, facet splits) whose collective form is not built —
+refused with the reason; interior surfaces, the fault case, run at any
+rank count.
 
 Both dimensions, one fill. The 2-D curve (:func:`place_along_lines`) and the
 3-D sheet (:func:`place_sheet`) both delegate their cavity fill to gmsh —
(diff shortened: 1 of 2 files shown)


lmoresi:

**What broke and taught.** The handedness census was 3-D-only in disguise (`_tet_vertices` inside `_cell_volumes_signed6`) — caught at np=2 on the first 2-D parallel run because the parallel path is the only 2-D path that runs the gates. Serial 2-D never called them; it does not yet either (the outer serial gates predate the battery). Follow-up candidate: route serial 2-D through `_validity_and_orientation_gates` too, once the ribbon instalment unifies the orchestration.

**The held gate's per-rank form is an argument, not an approximation.** A global count of interface edges needs owned-only reduction machinery; per-rank equality is strictly stronger here BECAUSE the surgery region is rank-interior — every other rank's counts must be bit-unchanged, and the target rank's may grow only in the label being written. If a future change lets surgery touch shared points, this gate must be revisited — the gather-contract assertion (no touched point is shared) sits directly upstream and would fire first.

**Scope honesty.** Interior-only in parallel is a real restriction, chosen over a half-collective end-settling that would move shared wall vertices without SF-propagating coordinates. The refusal is collective and names the serial path. Ribbons (interior by construction) and removal inherit no such restriction and arrive next.

PR #524, opened 2026-08-11
2-D ribbon + removal go parallel; the fault lifecycle composes in both dimensions

lmoresi:
Final instalments of the lifecycle ruling: every placement-family operation — line, sheet, ribbon, thin volume, removal — now runs serial AND parallel through one gather-first mechanism and one dimension-generic rebuild, with no redistribution anywhere.

- **`_place_thin_volume_2d` in collective form**: rank-0 assembly + broadcast, mark by skin distance, gather, target-rank carve (ring growth, orphan promotion), holes fill, the one rebuild, cell+skin labels by join, collective gates including the validity battery. Serial is the same path at size 1; the last 2-D serial refusal in the volume family is gone.
- **2-D removal** on the same skeleton: label-seeded ring carve, plain background-scale fill, labels gated globally empty; the face soup and interface readers generalise (edges as 2-D facets, plus the exclusion that stops a removal holding its own object against itself).
- **The lifecycle is the tested contract, both dimensions**: zone → FE-exact → remove → FE-exact → place a line in the SAME region → remove again, np=1..4, zero area drift. Serial composition test in test_0856; distributed 2-D and 3-D composition tests in ptest_0856, each stage held to the P2 quadratic-exactness oracle.

Suites 0853–0856 and all four parallel test files green at np=2 and 4. With this, the fat-fault session's mesh-machinery ledger is clean: add a stack, delete a stack, on the running distributed mesh, in either dimension — the remaining scoped exception (wall-touching 2-D line ends, serial-only) is refused with its reason and recorded.

### DIFF
--- a/src/underworld3/utilities/place_surface.py
+++ b/src/underworld3/utilities/place_surface.py
@@ -2812,14 +2812,17 @@ def _gmsh_fill_plain_3d(shell_xyz, shell_tris, h):
 
 
 def _labelled_face_soup(dm, names_values):
-    """The (points, triangles) soup of every labelled face, ALLGATHERED.
+    """The (points, facets) soup of every labelled facet, ALLGATHERED.
 
     A removal must mark against the object's global geometry whatever the
     current distribution — the object was embedded rank-locally, but a
     checkpoint reload or a later redistribution may have scattered it. The
     soup is small (an object's skin) and duplicates across strata are
-    harmless to a distance query.
+    harmless to a distance query. Facets are triangles in 3-D and edges in
+    2-D; the soup's connectivity is trivial (one facet per row).
     """
+    dim = dm.getDimension()
+    nvf = dim                       # vertices per facet
     vS, vE = dm.getDepthStratum(0)
     fS, fE = dm.getHeightStratum(1)
     X = _coords(dm)[: vE - vS]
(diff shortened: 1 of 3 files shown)


lmoresi:

**What we attacked**: the ribbon rewrite replaces a working serial path wholesale with the collective form — the risk is a silent behaviour change hiding behind passing tests. The guard is threefold: the serial suite (0855, 11 tests incl. kink quality and second-zone safety) runs the SAME code at size 1; the X-junction produces identical zone/skin counts (114/92) at np=1..4; and every stage of the composition is held to the quadratic-exactness oracle, which no chart, cone or SF disorder can pass.

**A real behaviour change, declared**: the parallel-capable paths measure local h as min incident cell diameter (SF-reconciled) where the old serial ribbon used edge-based vertex h — cavity extents shift slightly. The suites are insensitive by design (counts not pinned, quality thresholded); anyone pinning cavity sizes downstream would notice, and nothing does.

**The exclusion readers' known blind spot**: an edge carried by BOTH the removed label and another interface label is unlocked during a removal's carve. Documented at the definition; the other label's own count gate is what catches a genuine overlap, and the composition tests exercise exactly that adjacency (line placed where a zone was).

**Not done, stated**: wall-touching 2-D line ends remain serial-only (collective end-settling unbuilt — refusal names it); the serial 2-D LINE path still runs its pre-battery gates rather than _validity_and_orientation_gates (the ribbon and removal paths run the battery at any size now); refill cell counts remain partition-sensitive and outside the contract.

PR #525, opened 2026-08-11
Outcropping faults: sheets meet the upper surface (specify-long, bowl cavity, remeshed cap)

lmoresi:
The science case the fault meshes exist for: plate-boundary faults intersect the free surface. Spike-proven first (`outcrop_sheet_spike.py`, two dead ends measured and recorded), then production in `place_sheet`, serial and parallel.

- **Specify-long contract** (ruling 2026-08-11): sheets may extend past the domain. `_clip_sheet_to_box` clips each triangle against the box planes — cut points snap exactly onto the plane, computed in canonical edge order so shared edges dedupe bitwise — and `_outcrop_chain` extracts the trace as one ordered polyline on one wall. Box-edge and multi-wall outcrops refuse with the reason (recorded residuals, with curved/deformed tops — the manifold-prep slot).
- **The bowl cavity**: wall vertices inside the cap become victims; `_closed_shell_3d` returns dropped cells' wall faces as CAP faces (other wall contact still refuses); the manifold check runs on shell ∪ cap. The cap is **pre-meshed by the 2-D fill** in the wall plane's own coordinates (rim verbatim, outcrop embedded) because a geo surface bounded by the discrete rim RESAMPLES it — measured, and `MeshOnlyEmpty` does not prevent it. The 3-D fill takes the loop [shell + discrete cap] with the sheet embedded, outcrop nodes shared by tag.
- **Explicit relabelling**: the cap's new wall faces carry exactly the (label, value) pairs the replaced faces carried, with full closures — joins cannot recover new points, and an unlabelled boundary patch silently loses its Dirichlet conditions.

Measured, serial and np=2..4: identical facet counts, P2 quadratic FE-exactness (2e-11–5e-11) **with a Dirichlet condition gripping the remeshed Top**, and full Top-label coverage asserted face-by-face. Interior sheets byte-for-byte unaffected: suites 0853–0856 (44 serial) and the placement ptests green. New contract test in test_0854. The outcropping thin-volume BAND (cap-with-hole + interior skin as one closed loop) is the remaining instalment; the plan file carries its design.

### DIFF
--- a/src/underworld3/utilities/place_surface.py
+++ b/src/underworld3/utilities/place_surface.py
@@ -1528,7 +1528,8 @@ def _gather_region(dm, vertex_mark_chart, verbose=False):
 
 
 def _carve_cavity_3d(dm, X, cells, sheet_pts, sheet_tris, clearance,
-                     held_cells, h_vertex, on_wall, shared_chart):
+                     held_cells, h_vertex, on_wall, shared_chart,
+                     open_wall=None):
     """Victims, dropped tets and the closed cavity shell around the sheet.
 
     The same two-part rule as 2-D — vertices within the clearance go, and any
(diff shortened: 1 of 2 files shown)


lmoresi:

**What we attacked**: the cap is the first fill piece whose triangulation is OURS rather than gmsh's-under-gates, so its two failure modes got explicit probes. Resampling: the geo-surface route produced 84 renumbered rim nodes / volume 1.31 / Euler −46 before the pre-meshed discrete cap; the moved-node gate now covers rim and chain, and the cap-triangle-count gate refuses a remeshed cap. Label loss: the coverage assertion in test_0854 walks every boundary face on the top plane — an unlabelled patch cannot pass, and the FE oracle with Dirichlet on the remeshed Top would diverge from the quadratic if the condition lost grip anywhere.

**Scope honesty**: one flat wall per sheet; box-edge outcrops, multiple traces, closed traces, and curved/deformed tops refuse with their reasons (the last is the manifold-prep slot and matters for free-surface runs — recorded in the plan). The clip has no projection band yet: a sheet stopping short of the wall is honestly buried, not snapped; the band is a stated refinement.

**Process note for the record**: the working-directory drift bit again mid-landing (the branch was created in the wrong checkout and the main repo briefly sat on a feature branch — restored to development immediately). All git operations in this session now name their repo explicitly.

PR #526, opened 2026-08-11
Outcropping zones: the fault zone leaves its band on the surface

lmoresi:
Completes the outcrop implementation (with #525): the thin-volume form. A fault ZONE specified past the box is clipped in OCC; its clipped face becomes the BAND on the surface; the gap's boundary is one closed surface of three discrete pieces (interior shell + cap-annulus with the band outline as its hole, pre-meshed by the 2-D fill + interior skin, rim-matched). Band faces carry BOTH the zone's skin label and the wall's labels; the cap annulus is relabelled explicitly with full closures, counts gated.

Measured serial and np=2,4: identical zone counts, P2 FE-exactness (5e-11–7e-11) with Dirichlet gripping the remeshed Top, full Top-label coverage face-by-face. Interior zones byte-for-byte unaffected — suites 0853–0856 (45 serial) and the placement ptests green. Band contract test in test_0855.

Stated residuals (refuse with reasons): box-edge bands / multi-wall contact; curved or deformed tops (the manifold-prep slot); removal of an OUTCROPPING object (its cavity legitimately touches the wall — the removal carve does not yet take open_wall; next small instalment if needed).

### DIFF
--- a/docs/developer/subsystems/conforming-surfaces-and-fault-zones.md
+++ b/docs/developer/subsystems/conforming-surfaces-and-fault-zones.md
@@ -331,6 +331,30 @@ object can **pinch** — the drop set is grown at every non-manifold shell
 edge or pinch vertex until the boundary is simple, which only enlarges the
 fill.
 
+## Outcropping faults: meeting the free surface
+
+Plate-boundary faults intersect the upper surface — the science case this
+machinery serves — and both representations support it under the
+**specify-long contract**: define the fault generously PAST the domain, and
+prep clips it. A sheet is clipped triangle-by-triangle (cut points exactly
+on the plane); a zone's assembly is intersected with the box in OCC. The
+trace left on the wall becomes the **outcrop** (a sheet's chain, a zone's
+band), the cavity opens onto that wall as a bowl, and the patch of wall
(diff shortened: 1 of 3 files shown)


lmoresi:

**The three-piece boundary is the risk we attacked.** The gap's closed surface is assembled from three independently produced triangulations sharing nodes only by tag identity: shell↔cap at the wall rim, cap↔skin at the band outline. Every sharing is gated — moved-nodes zero across all constrained sets, cap/skin triangle counts refused if remeshed, and every band and cap face recovered by join with the relabel count compared against the expectation. The FE oracle with Dirichlet on the remeshed Top then closes over the whole chain: a single mis-sewn edge or unlabelled patch cannot reproduce the quadratic.

**A subtlety we verified rather than assumed**: OCC's clipped faces land within rounding of the wall plane, not on it; the band logic keys on EXACT plane values, so assembly nodes within 1e-9 snap onto the plane before anything reads them. Without the snap the band detection silently returns empty and the placement refuses at the wall — loud, but wrong.

**Stated residuals**: removal of an outcropping object is the one asymmetry left (the removal carve does not yet pass open_wall — a spent outcropping fault currently refuses removal rather than clearing its band); box-edge bands and curved tops refuse with reasons as before.

PR #530, opened 2026-08-12
Adapt-review fixes: collective pinned-band verdicts, partition-independent pinning, explicit-GAMG opt-out, the #492 CI disarm

lmoresi:
# Parallel pinned-band and custom-MG opt-out fixes from the #488 review

These are the findings from the #488 adversarial re-review, extracted as an
independent branch after #510/#511 superseded that PR. Each is a measured
defect: the np=4 pinned-band hang, 4.2e-3 motion of a pinned vertex, a crash
on an explicit GAMG choice, the #492 rebuild trap behind the test_0842 CI
segfault, and bare "Segmentation fault" CI logs. Rebased cleanly onto
development at 606583eb (#515–#526: place_sheet 3-D, parallel placement,
lifecycle, outcropping zones) — no conflicts; none of the five fixes gained a
development-side counterpart (`label_interface_band` there still has no
allreduce).

## What changed

1. **Collective pinned-band emptiness** (`label_interface_band`,
   `discretisation_mesh.py`). The band-emptiness check raised rank-locally, so a
   surface confined to one rank's corner at np=4 left three ranks raising while
   the fourth entered the collective mover and hung to the 300 s timeout
   (measured). The pinned count is now allreduced; the error raises on every
   rank only when the band is globally empty; the DMLabel is created on every
   rank, including band-less ones.

2. **Partition-independent pinned band** (same function). The straddle test and
   halo-ring growth walk rank-local cells, so a shared vertex whose cut cell
   lives on the neighbour rank was pinned there but not on the owner — the
   mover moved it and the pinned copy followed through the SF (measured at
   np=4: two pinned leaves moved 4.2e-3 and 1.9e-3; np=2/3 passed on partition
   luck). The pinned set is now synchronised across ranks by rounded coordinate
   after the core band and after each halo ring. The parallel test
   (`ptest_0845`) gains two guards for the band-less rank np=4 creates: the
   empty-DMLabel getStratumIS segfault (#291) and the zero-size max()
   collective desync.

3. **Explicit GAMG opt-out for the custom-P pickup** (`custom_mg.py`,
   `rotated_bc.py`). `solver.preconditioner = "gamg"` (or a user-latched
   pc_type) on an adapt child was silently clobbered back to the custom-P PCMG
   at solve time — both arms of test_0842's fmg-vs-gamg comparison ran
   pc_type=mg. `build_transfers` now stands down for an explicit choice,
   returning `(None, None)` (never bare `None`), and both call sites guard the
   unpack so "no hierarchy" degrades to the default preconditioner rather than
   a TypeError mid-solve.

4. **test_0842 hardened**. Both MeshVariables are created before any solve —
   creating one after a solve rebuilds mesh.dm under the custom-MG links
   (issue #492), which is what segfaulted Linux CI downstream. Each arm asserts
   the PC type it actually ran, so the comparison can never silently become
   FMG-vs-itself again. Tolerances: ksp_rtol 1e-9 against a 1e-7 nodal bound,
   chosen to catch the once-shipped true-residual stall (the geometric bundle
   now owns its fgmres outer pairing, #514/#515).

5. **`PYTHONFAULTHANDLER=1` in `scripts/test.sh`** — a CI crash prints a Python
   stack instead of a bare "Segmentation fault".

## Deliberately not included

- Development's own overlapping fixes are kept as landed: the 0753 skip-based
  tripwire, the 0844 self-built vertex blanket, the 0847 deterministic plotter
  close, and #511's reconciled reconnect/place-surface work (untouched here).

## Verification (post-rebase onto 606583eb, worktree env rebuilt)

- `test_0842` serial (2 passed) and np=2 (2 passed); explicit-gamg probe
  confirms each arm on its own PC (mg its=3, err 7.5e-10 vs gamg its=23,
  err 2.0e-9), no crash.
- `ptest_0845` at np=4: 3 passed on all ranks (the partition-independence case).
- Corner-band probe at np=4: relax with a one-rank corner surface completes on
  all ranks; `offset=5.0` raises ValueError on 4/4 ranks past a final barrier.
- `pytest tests/test_08*py -q` one process: 551 passed, 11 skipped, 11 xfailed,
  0 failed.
- Full gate `pytest tests -m "level_1 and tier_a"` (ignoring test_0050):
  580 passed, 17 skipped, 1 xfailed, 0 failed.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -46,6 +46,8 @@ if [ $PARALLEL_ONLY -eq 1 ] && [ $PARALLEL_RANKS -eq 0 ]; then
 fi
 
 export UW_NO_USAGE_METRICS=0
+# A hard crash must print a Python stack, not just "Segmentation fault".
+export PYTHONFAULTHANDLER=1
 PYTEST="pytest --config-file=tests/pytest.ini"
 
 # Run serial tests (unless --parallel-only specified)
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -6829,9 +6829,15 @@ class Mesh(Stateful, uw_object):
                     grown.update(vv)
             pinned |= grown
 
-        if not pinned:
-            # An empty DMLabel is not merely useless: querying its strata is a
-            # hard crash, not an exception, so refuse rather than hand one back.
+        # COLLECTIVE emptiness test. A rank whose subdomain the surface never
+        # enters legitimately has an empty local band — only a GLOBALLY empty
+        # band is a user error. The previous rank-local raise here deadlocked
+        # np=4 (measured 2026-08-06, review of PR #488: a corner-confined
+        # surface left three ranks raising while the fourth entered the
+        # collective mover and hung to the 300 s timeout).
+        n_global = uw.mpi.comm.allreduce(len(pinned))
+        if n_global == 0:
+            # Raised on EVERY rank, after the collective reduction.
             raise ValueError(
                 f"no cell is cut by distance == {offset} on surface "
                 f"{getattr(surface, 'name', surface)!r}, so there is no band to "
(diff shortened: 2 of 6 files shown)


lmoresi:
CI note for the record: the test job's 4 failures (`test_0851_fault_network_3d` ×3, `test_0851_std_reduction_method` error 79) are byte-identical to development's own failures at the base commit (development has been red since the 0851 tests landed — see the run history at c1cddb9b/6067bac7/d474a15e). This branch's own content is fully green locally: exact CI batch 551/0, full tier-a gate 580/0, np4 probes clean. Merging with admin on that provenance; the 0851 environment-sensitivity fix is going up separately.

Issue #486, opened 2026-07-30
multiplier_schur_pc: verify or remove — it is read, but no observable responds to it


lmoresi:
`Stokes_Constrained.multiplier_schur_pc` is a documented public property. It **is** read — at one site it swaps the multiplier block's Pmat term to the 1/μ Schur mass:

```python
hh_pc = self._pp_G0 if self._multiplier_schur_pc else self._hh_G0[k]
PetscDSSetJacobianPreconditioner(ds.ds, fid, fid, ext.fns_jacobian[i_jac[hh_pc]], NULL, NULL, NULL)
```

So it is not dead code. But we could not construct a case where it changes anything.

## Measured

Annulus, `Stokes_Constrained` with `add_constraint_bc` on both arcs (2 multipliers present, so the loop above is reached), `snes_type=ksponly`:

| configuration | `flag=False` → `flag=True` |
|---|---|
| constant viscosity | `\|u\|`, `\|p\|`, outer KSP its all identical |
| η contrast 1e4 | identical — including the solve **failing** identically both ways (DIVERGED_LINEAR_SOLVE, 100 KSP its) |
| η contrast 1e6 | identical, 4 KSP its both |
| assembled `\|Pmat\|` (η contrast 1e3) | identical to 12 digits |

The flag reads back correctly (`True`/`False` round-trip), so it is recorded.

Also notable in the same runs: `|Pmat|` equalled `|J|` **exactly** in both settings, despite `Pmat` being a distinct matrix.

## Why this is filed as verify-or-remove rather than as a bug

The negative could be honest — two good Schur preconditioners can both converge in four iterations on a problem this easy, and we may simply not have found the discriminating case. The comment at the call site says the point is to reuse pressure's already-compiled 1/μ term "so there is NO extra JIT term", which suggests the motivation was compile cost rather than convergence.

But per the "identical cells are a common defect, never 'this variable does not matter'" rule that found #477, four identical rows across two decades of viscosity contrast is not evidence the flag works.

So: either produce the case where it earns its keep — and add it as a test, so the next audit does not re-open this — or remove the property. A public flag nobody can demonstrate an effect for is the same class as #477 and `solver.strategy`: accepted, recorded, and invisible if it is doing nothing.

[...]

Issue #484, opened 2026-07-30
A warning is not observability: 10 of 12 preconditioner fallbacks leave no readable state


lmoresi:
Five defects of the same shape have been found independently over the last few weeks — #276/#478, #467, #468, #477, #425. Every one of them **degrades to something that succeeds**: GAMG converges, richardson converges, the solve returns a correct answer. So none presented as a failure. They presented months later as "this is a bit slow", or as a parameter sweep returning identical rows, which reads as *"this variable does not matter"* rather than *"this variable was never set"*. That cost a twelve-cell sweep on #477 and a wrong premise in #471.

This issue is about the property that makes them all invisible, rather than any one of them.

## Inventory of the preconditioner-configuration fallbacks

| site | substitutes | observable? |
|---|---|---|
| `_apply_preconditioner_options` single-field gate | geometric MG → GAMG | warns **only** on explicit `fmg` |
| `_apply_preconditioner_options` `n_levels <= 1` | geometric MG → GAMG | warns only on explicit `fmg` |
| `_apply_preconditioner_options` override latch | user's bundle → managed bundle | **nothing** (fixed in #471) |
| `auto_inject_custom_mg` barycentric→RBF retry | sparse transfer → **dense** | warns |
| `auto_inject_custom_mg` build failure | geometric MG → default PC | warns |
| `auto_inject_custom_mg` dimensional guard | skips the auto-pickup | warns |
| `_assert_finest_spans_operator` swallow | **skips a correctness guard** | **nothing** |
| `CustomMGHierarchy.build` `setUp` swallow | proceeds on a possibly-stale DM section | **nothing** |
| `_install_velocity_block_transfers` PETSc-error path | alternative assembly route | nothing (same outcome) |
| `_enforce_galerkin_for_geometric_mg` | forces `pc_mg_galerkin=both` | warns |
| `rotated_bc` no hierarchy | geometric MG → GAMG | **`velocity_pc`** |
| `rotated_bc` no pressure mass | 1/μ-mass Schur → `selfp` + jacobi | **`schur_pre`** |

Only the last two leave readable state. Both were added in #465 — and they are the **only** reason #467 was findable at all.

## Three things the current state cannot do

[...]

Issue #483, opened 2026-07-30
Stokes and Stokes_Constrained disagree about which solver options are reachable, in opposite directions


lmoresi:
Measured off the live PETSc objects after two solves, `tolerance = 1e-6`, on `development` @ `04379583`:

| option | `Stokes` | `Stokes_Constrained` |
|---|---|---|
| `ksp_rtol` | **reachable** | **NOT reachable** (→ `tolerance × 0.1`) |
| `fieldsplit_velocity_ksp_rtol` | **NOT reachable** (→ `tolerance × 0.033`) | **reachable** |
| `fieldsplit_pressure_ksp_rtol` | **NOT reachable** (→ `tolerance × 0.1`) | **reachable** |
| `fieldsplit_{velocity,pressure}_ksp_max_it` | reachable | reachable |
| `fieldsplit_{velocity,pressure}_ksp_type` | reachable | reachable |
| `snes_max_it` | NOT reachable (→ 50) | NOT reachable (→ 50) |
| `snes_rtol` | NOT reachable (→ `tolerance`) | NOT reachable (→ `tolerance`) |

Both are saddle-point solvers. They disagree about the outer KSP tolerance **and** about the inner ones, in opposite directions, and nothing documents the difference. `max_it` and `ksp_type` on the same prefixes are the control: same dict, same solve, they go through — so this is specific to the keys a `tolerance` setter writes.

Scalar (`Poisson`) and vector (`Vector_Projection`) families are **clean** — every option tested is reachable. The whole class is localised to the saddle-point path.

Repro: `~/+Simulations/solver_reachability_audit/reachability_matrix.py -uw_family stokes|constrained|scalar|vector`. It reads the **live** objects rather than the options database (a clobbered key leaves no trace there), solves **twice** (solve-time clobbers only bite on the second), and runs a **defaults arm** as the control so a "reachable" verdict cannot be an accident of the override matching the default.

## Two findings, one surface

**1. The disagreement itself.** Whatever the intended ownership rule is, it should be the same for both solvers.

[...]

Issue #477, opened 2026-07-29
fieldsplit_velocity_ksp_rtol and fieldsplit_pressure_ksp_rtol are silently un-settable — solve() clobbers them every call


lmoresi:
## What

`SNES_Stokes_SaddlePt.solve()` executes

```python
self.tolerance = tolerance          # pyx:8811 (and 8799 on the Picard branch)
self.snes.atol = self.atol
self.petsc_options.setValue("snes_max_it", snes_max_it)
self.snes.setFromOptions()
```

on **every** solve, and the `tolerance` setter writes (pyx:6300-6301):

```python
self.petsc_options["fieldsplit_pressure_ksp_rtol"] = self._tolerance * 0.1
self.petsc_options["fieldsplit_velocity_ksp_rtol"] = self._tolerance * 0.033
```

So a user's value for either option is overwritten *between* being set and being read by
`setFromOptions()`. **Both options are documented** (in the `tolerance` docstring itself)
and both are unreachable.

## Reproducer

```python
s.tolerance = 1e-8
s.petsc_options["fieldsplit_velocity_ksp_rtol"]  = 1.0e-4    # ignored
s.petsc_options["fieldsplit_velocity_ksp_max_it"] = 7        # honoured
s.solve()
pc = s.snes.getKSP().getPC(); pc.setUp()
k = pc.getFieldSplitSubKSP()[0]
print(k.getTolerances())     # rtol 3.300e-10 (= 1e-8 * 0.033), max_it 7
```

Measured, res-8 box, both orderings:

```
set BEFORE first solve -> vel_rtol=3.300e-10 (asked 1.000e-04)   vel_max_it=7 (asked 7)
set AFTER  first solve -> vel_rtol=3.300e-10 (asked 1.000e-04)   vel_max_it=7 (asked 7)
```

`max_it` is the control: same prefix (`Solver_N_fieldsplit_velocity_`), same dict, same
solve — it goes through, because the `tolerance` setter does not touch it. Only the two
keys the setter writes are affected. Re-asserting the option after the solve does not help
either; the next solve clobbers it again.

## Why it matters more than a missing knob

**The failure is invisible.** A sweep over the velocity inner tolerance returns rows that
are identical to five significant figures, which reads as "this variable does not matter"
rather than "this variable was never set". It cost a 12-cell configuration sweep here
before the identical cells were followed up:

```
outer  vel_rtol  schur_mu | ok  nl   vel      wall     |F|/|F0|
[...]

PR #534, opened 2026-08-12
Solver options: fallback observability, tolerance ownership, the single-field FMG unlock, and multiplier_schur_pc verified (#478 #483 #484 #486)

### LINKED ISSUE #478: Geometric multigrid is unreachable for every single-field solver, even when explicitly requested (the #276 lockout)

lmoresi:
Solver options: make fallbacks readable, tolerances honest, explicit geometric MG reachable, and multiplier_schur_pc truthful

Fixes #478 #483 #484 #486

Four related reachability/observability defects in the solver-options layer, landed as one commit per issue (recorder first — everything else writes into it).

**#484 — every preconditioner fallback leaves a readable record.** Ten of the twelve places a solver silently degrades its preconditioner (single-field FMG gate, missing hierarchy, transfer-build failures, guard skips, forced Galerkin, the rotated-path substitutions) left no queryable state — a warning at best, nothing at worst — so neither a user nor a test could ask "did I get what I asked for?". Every site now writes through one recorder into the public read-only `solver.pc_fallbacks` (keyed by site, fixed reason vocabulary: `unavailable` / `declined` / `build_failed` / `check_skipped` / `forced`), cleared whenever the preconditioner options re-resolve. What deliberately did NOT change: any option value, route, or warning — existing warnings stay, the silent "auto" declines gain a record but no new noise, and `rotated_bc`'s own ctx report keys stay authoritative for that path's tests. A clean solve records nothing (that empty record is the suite's global negative control).

**#483 — user-set `snes_rtol`/`ksp_atol` are honoured on the saddle-point solvers.** Both keys were re-pushed by every solve, silently discarding an explicit user value — documented as settable, actually owned. The re-push now runs through the recorded-ownership latch that made `snes_max_it` reachable (ruling D18), generalised to a dict: the framework keeps asserting the tolerance-derived values until the user sets a key, after which their value wins across solves. Alongside, the two hand-rolled tolerance-derivation code paths (Stokes vs Stokes_Constrained) collapse into one mechanism over per-class tables (`_TOLERANCE_DERIVED_KEYS`), with the real design difference — Constrained derives the outer `ksp_rtol` + Eisenstat-Walker pins instead of the inner fieldsplit margins, because EW pinning owns its outer accuracy — now documented in both `tolerance` docstrings as an ownership table. What did NOT change: a user who never touches the keys sees byte-identical behaviour (proved by no-override arms reading the live PETSc objects after two solves), the table *contents* are untouched, and the EW flags are still never re-asserted.

**#478 — explicit geometric MG on single-field solvers is unlocked (opt-in only).** `preconditioner="fmg"` on a scalar/vector solver was declined to GAMG with a warning, because native FMG needs `DMCreateInjection`, which PETSc cannot reliably build for a single-field discretisation on a refined DMPlex (err62, #276) — locking geometric MG out even though the injection-free custom-P route already existed. The explicit request is now honoured: `custom_mg.build_transfers` grows a third hierarchy source (requested-native: the mesh's own `dm_hierarchy` tail behind `_DMLevelView` adapters), resolved after solver-set and mesh-owned so it stays the single "which hierarchy" owner and composes with #530's explicit-gamg opt-out. The options DB deliberately keeps GAMG as the degrade base until the live PC is configured at first solve; build failure degrades through the recorded barycentric→RBF→default ladder. Contract line stated in both docstrings: `preconditioner="fmg"` is a preference (degrades readably), `set_custom_fmg` is a demand (raises). What did NOT change: `"auto"` still declines geometric MG on single-field solvers — flipping that default needs its own np2/np4 + adaptivity campaign per the issue — it keeps GAMG and records the decline as the migration probe. Verified on the exact #276 err62 geometries (annulus refinement=2 qdegree=3; flat high-degree box), serial and np=2. Found en route: a gmsh-imported base hierarchy level carries section-only coordinates (PetscContainer, no PetscFE) and `DMCreateInterpolation` from such a source silently returns a zero matrix — `_DMLevelView` works on a clone and installs a P1 Lagrange coordinate FE when missing.

**#486 — `multiplier_schur_pc` verified, kept, and instrumented.** The issue measured four identical convergence rows and asked verify-or-remove. Tracing PETSc's fieldsplit.c settles it: the flag swaps only the Pmat (h,h) block, and under Stokes_Constrained's defaults (`selfp` + `diag_use_amat`) selfp assembles the Schur preconditioner from Amat sub-blocks — the swapped block is provably never read (exactly what was measured). It IS read under `schur_precondition="a11"` and under a monolithic direct factorisation of the Pmat, so the property stays. An explicit opt-in that cannot reach the PC now records + warns (the one new warning in this PR — an opt-in silently doing nothing is the #477 class). The regression test is matrix-level (assembled Schur pre differs flag-on/off under a11; identical under selfp, pinning the PETSc semantics), immune to "both converge in 4 iterations". Docstring drift corrected: Constrained's `saddle_preconditioner` claimed selfp used the automatic 1/viscosity mass — it does not; that holds only under a11 — with a TODO(BUG) on whether selfp should see Pmat blocks at all (design question, not changed). What did NOT change: no default, no behaviour of any solve.

Tests: `test_1022` (fallback observability, every probe fired + silent-on-healthy), `test_1023` (tolerance ownership, live-object reads, both classes), `test_1020` rewritten from lockout test to unlock test + `ptest_1020` (np=2 route parity), `test_1024` (multiplier Schur regimes). Stakeholders green: 1013, 1014 (x2), 1015, 1016, 1017, 1018 (rotated), 1021, 1061, 1062, 1065, 0820. Full `level_1 and tier_a` gate: 580 passed, 0 failed.

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -762,6 +762,13 @@ class SolverBaseClass(uw_object):
                 f"preconditioner must be 'auto', 'fmg', or 'gamg' (got {value!r})"
             )
         self._preconditioner = choice
+        # A hierarchy cached by an earlier RESOLUTION (the auto/"fmg" install)
+        # must not outlive a new explicit choice: auto_inject_custom_mg
+        # re-installs solver._custom_mg unconditionally, which would leave a
+        # later preconditioner="gamg" unreachable with a clean pc_fallbacks
+        # record. A user registration (set_custom_fmg) is a demand and is kept.
+        if isinstance(self._custom_mg, dict) and self._custom_mg.get("auto_cached"):
+            self._custom_mg = None
         # Force a full rebuild so the new option bundle is pushed to PETSc.
         self.is_setup = False
 
@@ -8422,7 +8429,10 @@ class SNES_Stokes_SaddlePt(SolverBaseClass):
                           else "")
             _pc_type = (_opts.getString("pc_type")
                         if _opts.hasName("pc_type") else "")
-            _diag_amat = _opts.hasName("pc_fieldsplit_diag_use_amat")
+            # Read the VALUE, not the key's presence: diag_use_amat set to
+            # "false" means the Pmat block IS read and the opt-in is live
+            # (measured: Schur pre differs by rel-Frobenius 0.30 flag-on/off).
+            _diag_amat = _opts.getBool("pc_fieldsplit_diag_use_amat", False)
             if (_schur_pre != "a11" and _diag_amat
                     and _pc_type not in ("lu", "cholesky")):
                 self._record_pc_fallback(
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -7019,10 +7019,20 @@ class SNES_Stokes_SaddlePt(SolverBaseClass):
         the block solver's default ``newtonls`` defect-corrects a linear system in
         many steps when the Schur approximation is stiff.
 
-        Default ``False`` (opt-in). On the fieldsplit/iterative path it is
-        bit-identical on uniform ``mu`` and cracks the moderate-contrast wall;
-        but a monolithic ``lu`` solve factorizes the Pmat (``pc_use_amat`` is a
-        no-op there), so this term is not inert for direct solves — hence opt-in.
+        Default ``False`` (opt-in). **Reachability** (#486, from the PETSc
+        fieldsplit source): the flag swaps only the *Pmat* (h,h) block, so it
+        is live in exactly two regimes —
+
+        - ``pc_fieldsplit_schur_precondition = "a11"``: the Schur
+          preconditioner is the grouped ``[p,h]`` Pmat block, which carries
+          the swap;
+        - a monolithic direct factorisation (``pc_type = lu``/``cholesky``)
+          of the Pmat.
+
+        Under ``Stokes_Constrained``'s own defaults (``selfp`` +
+        ``diag_use_amat``) the Pmat (h,h) block is never read by the Schur
+        preconditioner and the flag is INERT — setting it there records a
+        ``multiplier_schur_pc`` entry in :attr:`pc_fallbacks` and warns.
         """
         return self._multiplier_schur_pc
 
@@ -8393,6 +8403,52 @@ class SNES_Stokes_SaddlePt(SolverBaseClass):
         # Lagrange-multiplier rows (block-constrained Stokes). Guarded: no-op
         # for ordinary Stokes. Register the interior screening residual and the
         # diagonal mass Jacobian/preconditioner on each multiplier's field.
+        #
+        # multiplier_schur_pc reachability (#486, resolved by tracing PETSc
+        # fieldsplit.c): the flag swaps only the Pmat (h,h) block below. With
+        # schur_precondition=selfp the Schur preconditioner Sp is assembled by
+        # MatSchurComplementGetPmat from sub-matrices split out of the AMAT
+        # whenever pc_fieldsplit_diag_use_amat is set (jac->mat[1], not
+        # jac->pmat[1]) — so under this class's defaults (selfp +
+        # diag_use_amat) the swapped block is provably never read. It IS read
+        # under schur_precondition=a11 (Sp = the grouped [p,h] Pmat block) and
+        # under a monolithic direct factorisation of the Pmat. An explicit
+        # opt-in silently doing nothing is exactly the #477 class -> record
+        # AND warn (unlike the auto declines, this one earns the warning).
+        if self._multipliers and self._multiplier_schur_pc:
+            _opts = self.petsc_options
+            _schur_pre = (_opts.getString("pc_fieldsplit_schur_precondition")
+                          if _opts.hasName("pc_fieldsplit_schur_precondition")
+                          else "")
+            _pc_type = (_opts.getString("pc_type")
+                        if _opts.hasName("pc_type") else "")
+            _diag_amat = _opts.hasName("pc_fieldsplit_diag_use_amat")
+            if (_schur_pre != "a11" and _diag_amat
+                    and _pc_type not in ("lu", "cholesky")):
+                self._record_pc_fallback(
+                    "multiplier_schur_pc",
+                    requested="1/mu multiplier Schur mass (Pmat h,h block)",
+                    installed="unread — selfp builds Sp from the Amat A11 block",
+                    reason="declined",
+                    detail=f"pc_fieldsplit_schur_precondition="
+                           f"'{_schur_pre or 'selfp'}' with diag_use_amat: the "
+                           f"Pmat (h,h) block never reaches the Schur "
+                           f"preconditioner; set schur_precondition='a11' (or "
+                           f"factorise the Pmat directly) to make the opt-in "
+                           f"live")
+                if uw.mpi.rank == 0:
+                    import warnings
+                    warnings.warn(
+                        f"[{self.name}] multiplier_schur_pc=True has no effect "
+                        f"under pc_fieldsplit_schur_precondition="
+                        f"'{_schur_pre or 'selfp'}' with diag_use_amat: the "
+                        f"1/mu multiplier Schur mass is written into the Pmat "
+                        f"(h,h) block, which selfp never reads (Sp is built "
+                        f"from the Amat). Use "
+                        f"petsc_options['pc_fieldsplit_schur_precondition'] = "
+                        f"'a11' to make it live. See solver.pc_fallbacks.",
+                        stacklevel=2,
+                    )
         for k, mvar in enumerate(self._multipliers):
             fid = mvar._solver_field_id
             # Operator (Amat) (lambda,lambda) block is ALWAYS the true screening eps so
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -210,6 +210,13 @@ class SolverBaseClass(uw_object):
         # helper below is a no-op.
         self._preconditioner = "auto"
         self._pc_option_prefix = None
+        # An explicit `preconditioner="fmg"` on a single-field solver cannot
+        # take the native route (#276), so it is honoured via custom-P
+        # transfers over the mesh's own dm_hierarchy instead (#478). This flag
+        # carries the request from _apply_preconditioner_options (build time)
+        # to custom_mg.build_transfers (first solve); it is re-derived on
+        # every resolution, same staleness rule as _pc_resolved.
+        self._pc_single_field_geo_requested = False
         # The pc_type value this helper last managed. Subclasses that opt in set
         # their __init__ default ("gamg"); used in "auto" mode to tell an
         # untouched framework default (eligible for FMG upgrade) apart from an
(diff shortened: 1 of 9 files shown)


lmoresi:
# Adversarial review — PR #534 (solver-options bundle: #478 #483 #484 #486)

Reviewed independently at `44e483b9` (4 commits over merge-base `a4a8ba17`), built
and measured in a clean worktree (`feature/r534-review`, amr-dev env, worktree
site-packages verified). We diff-read every hunk in the pyx, custom_mg,
solvers.py and rotated_bc, re-ran the PR's own suites, and ran our own probes:
a merge-base vs PR-head default-identity comparison, latch-lifecycle attacks,
an np=2 battery the PR did not cover, and negative controls by reverting hunks
in the installed copy.

## MERGE-BLOCKERS

**B1. After the fmg unlock installs, a later explicit `preconditioner = "gamg"`
is silently ignored — and the new `pc_fallbacks` record reports a clean
resolution.** Measured: Poisson, refinement=1, `preconditioner="fmg"`, solve
(live PC `mg`, `_custom_mg` cached) → `preconditioner="gamg"`, solve → live PC
**still `mg`**, `pc_fallbacks == {}`. Root cause: the requested-native install
stores itself in `solver._custom_mg` (`auto_inject_custom_mg` tail), and
auto_inject's first branch re-injects any cached `_custom_mg` unconditionally,
bypassing both `_apply_preconditioner_options` (which resolved gamg and cleared
the record) and the explicit-gamg guard inside `build_transfers`. The
`preconditioner` setter forces a rebuild (`is_setup = False`, pyx ~766) but
never drops the cached hierarchy. Three defects in one: the user's explicit
option is unreachable (the #477 class this PR exists to fight), the PR body's
"composes with #530's explicit-gamg opt-out" holds only *before the first
solve*, and the recorder's documented contract — "Empty means the resolved
preconditioner is exactly what was requested" — is violated by the PR's own new
route. The adapt-child arm of this stickiness pre-exists (#530's opt-out had
the same before-first-solve limit), but the requested-native route is new, is
driven by an ordinary property rather than a `set_custom_fmg` demand, and the
false-clean record is new. Fix is small: on `preconditioner` assignment (or in
the explicit-gamg branch of `_apply_preconditioner_options`), clear a cached
`_custom_mg` whose mode is `"hierarchy"` (never a user's `set_custom_fmg`
registration), or make auto_inject's early branch respect the same
explicit-gamg/user-override guard for hierarchy-mode caches. Add the
switch-after-solve arm to test_1020.

**B2. The #486 inertness probe reads the *presence* of
`pc_fieldsplit_diag_use_amat`, not its value — and warns falsely in a regime
where the flag is live.** pyx `_setup_solver`:
`_diag_amat = _opts.hasName("pc_fieldsplit_diag_use_amat")`. Measured: selfp +
`petsc_options["pc_fieldsplit_diag_use_amat"] = False` (stored `"false"`,
`hasName` still True) + `multiplier_schur_pc = True` → the assembled Schur pre
**differs flag-on/off by rel. Frobenius 0.30** (PETSc splits the Schur
sub-matrices from the Pmat when diag_use_amat is off, so the swapped (h,h)
block IS read) — yet the solver warns "has no effect" and records
`reason="declined"`. The one new warning in this PR tells a user whose opt-in
is working that it is doing nothing, in a PR whose stated purpose is truthful
observability. One-line fix: read the bool value (`getBool`-equivalent /
string compare), plus a probe arm in test_1024 with diag_use_amat=false.

Both fixes are localized, neither touches a default path; everything else we
attacked held up. With B1+B2 addressed we would merge.

## Verified claims (measured)

- **No default behaviour change** — the strongest claim, and it holds. We ran
  a 4-configuration probe (Poisson refinement=0 and =2 on `auto`; Stokes with
  3-decade viscosity contrast, tolerance=1e-6, two solves; default
  Stokes_Constrained with two constraint arcs, two solves) capturing live PC
  type, SNES/KSP iteration counts, **full KSP residual histories to 16
  digits**, live tolerances, and solution norms — on the merge-base build and
  the PR-head build of the same worktree. Byte-identical throughout, with one
  DB-presence exception (M3 below). The stakeholder suites are green
  (1013/1014×2/1015×2/1016×2/1017/1018/1021/1061/1062/1065×2/0820: 99 passed
  across our batches) and the full `level_1 and tier_a` gate passes:
  **580 passed, 0 failed**.
- **#483 latch** — user-set `snes_rtol`/`ksp_atol` honoured across two solves,
  read off live objects (test_1023, 6/6). Our lifecycle attacks: latch survives
  a forced re-setup (`is_setup=False` → user 4e-3 still live); setting
  `tolerance` *after* a latched user value re-derives correctly (the setter's
  DB write reads as a fresh user move — no #418-style stale-latch on any
  documented path); order rules both directions. Fail-before confirmed:
  test_1023's override test fails on the merge-base build. The ownership table
  is in both `tolerance` docstrings, including the EW-pins-vs-inner-margins
  design difference.
- **#478 unlock** — the exact #276 err62 geometries run custom-P geometric MG
  off the live PC (annulus r=2 q=3, flat q=3 box; `pc.getMGLevels()` == full
  hierarchy depth). ptest_1020 passes at np=2, and our **np=2 annulus** case
  (gmsh base level, i.e. the coordinate-FE repair exercised in parallel —
  a case the PR did not cover) also passes: pc `mg`, 3 levels, both ranks, no
  degrade record. The zero-interpolation collateral defect is real: we rebuilt
  the probe — a raw clone of the gmsh-imported base level (coordinate field is
  a plain PetscContainer, no PetscFE) hands back **all-zero coordinates
  silently** (`max|x| = 0.0`), while `_DMLevelView`'s P1 install recovers the
  true annulus geometry (`max|x| = 1.0`, `min r = 0.5`). Negative control:
  disabling the FE install in the installed copy fails the unlock test.
  No-hierarchy explicit fmg degrades gracefully (record `unavailable` + warn,
  live gamg, converged). The rotated composition question is structurally
  unreachable: `add_rotated_freeslip_bc` is defined on `SNES_Stokes_SaddlePt`
  only (pyx ~6066), so no single-field solver can reach the rotated path;
  the rotated Stokes suite (test_1018) is green.
- **#484 recorder** — record-on-all-ranks / warn-rank-gated is real: forced
  no-hierarchy fallback at np=2, `allgather`ed records identical on both
  ranks, warn counts `[1, 0]`. Staleness: flipping the preconditioner
  re-resolves and clears (test_1022 reset test) — but see B1 for the one route
  that plants state outside the record's reach. The rotated sites mirror only
  on actual substitution and the rotated ctx is rebuilt per solve call, so no
  cross-solve staleness there (code-read). Negative control: muting the
  transfer-builder recorder call in the installed copy fails the test that
  asserts it. Fail-before: test_1022's auto-decline test fails on merge-base
  (no `pc_fallbacks` attribute use / no record).
- **#486** — the a11 matrix oracle is genuine: it extracts the assembled Schur
  pre and asserts a flag-on/off difference (rel. Frobenius > 1e-6); the selfp
  inertness control asserts **exact** equality (`== 0.0`), pinning the
  fieldsplit.c semantics. Note the a11-liveness test also *passes on the
  merge-base build* — correct and expected (#486 is verify+instrument, the
  flag was always live under a11), so that test is a regression pin, not a fix
  validation; the fail-before tests for this commit are the record/warn arms.
  Default Constrained solves (flag False) produce zero multiplier warnings
  (measured in the identity probe and test_1024's negative control).

## Minor findings (non-blocking)

- **M1. Deleted user override resurrects.** Latch a user `snes_rtol=5e-3`,
  solve, `delValue("snes_rtol")`, solve (default 1e-6 pushed — correct), solve
  again → **5e-3 is back** (the re-push made `hasName` true, `current ==
  pushed`, and the stale `_owned_option_user` entry returns). Inherited from
  the D18 snes_max_it latch, but this PR widens it to three keys. One-line
  fix: clear the user latch when the key is absent.
- **M2. `type(default)(getString(key))` narrows what the latch accepts.** An
  int-typed owned key set to a non-canonical string (e.g. `"7.0"`) raises in
  `int()` and silently returns the framework default, where the old
  `getInt` path parsed it. Edge case; worth a `float()`-then-cast for int
  defaults.
- **M3. `Stokes_Constrained.__init__` now writes `ksp_rtol = 1e-5` into the
  options DB** (via `_derive_tolerance_margins`; previously absent until the
  user set `tolerance`). Live behaviour is measurably identical — the value
  equals PETSc's own ksp_rtol default, EW re-picks per Newton step, and no
  in-tree code branches on the key's presence (grepped) — but the PR body's
  "byte-identical" claim is about live objects, not the DB. The code comment
  discloses it; fine as is.
- **M4. The first solve latches the tolerance setter's own `snes_rtol` DB
  write as a "user" value.** Harmless today: every `_tolerance` write in-tree
  goes through a setter that also writes the DB (grepped all 9 sites), so a
  later `tolerance=` always re-latches. But any future internal
  `self._tolerance = x` direct write will silently not reach `snes_rtol`
  (measured: D-probe, stayed 1e-6 after `_tolerance=1e-7`) — under the old
  unconditional re-push it would have. Deserves one comment line at
  `_resolve_owned_option`.

## Run log

- New tests: test_1022 (11) + test_1023 (6) + test_1024 (5) + test_1020 (6)
  + ptest_1020 (np=2) — all pass on PR head.
- Fail-before (merge-base build): test_1023 override FAILS, test_1020 unlock
  FAILS, test_1022 auto-decline FAILS, test_1024 a11-oracle passes (expected —
  regression pin).
- Negative controls (installed-copy hunk reverts): FE install disabled →
  unlock test FAILS; transfer-builder recorder muted → its test FAILS. Both
  restored and rebuilt before the gate.
- Full gate on PR head: `pytest tests -m "level_1 and tier_a" -q`
  (--ignore test_0050): **580 passed, 17 skipped, 1 xfailed, 0 failed**
  (7:07) — matches the PR's claimed 580/0.

lmoresi:
Response commit 45acc08b: both blockers fixed — (1) the fmg-install hierarchy cache is now marked `auto_cached` and dropped by the `preconditioner` setter (a `set_custom_fmg` registration is kept), with a regression test whose negative control was verified live (fails with the marker disabled, 3.1 s); (2) the #486 inertness probe reads the `diag_use_amat` bool value, not the key's presence, with a flag=false silence test. All five bundle test files: 49 passed. The review's minors (the D18 latch-resurrection wart, int-coercion strictness) are inherited pre-existing behaviour, left with the review as the record.

PR #545, opened 2026-08-13
Place a tip vertex from the face that holds it, not from the mesh at large

### LINKED ISSUE #542: pull_vertex_onto can take the vertex from a neighbouring cell, so the tip cell has no on-line corner
## What happens

`cut_along_lines` refuses a fault whose tip is correctly placed:

```
ValueError: 1 triangle(s) are entered but not left, which means a line ends
inside the mesh. A line must cross from boundary to boundary; a terminating
tip needs a vertex placed at the tip and is not supported.
```

even though `add_fault` has already called `pull_vertex_onto` on every control point of the trace, and every control point does have a vertex exactly on it.

## Why

`pull_vertex_onto` moves the **globally nearest** vertex onto each target. That vertex need not be a corner of the cell that *contains* the target. When it is not, the containing cell is entered by the line and terminates there with no on-line corner, which is exactly the state `cut_along_lines` refuses.

Measured on the listric pair at DX=0.30 (`~/+Simulations/listric_extension/selfselect.py`), fault F2, trace start `(0.65, 0.88)` inside an embedded ribbon zone of width 0.02:

| vertex | position | distance to the start point | corner of the containing cell? |
|---|---|---|---|
| 2213 | (0.65361, 0.87071) | **9.963e-03** — chosen | no |
| 1953 | (0.64068, 0.87638) | 1.000e-02 | yes |
| 1889 | (0.65932, 0.88362) | 1.000e-02 | yes |

The margin is 0.4%. Both corners of the containing cell sit at *exactly* `width/2` from the trace, because they are the ribbon's own edge vertices; the winning vertex is an interior one that happens to fall a hair closer. All 22 control points ended up with a vertex exactly on them, and the cut still refused.

## Why it looks intermittent

- On a plain unstructured box the globally nearest vertex is usually a corner of the containing cell. A tip swept across a cell at 7 positions succeeded every time.
- Inside an embedded zone the vertices sit in rows at exactly `±width/2` from the trace, which manufactures near-ties. The same sweep fails at DX=0.30 and passes at DX=0.50 and DX=0.70 — luck, not geometry.

This is what limited the hybrid self-selection sweep to a subset of separations.

## The fix we think is right

Choose the nearest vertex **of the cell containing the target**, falling back to the globally nearest when the target lies outside the mesh. This keeps the existing principle — move the mesh, not the tip, so the tip stays exactly where it was asked for — and makes the choice depend on the containment relation that `cut_along_lines` actually tests, rather than on a distance comparison that can be decided by 0.4%.

Note that snapping the *tip* to the nearest vertex, the other obvious fix, is both rejected by `pull_vertex_onto`'s own docstring (measured: tip error 0.0000 against 0.0306, worst angle 8.79 deg against 5.29) and would not address this: the tip does not lack a vertex, the vertex was taken from the wrong cell.

## Reproducer

`~/+Simulations/listric_extension/selfselect.py 0.30` — fails; `0.50` and `0.70` pass.

lmoresi:
Closes #542.

`pull_vertex_onto` moved the **nearest vertex in the whole mesh** onto each target. That vertex need not belong to the cell containing the target — it can be the apex of a neighbouring cell — and moving it leaves the containing cell with no corner on the line. `cut_along_lines` then refuses the tip with "entered but not left", for a tip that does have a vertex on it.

## The measured case

Inside a ribbon zone of width 0.02, a fault trace starting at `(0.65, 0.88)`:

| vertex | distance to the target | on the edge the target lies on? |
|---|---|---|
| 2337 | **9.963e-03** — chosen | no, it is a neighbouring cell's apex |
| 2014 | 1.000e-02 | yes |
| 2015 | 1.000e-02 | yes |

The target sits exactly on the edge 2014–2015, whose ends are the ribbon's own edge vertices at exactly the half-width. Regular vertex rows at `±w/2` make near-ties like this the norm inside a zone rather than bad luck, which is why it showed up there and not on a plain mesh — a tip swept across a cell of an ordinary box succeeded at 7 of 7 positions.

## The fix

Candidates are the vertices of the **smallest face whose closure contains the target**: the three corners of the cell holding it, the two ends of the edge it lies on, or the vertex it coincides with. The face is read off the barycentric coordinates of one containing cell.

Reading it from **one** cell is what makes it right. Taking the union of the corners of every containing cell is not enough: a point on an edge lies in two cells, and the union then offers each cell's opposite apex, which covers one cell and not the other. That was the first attempt here and it did not clear the measured case — the apex still won on distance.

In parallel the reduction leads with an outside-flag, so a rank holding no part of the containing cell cannot win with its own nearer vertex. All ranks flagging outside means the target lies off the mesh, and the fallback is the previous nearest-anywhere behaviour.

## Why it matters

This is what limited the hybrid fault self-selection sweep to a subset of separations. With the fix, `selfselect.py` at DX=0.30 gets past the cut; it previously failed there while passing at 0.50 and 0.70.

## Tests

`test_the_pulled_vertex_covers_every_cell_holding_the_target` asserts the invariant the cut actually needs — every cell whose closure contains the target ends up with a corner exactly on it — over interior edge midpoints.

The mesh is **flattened in y first**, and that is load-bearing rather than cosmetic: on a well-shaped mesh an edge midpoint's nearest vertex is one of that edge's own ends, the defect never appears, and the test would pass without testing anything. A negative control (`exposed > 0`) asserts the mesh really does present the case; the first version of this test, on an unflattened mesh, duly failed that control. Against the old rule the invariant fails at **38 of 71** midpoints tested.

`test_0844`, `test_0845`, `test_0846`, `test_0850_faults`, `test_0850_fault_network_toolkit`, `test_0853` pass (102 passed, 1 pre-existing skip); `ptest_0844` and `ptest_0845` pass at np=2 (20 passed).

### DIFF
--- a/src/underworld3/utilities/line_cut.py
+++ b/src/underworld3/utilities/line_cut.py
@@ -379,6 +379,53 @@ def _cell_vertices(dm):
                     dtype=np.int64).reshape(cE - cS, 3)
 
 
+def _minimal_face_vertices(X, cell_verts, point):
+    """Vertices of the smallest mesh face whose closure contains ``point``.
+
+    A point sits in the relative interior of exactly one face: a triangle, an
+    edge, or a vertex. This returns that face's vertices — the three corners
+    of the cell holding it, the two ends of the edge it lies on, or the single
+    vertex it coincides with. Empty when the point lies outside the mesh.
+
+    The face is read off the BARYCENTRIC coordinates of the containing cell: a
+    coordinate is zero exactly when the point lies on the opposite edge, so
+    keeping the corners with a nonzero coordinate drops the corners the point
+    is not "on". Doing it this way means one cell answers the question. Taking
+    the union of the corners of every containing cell does NOT: a point on an
+    edge lies in two cells, and the union then offers each cell's opposite
+    apex, which belongs to one of them and not the other. Moving such an apex
+    leaves the other cell without a corner on the point, which is the defect
+    this whole path exists to avoid (issue #542).
+
+    The coordinates are scale-free, so one relative tolerance serves cells of
+    any size — which embedded meshes need, their cell sizes differing by an
+    order of magnitude either side of a zone skin.
+    """
+    A = X[cell_verts[:, 0], :2]
+    B = X[cell_verts[:, 1], :2]
+    C = X[cell_verts[:, 2], :2]
+    v0, v1, v2 = C - A, B - A, np.asarray(point, dtype=float)[:2] - A
+    den = v0[:, 0] * v1[:, 1] - v1[:, 0] * v0[:, 1]
+    # A degenerate cell has no interior to be inside of; it cannot hold the
+    # point and must not divide.
+    alive = np.abs(den) > 0.0
+    den = np.where(alive, den, 1.0)
+    lc = (v2[:, 0] * v1[:, 1] - v1[:, 0] * v2[:, 1]) / den
+    lb = (v0[:, 0] * v2[:, 1] - v2[:, 0] * v0[:, 1]) / den
+    la = 1.0 - lb - lc
+
+    tol = 1e-9
+    inside = np.flatnonzero(alive & (la >= -tol) & (lb >= -tol) & (lc >= -tol))
+    if not len(inside):
+        return np.empty(0, dtype=np.int64)
+    # One containing cell settles it; the first keeps the answer independent
+    # of how the cells happen to be ordered, since every containing cell names
+    # the same face.
+    c = int(inside[0])
+    on = np.array([la[c] > tol, lb[c] > tol, lc[c] > tol])
+    return cell_verts[c][on]
+
+
 def _cell_quality(X, cell_verts):
     """Scale-free triangle quality: 1 equilateral, 0 degenerate, <0 inverted.
 
@@ -839,6 +886,14 @@ def pull_vertex_onto(dm, targets):
     behaviour as ``snap_frac``, and anything optimising over fault geometry has
     to live with it.
 
+    The vertex is chosen from the corners of the cells CONTAINING the target,
+    not from the mesh at large. The globally nearest vertex can belong to a
+    neighbouring cell, and moving that one leaves the containing cell with no
+    corner on the line, so :func:`cut_along_lines` refuses the very tip this
+    call was made to place (issue #542). Restricting the candidates ties the
+    choice to the containment relation the cut actually tests. A target outside
+    the mesh falls back to the nearest vertex anywhere.
+
     The choice is made from the coordinates alone and reduced globally, so every
     rank moves the same vertex: a rank-local nearest-vertex search picks a
     different one on each rank, which is how the mesh stops being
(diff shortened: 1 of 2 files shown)


lmoresi:
## Adversarial review

**1. The rule is geometric, and the first version of it was wrong.** Restricting candidates to the corners of the containing cells is the obvious fix and it does not work: a point on an edge lies in two cells, and the union of their corners still offers each cell's opposite apex. On the measured case that apex still won on distance, and the cut still refused. The shipped rule reads the minimal face off the barycentric coordinates of **one** cell, which is both cheaper and correct. We are flagging this because "restrict to the containing cell" is what a reviewer would propose, and it is not sufficient.

**2. The barycentric tolerance is a real choice.** `tol = 1e-9` on scale-free coordinates decides when a point counts as "on" an edge rather than just inside. Too small and a point a rounding error off the edge keeps the opposite apex as a candidate, which is the defect back again; too large and a point genuinely inside a thin cell loses a legitimate candidate. The second failure is benign — it only removes a far vertex from consideration — so the tolerance errs deliberately toward the on-edge reading. We have not swept it.

**3. Parallel is argued, not measured, for the case that matters.** `ptest_0844` and `ptest_0845` pass at np=2, but neither places a tip on a partition seam, which is where the outside-flag reduction earns its keep. The flag has to lead the tuple: a rank holding no part of the containing cell proposes its own nearest vertex, which can be closer than the right answer, and a reduction on distance alone would take it. That is reasoning about a case the suite does not construct.

**4. The fallback is unchanged behaviour, deliberately.** A target outside the mesh still takes the nearest vertex anywhere. That is the pre-existing contract and callers may rely on it, but it means a target just outside the boundary gets no containment guarantee at all — the guarantee this PR adds is conditional on the target being inside.

**5. What the fix does not do.** It guarantees every cell holding the target ends up with a corner on it. It does not guarantee the resulting mesh is well shaped: the chosen vertex can be the farther of two edge ends, and moving it that distance can flatten a cell. `pull_vertex_onto` has no quality veto — unlike `_resolve_snapping`, which does — and this PR does not add one. Worth knowing before someone places tips on a mesh much coarser than the feature they are placing.

lmoresi:
On the stale-geometry finding — accepted as real, but measured before acting, and the measurement changes what the fix should be.

`pull_vertex_onto` resolves each target against the untouched `X` while `arr` accumulates the moves. Sweeping control-point spacing on an unstructured box (`h = 0.06`), comparing stale-`X` selection against live-`arr` selection:

| spacing | points | selections differing | collisions, stale | collisions, live |
|---|---|---|---|---|
| 1.00 h | 9 | 0 | 0 | 0 |
| 0.50 h | 17 | 0 | **5** | **5** |
| 0.25 h | 34 | 10 | 22 | 25 |
| 0.10 h | 84 | 36 | 72 | 75 |

And on the real geometry that motivated this PR — the listric trace, 22 control points at roughly cell spacing — the two schemes choose **identical** vertices with no collisions.

**So the staleness is latent rather than active**: it needs sub-`0.5 h` control-point spacing before it changes any selection.

**The more serious defect is next to it, and switching to `arr` does not fix it.** Collisions — two targets claiming the same vertex — appear at `0.5 h`, one spacing coarser than staleness matters, and occur *equally* under both schemes (25 vs 22 at `0.25 h`; marginally worse with `arr`). A collision means an earlier control point silently loses the vertex placed on it, which is exactly the state `cut_along_lines` reports as "entered but not left" — the failure this PR exists to fix, reachable by a different route.

Neither scheme excludes already-claimed vertices, and that is the actual repair.

We have therefore **not** made the `arr` change here. It would address the smaller problem, leave the larger one, and read as though the area had been dealt with. Filing the claimed-vertex exclusion as a follow-up with these numbers, and noting the `0.5 h` threshold as the practical limit on control-point density until it lands.

The scripts are `stale_geom.py` and `stale_dense.py` in the session scratch; happy to move them under the study if you would rather they were kept.

PR #548, opened 2026-08-13
Rotated solve: stop counting the pinned row, and report the verdict

lmoresi:
Two defects, both of which made a **converged** rotated solve report that it had not converged. Found while chasing what looked like a nonlinear stall in a fault-contact model.

## 1. The direct path counts a row it removed from the system

`_naive_pressure_pin` fixes the pressure gauge for the LU path by replacing one pressure row with the identity and zeroing its rhs. `rotated_residual` did not exclude that row, the way it already excludes `normal_rows`. The loop therefore measured itself against a floor it had itself defined as unreachable, and ran to `max_it`.

The residual breakdown on a fault-contact problem, per outer iteration:

| iteration | \|F\| | velocity part | pressure part |
|---|---|---|---|
| 0 | 1.731096e+01 | 1.731060e+01 | 1.115377e-01 |
| 1 | 3.274136e-04 | **6.301164e-12** | 3.274136e-04 |
| 2–8 | 3.274136e-04 | **0.000000e+00** | 3.274136e-04 |

The velocity problem is solved at the first increment. Everything after it is the pinned DOF alone — bit-identical, with the increment exactly zero, for eight further iterations that changed nothing. Confirmed by index: the residual's single non-zero row is exactly `_naive_pressure_pin(dm)`.

The tell that it was bookkeeping rather than physics: the reported `rel |F̂| = 1.89e-05` is **identical to three figures with the interface law at η_f = 0, 0.005, 0.05 and 0.5** — a residual that does not move when the physics is deleted is not a physical stall.

The pin is now decided once, before the residual closure is built, and excluded from the residual for the same reason `normal_rows` are: it is not an equation the solve is trying to satisfy, so it is not evidence about convergence. The LU branch reuses that same decision rather than recomputing it.

## 2. The loop never published its verdict

Every SNES call inside the rotated loop evaluates one residual or tangent, so `snes.getConvergedReason()` describes a linear step and stays 0 for the solve as a whole. `petsc_generic_snes_solvers` decides convergence from `snes.getConvergedReason() > 0`, so **every** rotated solve was reported as unconverged — including ones that exited cleanly at rel 1e-8.

The loop now sets `CONVERGED_FNORM_RELATIVE` / `CONVERGED_SNORM_RELATIVE` on a clean exit, and `DIVERGED_MAX_IT` / `DIVERGED_LINE_SEARCH` otherwise.

## Why this matters beyond the noise

A reason code that reads 0 on a converged solve is not merely cosmetic here: it is the value the library itself uses to decide whether a solve worked, and it trains users to ignore the check. In the campaign that found this, four fault models were reported as unconverged and their results treated as provisional; the fields were converged the whole time, which is why an exact LU velocity block and an unconverged GMRES one agreed on slip to four significant figures.

## Tests

`test_the_rotated_solve_reports_its_own_convergence`, parametrised over both paths because the pin exists only in the direct one. **Both parametrisations fail before this change** and pass after.

Passing: `test_1018`, `test_1061`, `test_1062`, `test_1065` (41); `test_1058`, `test_0846`, `test_0847`, `test_0848`, `test_0203`, `test_1021` (63) — the suites that read the converged reason downstream; `test_1064` and `test_1066` at np=2 (11).

### DIFF
--- a/src/underworld3/utilities/rotated_bc.py
+++ b/src/underworld3/utilities/rotated_bc.py
@@ -558,6 +558,14 @@ def _destroy_rotated_linear_cache(cache):
 # --------------------------------------------------------------------------- #
 #  The rotated solve
 # --------------------------------------------------------------------------- #
+# PETSc SNESConvergedReason values, named rather than spelled as integers at
+# the point of use.
+_SNES_CONVERGED_FNORM_RELATIVE = 2
+_SNES_CONVERGED_SNORM_RELATIVE = 4
+_SNES_DIVERGED_MAX_IT = -5
+_SNES_DIVERGED_LINE_SEARCH = -6
+
+
 def _naive_pressure_pin(dm):
     """One owned pressure DOF (datum) for the direct-LU gauge pin — row only, so
     the B^T coupling is kept. Only the direct solve needs it; the iterative path
@@ -842,6 +850,15 @@ def solve_rotated_freeslip(solver, boundaries, remove_rotation_gauge=True,
     # Direct LU per increment: no FMG prolongation / null space to build — the
     # gauge is fixed by the naive pressure pin instead (see _naive_pressure_pin).
     use_lu = bool(getattr(solver, "_rotated_use_lu", False))
+    # The pin is decided ONCE, here, because the residual has to agree with the
+    # operator about it. The direct path replaces the pinned pressure row with
+    # the identity, so that equation is no longer part of the system being
+    # solved; a residual that still counts it can never fall below whatever sits
+    # there, and the loop then iterates to max_it against a floor it has itself
+    # defined as unreachable. Measured before this was hoisted: the velocity
+    # residual reached 6e-12 at the first increment and the remaining |F̂| was
+    # the pinned DOF alone, bit-identical for eight further no-op iterations.
+    lu_pin = _naive_pressure_pin(dm) if use_lu else None
     interface_laws = bool(getattr(solver, "_fault_interface_laws", {}))
 
     # ---- cross-solve workspace cache (issue #417; see the block comment
@@ -1036,6 +1053,10 @@ def solve_rotated_freeslip(solver, boundaries, remove_rotation_gauge=True,
             # in the rotated frame)
             interface.residual_add(solver, uvec, Fh)
         _zero_rows_local(Fh, normal_rows)
+        if lu_pin is not None:
+            # Same reason as normal_rows: the pinned row is not an equation the
+            # solve is trying to satisfy, so it is not evidence about convergence.
+            _zero_rows_local(Fh, [lu_pin])
         if use_pnull:
             sp = Fh.getSubVector(pres_is)
             n_p = sp.getSize()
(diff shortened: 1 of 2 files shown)


lmoresi:
Note on the existing reporting channel, so this reads as complementary rather than duplicative.

`solver._rotated_freeslip_info` already exists — it is the dict `solve_rotated_freeslip` returns, and the `adapt-on-top-faults` skill correctly tells people to read it rather than `snes.getConvergedReason()`. That guidance stands and this PR does not replace it.

What this PR fixes is that the two disagreed, and the library itself trusted the wrong one: `petsc_generic_snes_solvers` decides convergence with

```python
converged = self.snes is not None and self.snes.getConvergedReason() > 0
```

so an internal check was reading a value the rotated path never set. After this change the SNES reason agrees with the info dict instead of contradicting it, and the internal check becomes correct without callers having to know which of the two channels applies to their solve.

The skill's table entry ("rotated free-slip Stokes 'not converged' — `s.snes` isn't the solving object") is worth revisiting once this lands, since the stated symptom will no longer occur.

lmoresi:
## Adversarial review

The diagnosis is right and the one-decision hoist is the correct shape. Two findings, one of them a residual (sic) of the very defect being fixed.

**1. The pressure-nullspace projection re-populates the pinned row after it is zeroed.** In `rotated_residual` the order is: zero `normal_rows`, zero `lu_pin`, then `sp.shift(-mean)` over the whole pressure block — which writes `-mean` into the pinned row. `use_pnull` is decided independently of `use_lu` (line 1028), the combination is reachable, and the regression test itself runs it (`petsc_use_pressure_nullspace = True` with `use_lu=True`). It passes because a compatible datum drives `mean -> 0` with the residual, so the floor decays instead of sitting at O(1) — the original defect in miniature, vanishing asymptotically rather than fixed. One-line change: zero the pin AFTER the projection block (which also unbiases the mean by not averaging in a zeroed entry). We would not merge without this; it costs nothing and removes the only path by which the pinned row can still leak into `rnorm`.

**2. The "answer is right as well as reported right" assertion is `np.isfinite`.** The comment above it correctly names the risk — a reason set unconditionally passes the `reason > 0` check and means nothing — and then a finiteness check cannot discharge it either: a garbage-but-finite field passes. Assert something the solve actually promises: wall-normal velocity on the four rotated walls at ~1e-8 (machine-precision free-slip is the feature's own headline), or the SolCx velocity itself. Either turns the guard from decorative into load-bearing.

**Minor.** `ctx["pin"]` on a workspace cache hit and the freshly decided `lu_pin` agree only because `_naive_pressure_pin(dm)` is deterministic in `dm`. True today; if the pin ever gains state (ordering, ownership), the residual and operator can disagree again — the exact defect class this PR removes. A one-line comment at the ctx reuse recording that invariant is cheap insurance.

The exit-reason taxonomy (FNORM/SNORM on the two clean exits, MAX_IT/LINE_SEARCH otherwise) is faithful to the loop's actual exits, and the parametrized test failing on both paths before the change is the regression discipline we want.

lmoresi:

is semantic.

### Cluster

|---|---|---|---|

that branch or an equivalent.

### #584

open:

  an oversight.
  than to this follow-up batch.

### #548

PR #552, opened 2026-08-13
2-D fault zones reach the domain boundary: the ribbon outcrops (#549)

### LINKED ISSUE #549: 2-D fault zones cannot reach a boundary — the outcrop machinery exists but is never wired into the ribbon path

lmoresi:
Closes #549 (the ribbon half; the fault_split contact side is a separate design decision, tracked there).

## What this does

A 2-D fault zone specified past a domain wall now embeds, with its clipped edge on the wall. This is the 3-D outcrop machinery (#525/#526) brought one dimension down:

- `_occ_assembly_2d` gains the `box=` clip of its 3-D twin — same specify-long contract (polylines may protrude; the assembly is intersected with the domain rectangle in OCC and snapped to exact wall values after meshing). `cad_area` is computed after the boolean, so the meshed-vs-CAD area gate holds unchanged.
- `_split_skin_band` is now dimension-free: facets are triangles against wall planes in 3-D, edges against wall lines in 2-D. One wall only, as before.
- The 2-D case is *simpler* than the 3-D one that ships, and the code says so: the clipped skin is still a closed loop (clipping puts part of the loop ON the wall, it does not open it), so `_outcrop_chain_2d` splits it into the band and an open interior chain; and the cap degenerates from a wall-annulus meshing problem to two straight wall segments. The outcrop fill is therefore one simply-connected polygon — the cavity ring with its wall span replaced by the interior chain (`_outcrop_ring_splice`) — with no separate cap fill at all.
- The splice segments and the band are relabelled with exactly the labels the deleted wall span carried, full closures included, gated by an exact count — mirroring the 3-D cap relabel.

## What still refuses

- An interior ribbon whose cavity reaches the wall (the outcrop path must not widen the interior contract — tested).
- A zone meeting more than one wall (box-edge outcrop bands are not built, matching the 3-D refusal).
- A zone meeting one wall in more than one band (the 2-D face of the same corner problem).
- Wall vertices become deletable only in the open wall's line, never at a domain corner.

## Tests

- Band census with an interior-twin **negative control**: the identical probe counts zero band on a non-outcropping ribbon before it is trusted on the outcropping one.
- The P2 Poisson quadratic-exact oracle on the outcropped mesh with Dirichlet on the remeshed Top — topological gates alone have missed real defects in this machinery before (#518, #520).
- All three refusals asserted with their reasons.
- Parallel: the outcrop embeds at np=2 (band present on owned points, no top-wall edge left without its label, info identical across ranks) and refusals are collective; also run by hand at np=4.
- The interior suites are untouched and pass (18 serial in test_0855, 21 with test_0854; 4 at np=2).

## Base

Branched from `development` with #541 (`feature/fuse-zone`) merged in, because #541 rewrites `_occ_assembly_2d` — the clip is written against the fuse-era code. Once #541 merges this PR's diff reduces to the outcrop work alone; it should land after #541.

### DIFF
--- a/src/underworld3/utilities/place_surface.py
+++ b/src/underworld3/utilities/place_surface.py
@@ -2801,21 +2801,24 @@ def _assembly_skin(points, cells):
     return points[node_ids], skin_local, node_ids
 
 
-def _split_skin_band(skin_xyz, skin_tris, box_lo, box_hi):
+def _split_skin_band(skin_xyz, skin_facets, box_lo, box_hi):
     """Split a skin into its interior part and the wall BAND — the strip of
     the assembly's clipped face lying exactly in a wall plane (the zone's
     outcrop). One wall only; more refuses (box-edge bands are not built).
     Returns ``(interior_idx, band_idx, wall_code)`` with wall_code None
     when nothing touches a wall.
+
+    Dimension-free: facets are triangles against wall planes in 3-D, edges
+    against wall lines in 2-D.
     """
-    codes = np.full(len(skin_tris), -1, dtype=np.int64)
-    for axis in range(3):
+    codes = np.full(len(skin_facets), -1, dtype=np.int64)
+    for axis in range(skin_xyz.shape[1]):
         for side, value in ((0, box_lo[axis]), (1, box_hi[axis])):
-            on = (skin_xyz[skin_tris][:, :, axis] == value).all(axis=1)
+            on = (skin_xyz[skin_facets][:, :, axis] == value).all(axis=1)
             codes[on] = 2 * axis + side
     walls = set(int(c) for c in codes[codes >= 0])
     if not walls:
-        return np.arange(len(skin_tris)), np.empty(0, dtype=np.int64), None
+        return np.arange(len(skin_facets)), np.empty(0, dtype=np.int64), None
     if len(walls) > 1:
         raise NotImplementedError(
             "the zone meets more than one domain wall; box-edge outcrop "
@@ -3038,7 +3041,7 @@ def _gmsh_fill_annulus_3d(shell_xyz, shell_tris, skin_xyz, skin_tris,
         gmsh.finalize()
 
 
-def _occ_assembly_2d(polylines, width, size, assembly="fuse"):
+def _occ_assembly_2d(polylines, width, size, assembly="fuse", box=None):
     """Thicken each polyline into a ribbon, resolve overlaps, mesh.
 
     The 2-D thin volume: a ribbon is the mitred outline of one polyline, and
@@ -3048,6 +3051,13 @@ def _occ_assembly_2d(polylines, width, size, assembly="fuse"):
     the same region; they differ in the internal seams the mesher must
     honour. Returns ``(points, triangles, cad_area)``, the area being that of
     the resolved faces — the union — under either choice.
+
+    ``box = (lo, hi)`` applies the specify-long contract one dimension down
+    from :func:`_occ_assembly_3d`: the resolved faces are INTERSECTED with
+    the domain rectangle, so polylines may protrude — a ribbon reaching the
+    top surface leaves its clipped edge exactly in the wall line (snapped to
+    the line value after meshing, defensively). ``cad_area`` is then the
+    clipped area, so the meshed-vs-CAD gate holds unchanged.
     """
     import gmsh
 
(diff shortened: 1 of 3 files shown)


lmoresi:
Adversarial review (self-review, posted per our convention). We went looking for ways this breaks; findings with evidence.

**1. Exact-coordinate wall identity — inherited limitation, worth knowing.** Band identification, the deletable-wall mask, and the splice all test `X[:, axis] == value` bitwise. On gmsh box meshes wall coordinates are exact and the assembly is snapped, so this holds; on a mesh whose wall has moved (free surface, MMPDE relaxation) the box min/max would be the extreme vertex only and the outcrop would refuse rather than misbehave. This is exact parity with the 3-D path — not new exposure — and the failure mode is a refusal, not a wrong mesh. The curved/deformed-wall generalisation is #555.

**2. Chain orientation at the splice.** The chain is oriented by comparing its two end t-coordinates against the left corner. Ambiguity would need the two ends equidistant from the corner, which requires them coincident — excluded because both ends lie strictly inside the corner interval and are distinct band endpoints. No tiebreak needed.

**3. Labels on ghost copies after the target-rank relabel.** The relabel runs on the surgery rank only. This is safe for the same reason the skin labelling is: the gathered region is rank-interior (the shared-point check refuses otherwise), so every point the relabel touches exists only on the target rank. Evidence: the np=2 and np=4 runs assert no owned top-wall edge is missing its label, and pass.

**4. A wall span deleted beyond the band.** Wall vertices outside the band interval but within carve reach are deleted and their span is replaced by the straight splice segments — geometry preserved exactly because the wall is straight (finding 1's restriction is load-bearing here). The relabel count gate (`2 + n_band` edges, exact) would catch a splice that lost an edge.

**5. What we could not break in testing:** the interior contract (18 serial tests incl. the pre-existing 14, unchanged), area conservation (bitwise gate to 1e-9 relative, exercised in every test), the fill's verbatim-segment gates (any subdivision of a splice segment would refuse), and the three refusals each assert their message.

**6. One deliberate asymmetry.** `place_thin_volume`'s 3-D path refuses a second wall via `_split_skin_band` only; the 2-D path additionally refuses a *second wall contact by the cavity* away from the band ("the carve reached the domain wall in two separate spans"). The 3-D carve handles that case by opening the bowl on the single wall; in 2-D the equivalent would keep old wall edges of dropped cells alive through the rebuild, whose label survival we did not want to assert without a test. Refusal is the conservative choice; the message names `clearance` as the lever.

Issue #551, opened 2026-08-13
evaluate / global_evaluate: one pass over the locator, not five patches (anti-scaling to np=8, a smoothed substitution, an uncontained cell hint)


lmoresi:
# `uw.function.evaluate` / `global_evaluate`: the subsystem needs one pass, not five patches

We have five open issues against the point-evaluation entry points. They are not five
independent bugs. Reading the code and measuring at `development` @ `e475246f`, the
subsystem has **one architectural problem and four defects that sit on top of it**: point
location is done by a hand-rolled kd-tree + half-space walk that has no cheap rejection
path, is re-run several times per call, is not shared between the classifier and the
interpolator, and is bypassed entirely on the one path where its answer is trusted.

Everything below is measured on a 2-D unstructured simplex box, `cellSize=1/100`
(23 264 cells, 46 929 P2 DOF) unless stated, `amr-dev`, Darwin arm64, runs sequential
(concurrent PETSc contends ~26%).

---

## 1. Parallel anti-scaling: `global_evaluate` gets *slower* as ranks are added

Not previously filed. This is the finding with the largest practical cost.

Fixed global query set — 20 000 points, **identical on every rank**, so the answer is
rank-count independent and wall time should be flat or better as `np` grows. Best of 3
calls, max over ranks:

| np | `global_evaluate` (s) | vs np=1 | efficiency T₁/(np·T_np) |
|---:|---:|---:|---:|
| 1 | 0.0790 | 1.00× | 1.00 |
| 2 | 0.1325 | 0.60× | 0.30 |
| 4 | 0.1789 | 0.44× | 0.11 |
| 8 | 0.2232 | 0.35× | **0.044** |

**np=8 is 2.83× slower than serial for identical work.**

`evaluate` on the same fixed global set (timing only — in parallel it classifies against
the rank-local mesh, so the answer is rank-local): 0.0518 / 0.0436 / 0.0382 / 0.0359 s at
np = 1/2/4/8. Efficiency 1.00 / 0.59 / 0.34 / **0.18**. It is nearly flat because
`points_in_domain` sees all 20 000 points on every rank at every np (measured: 60 000
points classified per 3 calls, unchanged from np=1 to np=8) — **per-rank work proportional
to the GLOBAL point count**, the classic signature.

The benign case, for contrast: **rank-local** nodal points (each rank asks only for its own
DOF coordinates) scale acceptably. At `cellSize=1/150` (105 185 P2 DOF), `evaluate` repeat
[...]

Issue #491, opened 2026-08-02
uw.function.evaluate returns negative values for a squared DERIVATIVE expression


lmoresi:
## Summary

`uw.function.evaluate` returns **negative values for a squared derivative**, and negative values for `sqrt(...)` of one. A square cannot be negative, so any invariant built the obvious way — strain-rate second invariant, stress invariant, `|grad T|` — is silently wrong.

Expressions that do **not** contain a derivative are exact, so this is specific to the derivative path.

## Reproducer

Plain box mesh, no cuts, no adaptation, ~20 lines:

```python
import numpy as np, sympy, underworld3 as uw

mesh = uw.meshing.UnstructuredSimplexBox(
    minCoords=(0., 0.), maxCoords=(1., 1.), cellSize=1/8,
    regular=False, qdegree=2)
x, y = mesh.X
f = uw.discretisation.MeshVariable("f", mesh, 1, degree=2)
c = np.asarray(f.coords)
f.data[:, 0] = np.sin(6.0 * c[:, 0]) * np.cos(6.0 * c[:, 1])

pts = np.random.default_rng(0).uniform(0.05, 0.95, size=(4000, 2))

d  = np.asarray(uw.function.evaluate(f.sym[0].diff(x), pts)).ravel()
d2 = np.asarray(uw.function.evaluate(f.sym[0].diff(x)**2, pts)).ravel()
s  = np.asarray(uw.function.evaluate(sympy.sqrt(f.sym[0].diff(x)**2), pts)).ravel()

print("evaluate(df/dx)          min", d.min(),  "max", d.max())
print("evaluate(df/dx)**2       min", (d**2).min())          # correct
print("evaluate((df/dx)**2)     min", d2.min(), "negatives", (d2 < 0).sum())
print("evaluate(sqrt((df/dx)**2)) min", s.min(), "negatives", (s < 0).sum())
print("max |evaluate(d**2) - evaluate(d)**2| =", np.abs(d2 - d**2).max())
```

## Observed

```
evaluate(df/dx)            min -5.6261   max +6.0884
evaluate(df/dx)**2         min +0.0000                 <- correct
evaluate((df/dx)**2)       min -2.6386   negatives 614 of 4000
evaluate(sqrt((df/dx)**2)) min -0.0622   negatives 25
max |evaluate(d**2) - evaluate(d)**2| = 3.3521
```

`sqrt((df/dx)**2)` is `|df/dx|` and cannot be negative.

## Expected

[...]

PR #556, opened 2026-08-13
Point locator: reject foreign points cheaply, locate once, and containment-check the cell hint (items 1-3 of #551, fixes the #432 class)

lmoresi:
# The point locator: reject early, shrink the working set, and check the hint contains the point

Addresses items 1, 2 and 3 of #551. Items 4 (rank-independent collective structure),
5 (pointwise derivative evaluation) and 6 (the cheap follow-ups) are separate PRs.

Three changes, one file each side of the JIT boundary.

## 1. The walk gets a rejection path and a shrinking working set

`Mesh._get_closest_local_cells_internal` handles a point the first containment test
rejects by walking the 50 nearest cell centroids, and the only way out of that loop was
every lost point being found. A point no local cell could own therefore paid all 50
rounds, and every point already located was re-tested on each of them.

Two things now bound it.

**A rejection radius.** Every cell contributes its centroid to the control-point
kd-tree, so a point inside cell *c* is no further from its nearest control point than
from *c*'s centroid, which for a convex cell is within that cell's vertex reach. The
largest local reach is recorded when the kd-tree is built (`Mesh._local_cell_reach`,
invalidated with `_index` by `deform` and `adapt`) and a lost point beyond twice that
distance is rejected before the walk starts. The factor of two is deliberate slack: the
in-cell test admits a thin slab outside each face, and a badly shaped cell expands
further under that slab than a well-shaped one. Two reaches is still about one cell.

**A shrinking working set.** A point leaves the walk as soon as a cell claims it, or as
soon as the (sorted) neighbour distances pass the rejection radius. The nearest
containing centroid wins.

## 2. Classification and interpolation share one location pass

`points_in_domain` located the near-boundary points, returned a boolean mask and threw
the owning cells away; `petsc_interpolate` then located every interior point again.
`Mesh._classify_points_in_domain` returns both — the mask and the cells the
classification actually looked up. `evaluate` passes those to `petsc_interpolate` as
`cell_hints`, which locates only the entries still marked `-1`, and only on the
DMInterpolation cache miss that needs them. `points_in_domain` keeps its signature, its
answer and its cost: it is now a one-line wrapper and does not search on the
interpolator's behalf.

## 3. The cell hint is containment-checked (#432)

Serial simplex meshes assert `"exact"` location capability, which makes the UW3 hint
authoritative and bypasses `DMLocatePoints` — while the hint they handed over was
`get_closest_cells`, a nearest-**control-point** kd-tree lookup with no containment test
at all. On a tetrahedron nothing downstream can rescue that: the reference-coordinate
guard in `petsc_tools.c` is a componentwise box clamp, and the reference tet is not the
reference box. A query on a shared edge was answered by extrapolating the basis of a
cell that does not contain it.

Fixed by option (a) of the three the issue lists: **containment-check the hint**. The
serial-simplex branch now takes `_robust_owning_cells`, the same containment-checked
locator every other authoritative path already used, so the three branches collapse to
one call. Points it returns `-1` for surface in `unlocated_mask` and take the RBF
fallback that is already plumbed.

Why not the other two:

* **(b) restrict the `"exact"` assertion** would push serial simplex meshes back onto
  `DMLocatePoints`, which is slower and re-opens the #390 class of silent drops the
  bypass was added to close.
* **(c) a barycentric ξ test in C** only clamps a wrong cell's reference coordinates to
  that cell's boundary. On a shared edge that returns the edge value of the *wrong*
  cell, which is right only for continuous fields, and it does nothing when the
  nominated cell is not adjacent to the point at all.

Option (a) costs one extra containment test per point on the fast path — measured at
1.0 tests per point for interior, on-face and on-edge queries — and it makes the word
"authoritative" true.

## Measurements

2-D unstructured simplex box, `amr-dev`, Darwin arm64, runs sequential. Probes are the
ones from the #551 evidence dossier, unchanged, plus a cell-identity probe that dumps
the located cell for every point of a battery.

### The headline: does the rejection path recover the anti-scaling? **No.**

`global_evaluate` on the fixed 20 000-point global query set, `cellSize = 1/100`, best
of 3, max over ranks:

| np | before (s) | after (s) |
|---:|---:|---:|
| 1 | 0.0770 | 0.0783 |
| 2 | 0.1305 | 0.1287 |
| 4 | 0.1858 | 0.1848 |
| 8 | 0.2564 | 0.2319 |

Flat. The issue's expectation that item 1 "alone should recover most of §1" is **not
confirmed**, and the reason is worth recording: `points_in_domain` already had its own
cheap rejection — the boundary-sign / `far_from_domain` test — so the migrate claim loop
was never paying the 51× penalty in the first place. Phase attribution after the change
(3 calls, rank 0):

| phase | np=1 | np=4 | np=8 |
|---|---:|---:|---:|
| `Swarm.migrate` | 31.4 % | 62.7 % | 68.8 % |
| …of which `points_in_domain` | 31.2 % | 56.0 % | 58.1 % |

against 31 / 64 / 69 % for migrate before. The residual is exactly the signature the
dossier named: the migrate claim loop classifies the **whole global point set on every
rank on every round**, and the number of rounds grows with rank count. That is item 4
(collective structure), not item 1 — the cost is the kd-tree query plus numpy over
20 000 points per rank per round, with the walk now contributing almost nothing.

Containment work inside those same calls did drop (rank 0, 3 calls, point-tests):

| | np=2 | np=4 | np=8 |
|---|---:|---:|---:|
| `global_evaluate` before | 187 152 | 298 686 | 414 039 |
| `global_evaluate` after | 164 286 | 254 340 | 336 345 |

### Where the 51× actually was: a direct locator call

`_robust_owning_cells` on 4 000 points, `cellSize = 1/100`:

| | before µs/pt | before tests/pt | after µs/pt | after tests/pt |
|---|---:|---:|---:|---:|
| np=1, owned (own nodal coords) | 0.87 | 1.0 | 0.72 | 1.0 |
| np=1, foreign | 7.99 | **51.0** | **0.53** | **1.0** |
| np=4, owned | 0.70–0.73 | 1.0 | 0.66–0.70 | 1.0 |
| np=4, foreign | 9.55–9.67 | **50.7** | **0.77–0.79** | **1.04** |

Foreign / owned wall-clock ratio: 9.2× → 0.7× at np=1, 13.1–13.8× → 1.1–1.2× at np=4.

`evaluate` on the fixed global set (rank-local classification) improves in step, both in
containment work and in time:

| np | point-tests per located point, before → after | `evaluate` (s), before → after |
|---:|---|---|
| 1 | 1.2 → 1.1 | 0.0491 → 0.0500 |
| 2 | 2.28 → 1.10 | 0.0429 → 0.0427 |
| 4 | 3.35 → 1.20 | 0.0382 → 0.0361 |
| 8 | 4.19 → 1.27 | 0.0388 → 0.0357 |

### Containment tests per point, by query class

Cell-identity probe, `tol = _EVAL_FACE_TOL`; 2-D `cellSize = 1/40`, 3-D `cellSize = 1/12`,
4 000 points per set (before → after):

| query set | 2-D np=1 | 2-D np=4 | 3-D np=1 | 3-D np=4 |
|---|---|---|---|---|
| mesh vertices | 1.0 → 1.0 | 1.0 → 1.0 | 1.0 → 1.0 | 1.0 → 1.0 |
| edge midpoints | 1.0 → 1.0 | 1.0 → 1.0 | 1.0 → 1.0 | 1.0 → 1.0 |
| face centroids (3-D) | — | — | 1.0 → 1.0 | 1.0 → 1.0 |
| cell centroids | 1.0 → 1.0 | 1.0 → 1.0 | 1.0 → 1.0 | 1.0 → 1.0 |
| random interior | 1.01 → 1.01 | 38.5 → **1.12** | 6.11 → **1.63** | 41.1–42.4 → **8.9–11.2** |
| on the domain boundary | 1.0 → 1.0 | 12.6–51.0 → **1.0–1.10** | 1.60 → **1.21** | 38.5–43.0 → **6.7–9.1** |
| outside the domain | 51.0 → **1.0** | 51.0 → **1.0** | 51.0 → **1.0** | 51.0 → **1.0** |
| foreign (other ranks' nodes) | 51.0 → **1.0** | 49.1–49.4 → **1.03** | 51.0 → **1.0** | 45.3–45.9 → **1.0** |
| 2 001 points, 1 unfindable | 1.23 → **1.01** | 37.4–39.9 → **1.12** | 14.8 → **1.63** | 41.2–42.9 → **9.1–11.1** |

The 3-D residual (8.9–11.2 for random interior at np=4) is real work, not the old
pathology: those are points near the partition seam whose nearest control point is
inside the rejection radius, so the walk legitimately runs — and it stops as soon as the
sorted neighbour distances pass the radius, which is what turns 42 into 10.

### Answers

The cell-identity probe dumps the located cell for every point of nine query sets
(vertices, edge midpoints, 3-D face centroids, cell centroids, random interior, on the
domain boundary, outside the domain, other ranks' nodes, and a batch poisoned with one
unfindable point) at two tolerances, in 2-D and 3-D, at np = 1, 2 and 4.

**15 of roughly 450 000 located cells differ.** Every one is 3-D, at the loose
evaluation tolerance, in the random-interior set, and every one is a point that **both**
the old and the new cell contain (verified point by point: 15 checked, 0 failures).
These are points sitting within the containment test's ~1e-5 face slab, i.e. effectively
on a shared face, where more than one cell qualifies. Which one was returned was
previously decided by whether some *unrelated* point in the same batch happened to be
findable — the walk kept going and a later round overwrote the earlier answer. It is now
the nearest containing centroid, deterministically. Nothing changed in 2-D, nothing at
`tol = 0`, and nothing for vertices, edges, faces, centroids, boundary or exterior
points.

## Tests

`tests/test_0761_point_locator.py`, 2-D and 3-D, `level_1` / `tier_a`, 18 passed
2 skipped.

* **on-edge / on-vertex / on-face against the closed-form P1 value.** A P1 field is set
  from a high-frequency signal of position, so the reference `(1-t)u_a + t u_b` is
  computed without `uw.function.evaluate` — the same discipline as
  `test_0753_nested_mg_prolongation.py::test_reproduces_an_arbitrary_coarse_field`, and
  for the same reason. Covers `t = 0` (shared vertex), `t = 0.5` (edge midpoint),
  `t = 0.25` (quarter point) and 3-D face centroids.
* **fail-before, validated.** With the fix stashed and the tree rebuilt:
  `3-D P1 evaluation at t=0.25 along cell edges is off by 3.036e-01 at 17 of 1318
  points`. That is #432. Also failing before: `a point outside the 2-D mesh costs 51.0
  containment tests per point`, and `one unlocatable point added 991 (2-D) / 12251
  (3-D) containment point-tests to a batch of 1000`.
* **negative controls.** (i) The nodal signal is measured to vary by more than 0.1
  between an edge and its midpoint, so a wrong-cell interpolant cannot pass by being
  smooth. (ii) `get_closest_cells` is shown to nominate cells that do *not* contain the
  query, so the edge test is pinning something real — 3-D quarter points, where a
  midpoint is (measured) never misassigned and 2-D never is either, which is why #432 is
  a 3-D report; the 2-D case skips with that count in the message. (iii) The containment
  counter is asserted to fire (an owned point must cost at least one test), so the
  bounds above cannot pass by instrumenting nothing.
* **structural, not a stopwatch.** The performance guard counts containment tests per
  point rather than seconds — the count is what the algorithm does, a wall time is what
  the machine was doing at the time. The parallel timings live in this PR, not in a
  brittle assert.
* **contract.** Every cell the locator returns is asserted to contain its point, at both
  tolerances, on random interior points. That is the property the shrinking working set
  must preserve, and it is the one that is actually defined — *which* qualifying cell is
  returned for a point on a shared face never was.
* **item 2.** `_classify_points_in_domain` is asserted to agree with `points_in_domain`
  exactly, to offer no hint for exterior points, and to offer only hints that contain
  their point.

Also run green: `tests/test_1018_rotated_freeslip.py` (22), the evaluate/locator set
(`0503`, `0503_evaluate2`, `0506`, `0507`, `0730`, `0753` ×2, `0755`–`0757`, `0760`,
`0820_in_cell_test_loose_semantics`, `0057_deformed_domain_membership` — 116), and the
fault set `0845`–`0848` (48).

Full gate `pytest tests -m "level_1 and tier_a" -q -p no:cacheprovider
--ignore=tests/test_0050_utils.py`: **622 passed, 19 skipped, 1 xfailed, 0 failed**
(1545 deselected), 409 s. `scripts/check_deprecated_patterns.py`: clean, allowlist
unchanged at 79.

## What this PR does not do

* **The parallel anti-scaling of `global_evaluate` is still there.** It is the migrate
  claim loop classifying the global point set on every rank on every round (#551 item 4),
  not the locator walk. This PR removes the amplifier and measures the residual; it does
  not fix it.
* Item 5 (`evaluate` substituting a smoothed degree-1 projection for any expression
  containing a derivative, #491) is untouched.
* Item 6 — the empty-rank reductions (#405, in flight separately), the `UnitAwareArray`
  guard (#279), and the O(cells) `_mark_faces_inside_and_out` rebuild (0.73 s at 23 264
  cells, redone on every `deform`) — is untouched. The face-marking rebuild is now the
  largest single fixed cost in a cold evaluation after a mesh move.

Two things noticed in passing and deliberately not fixed here (Charter §9):

* `points_in_domain` calls `get_max_radius()`, which is `@uw.collective_operation` and
  does an allgather. Every current caller reaches it on all ranks, so it is not firing
  today — but it is a collective inside the routine `evaluate` uses to decide which
  ranks hold interior points, which is precisely the #314 shape. It belongs with item 4,
  and the surrounding lines are being edited by the #405 work.
* `Mesh._build_kd_tree_index_PIC` has no callers anywhere in the tree.

### DIFF
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -5879,6 +5879,38 @@ class Mesh(Stateful, uw_object):
             Whether to perform strict validation near boundaries
 
         """
+        return self._classify_points_in_domain(points, strict_validation)[0]
+
+    def _classify_points_in_domain(self, points, strict_validation=True):
+        """In/out classification, keeping the owning cells it had to locate.
+
+        ``points_in_domain`` located the near-boundary points and threw the
+        owning cells away, leaving the interpolator to locate them a second
+        time. This hands them over instead, so evaluation locates each point
+        once (#551 item 2).
+
+        Parameters
+        ----------
+        points : array-like
+            Coordinate array in any physical unit system (will be
+            auto-converted).
+        strict_validation : bool
+            Whether to perform strict validation near boundaries.
+
+        Returns
+        -------
+        in_or_not : numpy.ndarray of bool
+            Exactly what :meth:`points_in_domain` returns.
+        cells : numpy.ndarray of int
+            Owning cell for the points the classification actually located,
+            at the evaluation face tolerance (:meth:`_robust_owning_cells`).
+            ``-1`` everywhere else — for an EXTERIOR point that means "not in
+            the local mesh"; for an INTERIOR point it means "not looked up",
+            because the boundary-sign test settled it without a search. A
+            caller that needs a cell for every interior point locates the
+            ``-1`` entries itself, and only when it needs them: nothing here
+            searches on the classifier's behalf.
+        """
         # Convert points to model coordinates using the unified conversion function
         # This handles all coordinate formats: plain numbers, unit-aware coordinates, lists, tuples, arrays
         import underworld3 as uw
@@ -5893,7 +5925,10 @@ class Mesh(Stateful, uw_object):
         max_radius = self.get_max_radius()
 
         if model_points.shape[0] == 0:
-            return numpy.array([], dtype=bool)
+            return (numpy.array([], dtype=bool),
+                    numpy.array([], dtype=numpy.int64))
+
+        cells = numpy.full(model_points.shape[0], -1, dtype=numpy.int64)
 
         # Cd-1 surface mesh: no boundary-face control points exist
         # (see _mark_local_boundary_faces_inside_and_out). Per the
@@ -5901,7 +5936,8 @@ class Mesh(Stateful, uw_object):
         # the manifold; the closest-local-cell test is the right
         # filter, not an inside/outside split.
         if self.boundary_face_control_points_kdtree is None:
-            return self._get_closest_local_cells_internal(model_points) != -1
+            in_or_not = self._get_closest_local_cells_internal(model_points) != -1
+            return in_or_not, cells
 
         dist2, closest_control_points_ext = self.boundary_face_control_points_kdtree.query(
             model_points, k=1, sqr_dists=True
@@ -5929,11 +5965,17 @@ class Mesh(Stateful, uw_object):
         # cell (>= 0) for any point genuinely in/on the mesh and -1 only for
         # true exterior. Serial / non-simplex keep the cell-wall test
         # (bit-identical to the validated baseline).
+        #
+        # Only the robust locator's answer is kept as a cell hint: it is the
+        # same call the evaluation path makes, so keeping it saves a repeat.
+        # The cell-wall test runs at a different face tolerance and its answer
+        # is a classification, not a hint.
         near_boundary = numpy.where(dist2 < 2 * max_radius**2)[0]
         near_boundary_points = model_points[near_boundary]
 
         if self._eval_use_robust_location():
-            in_or_not[near_boundary] = self._robust_owning_cells(near_boundary_points) >= 0
+            cells[near_boundary] = self._robust_owning_cells(near_boundary_points)
+            in_or_not[near_boundary] = cells[near_boundary] >= 0
         else:
             in_or_not[near_boundary] = (
                 self._get_closest_local_cells_internal(near_boundary_points) != -1
@@ -5943,11 +5985,15 @@ class Mesh(Stateful, uw_object):
             chosen_ones = numpy.where(in_or_not == True)[0]
             chosen_points = model_points[chosen_ones]
             if self._eval_use_robust_location():
-                in_or_not[chosen_ones] = self._robust_owning_cells(chosen_points) >= 0
+                cells[chosen_ones] = self._robust_owning_cells(chosen_points)
+                in_or_not[chosen_ones] = cells[chosen_ones] >= 0
             else:
                 in_or_not[chosen_ones] = self._get_closest_local_cells_internal(chosen_points) != -1
 
-        return in_or_not
+        # A point demoted to exterior keeps no hint: it goes to RBF.
+        cells[~in_or_not] = -1
+
+        return in_or_not, cells
 
     @timing.routine_timer_decorator
     def get_closest_cells(self, coords: numpy.ndarray) -> numpy.ndarray:
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -5288,6 +5288,12 @@ class Mesh(Stateful, uw_object):
         control_points_list = []
         control_points_cell_list = []
         centroids_list = []
+        # Largest distance from a cell centroid to one of that cell's own
+        # vertices, maximised over local cells. A convex cell is the convex
+        # hull of its vertices, so every point of it lies within this distance
+        # of its centroid — which makes it the rejection radius the locator
+        # needs (see _get_closest_local_cells_internal).
+        cell_reach = 0.0
 
         for cell, cell_id in enumerate(range(cStart, cEnd)):
 
@@ -5297,6 +5303,8 @@ class Mesh(Stateful, uw_object):
             cell_point_coords = nav_coords[self._coord_rows_for_points(nav_dm, points)]
             cell_centroid = cell_point_coords.mean(axis=0)
             centroids_list.append(cell_centroid)
+            cell_reach = max(cell_reach, float(numpy.linalg.norm(
+                cell_point_coords - cell_centroid, axis=1).max()))
 
             # for face in range(cell_num_faces):
 
@@ -5359,6 +5367,10 @@ class Mesh(Stateful, uw_object):
         self._nav_centroids = numpy.array(
             centroids_list, dtype=numpy.float64).reshape(-1, self.cdim)
         self._centroid_index = uw.kdtree.KDTree(self._nav_centroids)
+
+        # Rejection radius for the lost-point walk. Rebuilt with the kd-tree
+        # (i.e. invalidated by deform / adapt along with _index).
+        self._local_cell_reach = cell_reach
 
         return
 
@@ -5984,6 +5996,10 @@ class Mesh(Stateful, uw_object):
             # CRITICAL: Must return 1D array, not 2D, for Cython buffer compatibility
             return numpy.array([], dtype=numpy.int64)
 
+    # Safety factor on the local cell reach used to reject a query point
+    # before the lost-point walk. See _get_closest_local_cells_internal.
+    _LOCATOR_REACH_MARGIN = 2.0
+
     def _get_closest_local_cells_internal(
         self,
         coords: numpy.ndarray,
@@ -5996,6 +6012,12 @@ class Mesh(Stateful, uw_object):
         be exactly the owning cell, but if the mesh is deformed, this
         is not guaranteed. Also compares the distance from the cell to the
         point - if this is larger than the "cell size" then returns -1
+
+        A point the first containment test rejects is looked for among the
+        nearest cell centroids. Points too far from the local mesh to be in
+        any of its cells are rejected before that walk starts, and a point
+        leaves the walk as soon as a cell claims it, so the walk costs what is
+        still lost rather than what was asked for.
 
         ``on_boundary`` and ``tol`` are forwarded to the in-cell
         containment test (see ``_test_if_points_in_cells_internal``).
@@ -6046,7 +6068,8 @@ class Mesh(Stateful, uw_object):
         self._build_kd_tree_index()
 
         if len(coords) > 0:
-            dist, closest_points = self._index.query(coords, k=1, sqr_dists=False)
+            control_point_distance, closest_points = self._index.query(
+                coords, k=1, sqr_dists=False)
             # >= : valid indices are 0..n-1, and the empty-tree sentinel
             # (0 with n=0) must trip this guard, not index _indexMap (#399).
             if np.any(closest_points >= self._index.n):
@@ -6072,7 +6095,32 @@ class Mesh(Stateful, uw_object):
         cells[~inside] = -1
         lost_points = np.where(inside == False)[0]
 
-        # Part 2 - try to find the lost points by walking nearby cells
+        if lost_points.shape[0] == 0:
+            return cells
+
+        # Part 2 - try to find the lost points by walking nearby cells.
+        #
+        # Reject what cannot possibly be found, first. Every cell contributes
+        # its centroid to the control-point kd-tree, so a point lying in cell c
+        # is at most |p - centroid_c| from its NEAREST control point, and a
+        # convex cell puts that within the cell's vertex reach. A lost point
+        # whose nearest control point is beyond the largest local reach is in
+        # no local cell and the walk has nothing to find for it. Without this
+        # every genuinely foreign point pays the full 50-neighbour walk — 51
+        # containment tests against 1 for an owned point — and the foreign
+        # fraction is exactly what grows with rank count (#551).
+        #
+        # The margin is deliberately loose. The in-cell test admits a thin
+        # slab outside each face (``tol``), and a badly shaped cell expands
+        # further under that slab than a well-shaped one; a factor of two on
+        # the reach covers both with room to spare while still rejecting
+        # anything more than about one cell away from the local mesh.
+        reach = getattr(self, "_local_cell_reach", None)
+        if reach is not None and reach > 0.0:
+            reach = self._LOCATOR_REACH_MARGIN * reach
+            lost_points = lost_points[control_point_distance[lost_points] <= reach]
+            if lost_points.shape[0] == 0:
+                return cells
 
         # Size by the nav-DM cell count, which is what _centroid_index
         # was built from (includes ghost cells on manifold meshes).
@@ -6082,22 +6130,42 @@ class Mesh(Stateful, uw_object):
         num_local_cells = nav_centroids.shape[0]
         num_testable_neighbours = min(num_local_cells, 50)
 
-        dist2, closest_centroids = self._centroid_index.query(
+        centroid_distance, closest_centroids = self._centroid_index.query(
             coords[lost_points], k=num_testable_neighbours, sqr_dists=False
         )
+        # The kd-tree drops the neighbour axis at k == 1 (a rank owning a
+        # single cell); the walk indexes it either way.
+        centroid_distance = centroid_distance.reshape(lost_points.shape[0], -1)
+        closest_centroids = closest_centroids.reshape(lost_points.shape[0], -1)
 
         # This number is close to the point-point coordination value in 3D unstructured
         # grids (by inspection)
 
+        # The working set shrinks: a point drops out as soon as a neighbour
+        # claims it, or as soon as the neighbour distances (sorted, so
+        # monotonic in i) pass the rejection radius. The nearest containing
+        # centroid therefore wins, which also makes the answer independent of
+        # whether some OTHER point in the same batch is findable — previously
+        # a single unlocatable point kept every already-found point in the
+        # test set for all 50 rounds, and a shared-face point could be
+        # reassigned to a further cell in a later round.
+        working = np.arange(lost_points.shape[0])
         for i in range(0, num_testable_neighbours):
 
+            if reach is not None and reach > 0.0:
+                working = working[centroid_distance[working, i] <= reach]
+                if working.shape[0] == 0:
+                    break
+
+            candidate_cells = closest_centroids[working, i]
             inside = self._test_if_points_in_cells_internal(
-                coords[lost_points], closest_centroids[:, i],
+                coords[lost_points[working]], candidate_cells,
                 on_boundary=on_boundary, tol=tol,
             )
-            cells[lost_points[inside]] = closest_centroids[inside, i]
+            cells[lost_points[working[inside]]] = candidate_cells[inside]
 
-            if np.count_nonzero(cells == -1) == 0:
+            working = working[~inside]
+            if working.shape[0] == 0:
                 break
 
         return cells
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -5198,29 +5198,6 @@ class Mesh(Stateful, uw_object):
 
         return arrcopy
 
-    def _build_kd_tree_index_DS(self):
-
-        if hasattr(self, "_index") and self._index is not None:
-            return
-
-        # Build this from the PETScDS rather than the SWARM
-
-        centroids = self._get_coords_for_basis(0, False)
-        index_coords = self._get_coords_for_basis(2, False)
-
-        points_per_cell = index_coords.shape[0] // centroids.shape[0]
-
-        cell_id = numpy.empty(index_coords.shape[0])
-        for i in range(cell_id.shape[0]):
-            cell_id[i] = i // points_per_cell
-
-        self._indexCoords = index_coords
-        self._index = uw.kdtree.KDTree(self._indexCoords)
-        # self._index.build_index()
-        self._indexMap = numpy.array(cell_id, dtype=numpy.int64)
-
-        return
-
     def _coord_rows_for_points(self, nav_dm, points):
         """Row indices into the navigation coordinate array (``_nav_coords``)
         for the given vertex plex points, via the coordinate PetscSection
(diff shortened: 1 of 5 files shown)


lmoresi:
# Adversarial review — PR #556, point locator (items 1-3 of #551, #432 class)

Reviewed at `82145992` (4 commits, merge base `e475246f`), in a clean worktree
(`bugfix/r556-review`), `amr-dev`, Darwin arm64, every measurement below run
sequentially. Baseline numbers come from a second build of the same worktree
with `src/` checked out at the merge base, so before/after differ only in the
two changed files.

We are independent of the implementer.

**Verdict: request changes.** One merge-blocker, with a one-line fix we have
measured. The locator change itself we could not break: the rejection radius
survived every attack we could construct, in 2-D and 3-D, serial and at
np2/np4, on graded, sliver, curved and deformed meshes, against a brute-force
oracle. The blocker is not in the radius — it is that item 3 routes points into
a fallback rung that has never actually run.

---

## MERGE-BLOCKERS

### B1. `uw.function.evaluate` now returns NaN on serial simplex meshes: the RBF fallback item 3 relies on is dead code

The PR justifies option (a) with "points it returns `-1` for surface in
`unlocated_mask` and take the RBF fallback that is already plumbed"
(`_function.pyx:1344-1346`, and the PR body). It is not plumbed. The rung is

```python
vars = mesh.vars.values()          # _function.pyx:1273
...
for var in vars:                   # :1280  dofcount loop
...
unlocated = getattr(cached_info, "unlocated_mask", None)   # :1399
if unlocated is not None and unlocated.any():
    for var in vars:               # :1402  <- iterates NOTHING
```

`mesh.vars` is a `weakref.WeakValueDictionary`, whose `.values()` returns a
**generator**, not a view. The dofcount loop at `:1280` consumes it, so by
`:1402` it is exhausted. Measured directly (debug print compiled into the
rung):

```
RUNG: unlocated is <class 'numpy.ndarray'> sum 1 nvars 0
```

The mask is correct — one point flagged — and the fill loop has zero variables
to fill. The NaN written by `DMInterpolationEvaluate_UW` (`petsc_tools.c:271`)
survives all the way out of `uw.function.evaluate`.

Before this PR the rung never mattered on a serial simplex mesh: the hint was
`get_closest_cells`, which never returns `-1`. Item 3 replaces it with
`_robust_owning_cells`, which does. Same probe, same mesh (3-D simplex box
`cellSize=1/8`, deformed `x -> x**4`), 3000 uniform interior queries, one P1
field:

| build | interior points classified in-domain | locator returns -1 | **NaN out of `evaluate`** |
|---|---:|---:|---:|
| merge base `e475246f` | 2992 | 1 | **0** |
| PR head `82145992` | 2992 | 1 | **1** |

The point is at `(0.5254, 0.2065, 0.3302)`, `points_in_domain` says it is
inside, and `V.rbf_interpolate` at that exact coordinate returns `0.2802` — the
value the rung was supposed to write.

This is a silent NaN in the return value of the single most-used function in
the library, on the most common configuration (serial simplex), reachable on
any mesh where the 50-neighbour walk can miss — i.e. deformed, graded and
adapted meshes, which is the free-surface / adaptivity / fault workload.

**Fix (measured):**

```python
vars = list(mesh.vars.values())    # _function.pyx:1273
```

With that one line applied and rebuilt, the same probe returns **0 NaN** and
the point takes the RBF value. Nothing else changed.

Two riders on the fix, both of which belong in this PR because it is this PR
that makes the rung load-bearing:

* The same exhausted generator is read again at `_function.pyx:1308`,
  `all_continuous = all(getattr(var, "continuous", True) for var in vars)`.
  `all()` over an empty generator is `True`, so **the continuity gate on the
  `"continuous"` location capability has never bound**. Fixing `vars` makes it
  bind — correctly, but for the first time, on cubed-sphere-class warped-hex
  meshes. On `"exact"` meshes it is a no-op either way
  (`_hint_is_authoritative(True) == _hint_is_authoritative(False) == True`,
  measured on a quad box carrying a P1 and a P0 variable), so the exposure is
  bounded, but it needs its own line of measurement before it lands.
* Add a test. The natural one is the probe above: a graded serial 3-D mesh, a
  random interior batch, assert `np.isfinite(evaluate(...)).all()`. It fails at
  PR head and passes with the fix.

Note the pre-existing half of this: in **parallel** the rung was already
reachable before the PR, because parallel already used `_robust_owning_cells`.
Same deformed 3-D mesh, node-point queries:

| np | NaN at merge base | NaN at PR head |
|---:|---:|---:|
| 1 | 0 | 0 |
| 2 | 18 | 18 |
| 4 | 120 | 120 |

So the dead rung is not this PR's bug. Opening the serial door onto it is.

---

## Major

### M1. The answer-identity claim is measured on the point sets that bypass the walk

"15 of roughly 450 000 located cells differ ... nothing changed in 2-D, nothing
at `tol = 0`, and nothing for vertices, edges, faces, centroids, boundary or
exterior points."

We reproduced the 15 exactly, from the artifacts, independently of their
harness (`before/` vs `after/` `cells_d{2,3}_np{1,2,4}.json`, every set, every
rank, every tolerance): 15 differences over 661 962 compared entries, all 3-D,
all at `_EVAL_FACE_TOL`, all in `scatter`/`mixed`, **0 found/lost flips**, and
only **4 distinct query points** (indices 1255, 2439, 3122, 3652) appearing
across ranks and batches. The claim is true as stated.

It is also not the interesting population. Exact mesh vertices, edge midpoints
and face centroids are answered by the **first-pass** control-point test
(control points are `0.99*vertex + 0.01*centroid`), so they never enter the
walk and cannot change. The points that do enter the walk are the ones sitting
*near* a shared vertex/edge/face. On a plain uniform 3-D simplex box
(`cellSize=1/8`, the PR's own mesh class), 2615 near-vertex barycentric queries:

| set | tol | cells differ (new vs old walk) | not double-contained |
|---|---|---:|---:|
| near-vertex barycentric | 0.0 | **922 / 2615 (35%)** | 0 |
| near-vertex barycentric | `_EVAL_FACE_TOL` | **411 / 2615 (16%)** | 0 |
| uniform random interior | `_EVAL_FACE_TOL` | 1 / 3000 | 0 |

and on a deformed (sliver / graded) mesh even uniform random interior points
move (338 / 3000 on the sliver mesh at the eval tolerance). Every single
difference we found, on every mesh, is a point **both** cells contain under the
same test — the contract holds, and the new rule (nearest containing centroid)
is strictly better than the old one (last of up to 50 rounds, dependent on
whether an unrelated point in the batch was findable). We are not disputing the
change; we are disputing the size claimed for it.

It matters because the cell choice **is** the answer for a discontinuous field.
Measured on that same box with a P0 (`continuous=False`) variable, values drawn
uniform on [-1, 1]:

* `uw.function.evaluate` returns the **new** cell's value for 2615/2615 points
  and the old cell's for 2208/2615 — the located cell fully determines the
  answer, as expected;
* over the 407 points whose cell changed, `|P0(new) - P0(old)|` has
  **max 1.935, mean 0.646** on a field of range 2.

P2/P0-discontinuous is the pressure space the fault work runs on. The fault set
(`0845`-`0848`) and `1018` pass (70 passed, 167 s), so nothing is broken — but
"15 of 450 000" reads as "no downstream consumer can notice", and a consumer
evaluating a discontinuous field at node/face coordinates can notice. Please
restate the identity result as what it is: *the walk population moves, and it
moves by O(jump) for discontinuous fields; the sets measured were mostly not
the walk population.*

For the record on the rest of the downstream question: swarm cell ownership
(`swarm.py:5145`, `ddt.py:2401`), `adaptivity.py:752` and
`gradient_evaluation.py:240` all use `get_closest_cells`, which the PR does not
touch; `_dminterpolation_cache` keys on an xxhash of the coordinate bytes plus
dofcount plus policy, so a cache hit implies identical coordinates and cannot
serve a stale hint; `points_in_domain`'s returned mask is bit-identical by
construction (the new `cells[...]` assignments are the only additions, and
`in_or_not[...] = cells[...] >= 0` is the same expression). We could not find a
consumer that depends on the old tie-break other than evaluation of
discontinuous fields.

### M2. The dangerous direction of the rejection radius is untested, and item 2's contract assertion never fires in serial

`tests/test_0761_point_locator.py` pins "a point outside costs O(1) tests" and
"every returned cell contains its point". Neither is the failure mode that
would be silent: **a point that IS in a local cell being rejected**. Nothing in
the file would catch a wrong reach.

That gap is worth closing because the guard is cheap and the harness works. Our
negative control (brute-force oracle, 3-D box, 2000 interior points, sweeping
`Mesh._LOCATOR_REACH_MARGIN`):

| margin | false rejections |
|---:|---:|
| 2.0 (shipped) | **0** |
| 1.0 | 0 |
| 0.5 | 9 |
| 0.25 | 469 |
| 0.1 | 607 |

So the bound is tight at ~1.0 and the shipped 2.0 has a factor-2 margin over
the first observable failure — good, and exactly the kind of number that should
be in the test file rather than in a PR table.

Second, `test_the_classifier_hands_over_the_cells_it_located` ends with

```python
offered = in_or_not & (cells >= 0)
if offered.any():
    assert mesh._test_if_points_in_cells_internal(...).all()
```

In serial `_eval_use_robust_location()` is `False`, so the classifier never
fills a hint and `offered` is **always empty**: measured 0 hints for 1465
interior points in serial, against 462/462 at np4. The strongest assertion in
item 2's only test is skipped in the default (serial) test run. The file
applies the "confirm the probe fires" discipline elsewhere
(`assert per_owned_point >= 1.0`, with a comment explaining why) — apply it
here: assert the serial expectation explicitly (`offered.sum() == 0` in serial)
and require `offered.any()` under MPI.

---

## Minor

### m1. Reach invalidation is asserted in prose only
The comment says `_local_cell_reach` is "invalidated with `_index` by `deform`
and `adapt`". It is, and we checked it: a 2-D box deformed by ×50 gives
`0.0706932 -> 3.53466` (ratio exactly 50.0), deformed back by /50 returns to
`0.0706932`, and 2000 queries in the expanded domain are located with 0 false
rejections. A stale *small* reach is the one thing that silently breaks the
rejection, so pin it with a two-line test.

### m2. `cell_hints` is mutated in place
`_function.pyx:1354`, `cells = np.ascontiguousarray(cell_hints, dtype=np.int64)`
returns the *caller's array* when it is already int64 and contiguous (verified),
and `cells[unhinted] = ...` at `:1362` then writes into it. Today the only
caller passes `cell_hints[in_or_not]`, a fresh fancy-index copy, so nothing is
observable — but `petsc_interpolate` takes `cell_hints` as a documented
keyword. Use `np.array(cell_hints, dtype=np.int64, copy=True)`.

### m3. `_build_kd_tree_index_PIC` would leave a stale reach
It sets `self._index` (and a differently-built `_centroid_index`) but not
`_local_cell_reach`, so if it ever ran after a normal build it would leave the
old reach in place against new geometry — the one stale-value scenario the
`getattr(..., None)` guard does not cover. The PR already notes it has no
callers; we confirmed (one hit in the whole tree, the `def` itself). Delete it
or set the reach — do not leave a second index builder that skips the new
invariant.

### m4. The face tolerance is absolute, the rejection radius is mesh-relative
`_mark_faces_inside_and_out` places the inner/outer control points at a fixed
`±1e-3` in physical units, so the loose test admits a slab of
`tol * 1e-3 ~ 1e-5` **absolute**, while the rejection radius is
`2 * max cell reach`. They cross over when the largest local cell has reach
below ~5e-6, at which point the radius can reject something the containment
test accepts. Not reachable in any realistic model — we built a mesh with the
whole domain 1e-4 across (reach 6.9e-6) and got 0 false rejections — and the
containment test is already meaningless at that scale for its own reasons. But
the two scales should be tied to each other rather than to each other's
accident.

---

## Pre-existing defects surfaced by this review (separate issues, not blockers)

* **The 50-neighbour cap loses genuinely interior points.** Attribution probe
  (a Python transcription of the pre-#556 walk, run against the shipped one on
  the same mesh and points, with the brute-force oracle as referee):
  **newly-lost = 0 in every case we ran** — the PR loses nothing the old code
  found. But both lose:

  | mesh | set | tol | oracle-contained | lost (old) | lost (new) |
  |---|---|---|---:|---:|---:|
  | graded 3-D (`x**4`) | uniform interior | `_EVAL_FACE_TOL` | 3000 | 6 | 6 |
  | graded 3-D | near-vertex | 0.0 | 2615 | 29 | 29 |
  | sliver 2-D (y/200) | uniform interior | 0.0 | 3000 | 2 | 2 |
  | uniform 3-D box | largest cell's vertices | 0.0 | 24 | 2 | 2 |

  This is the population that hits **B1**. Worth an issue of its own: the walk
  should widen `k` (or walk the cell adjacency instead of the centroid kd-tree)
  rather than cap at 50 nearest centroids.
* `points_in_domain` still calls `get_max_radius()`, an `@collective_operation`
  allgather, inside the routine `evaluate` uses to decide which ranks hold
  points. The PR flags it and defers to item 4. Agreed.

---

## What we attacked and could not break

* **The rejection bound is sound by construction and holds in practice.** Every
  cell contributes its centroid to the control-point kd-tree and the reach is
  the max over local cells of `|vertex - centroid|`, so for any point `p` in
  cell `c`, `dist(p, nearest control point) <= dist(p, centroid_c) <= reach_c
  <= max reach`. That holds for P1 simplices and for bi/trilinear quads and
  hexes (their maps are convex combinations of the vertices), independently of
  grading, aspect ratio or how far `p` is from `centroid_c`. Confirmed against
  a brute-force oracle (every query point tested against every local cell with
  the same containment test) with **0 newly-lost points** on: uniform 2-D/3-D
  simplex, quad 2-D, hex 3-D, annulus, spherical shell, graded 2-D/3-D
  (`x**4`, grading 6.2-7.4), sliver 2-D (aspect 200), a 1e-4-wide mesh, a
  6371-wide mesh, and all of the above at np1/np2/np4. `sqr_dists=False`
  genuinely returns distances, not squares (checked), so the comparison is
  dimensionally right.
* **Parallel, graded partition.** 2-D np4 gives per-rank reaches
  `[0.1015, 0.0286, 0.1040, 0.0267]` (spread 3.9x); 3-D np4
  `[0.382, 0.385, 0.164, 0.387]` (2.4x). Across the *global* node set:
  **0 points claimed by 0 ranks** in every configuration, mask identical to
  `points_in_domain` on every rank, **0 hints that do not contain their point**,
  **0 hints offered for exterior points**.
* **Item 2 / the cache.** Cache key is `(xxhash(coords), dofcount, policy)`, so
  a hit implies byte-identical coordinates; the hint is only computed on a
  miss; a stale hint cannot be reused. Splitting the location into two batches
  (classifier + lazy fill) is only answer-safe *because* of item 1 — the old
  walk's answer depended on the rest of the batch. Worth stating in the commit
  message: commit 3 depends on commit 1 for correctness, not just for speed.
* **Item 3 / no performance cliff.** Containment tests per point after the
  collapse to one branch, serial: simplex `cellSize=1/60` 1.004 (interior) /
  1.000 (nodes); quad 60x60 1.000 / 1.000. The claimed 1.0 holds on both the
  paths that previously had their own fast branch. Outside points, measured
  on our own meshes at PR head: 2-D 1.000, 3-D 1.000 tests per point, against
  the 51.0 the merge-base build produces on the same query set.
* **The multi-mesh hint guard (`mesh is hinted_mesh`) is unreachable today.**
  `functions_unit_system._evaluate_impl` rejects an expression spanning two
  meshes before `petsc_interpolate` sees it ("Expression contains MeshVariable
  symbols from 2 different meshes"), so the guard is defensive-only and
  untested. Correct to have; worth knowing it is not exercised.
* **Fail-before, independently reproduced** by rebuilding `src/` at the merge
  base with the PR's tests in place: 7 failed, 11 passed, 2 skipped, including
  `3-D P1 evaluation at t=0.25 along cell edges is off by 3.036e-01 at 17 of
  1318 points` (#432) and `a point outside the 2-D mesh costs 51.0 containment
  tests per point`.

## Test results at PR head

* `tests/test_0761_point_locator.py`: 18 passed 2 skipped (serial), 18/2 at
  np2, 18/2 at np4.
* `tests/test_0845*`-`0848*` + `tests/test_1018_rotated_freeslip.py`:
  70 passed, 167 s.
* Full gate `pytest tests -m "level_1 and tier_a" -q -p no:cacheprovider
  --ignore=tests/test_0050_utils.py`: **622 passed, 19 skipped, 1545
  deselected, 1 xfailed, 0 failed** in 396 s — reproduces the PR's claim
  exactly.
* `scripts/check_deprecated_patterns.py`: clean, allowlist unchanged at 79.

Note what the gate does **not** cover, and why B1 got through it: every
`level_1 / tier_a` evaluation test runs on a quasi-uniform, undeformed mesh,
where `_robust_owning_cells` never returns `-1` for an interior point, so the
dead RBF rung is never asked to run.

## Artifacts

Probes written for this review (scratchpad):
`oracle.py` (brute-force containment oracle, 11 mesh cases),
`attribute.py` (pre-#556 walk transcribed and run side by side),
`negctl_deform.py` (reach-margin negative control + deform invalidation),
`downstream.py` (discontinuous-field consequence, serial fast-path counts),
`nanprobe.py` / `nanprobe2.py` / `parnan.py` (B1),
`parallel_seam.py` (graded-partition parallel contract).

lmoresi:
Response commits `60ab3d63`, `61326ee6`, `de87c7e7`.

**B1 fixed** — `vars = list(mesh.vars.values())`. Fail-before validated by reverting that one line and rebuilding (the new test fails with exactly the injected points NaN). **Bonus: the dead rung was also the cause of the parallel NaNs** — `parnan.py` was 0/18/120 at np1/2/4 at *both* the merge base and the PR head, and is now **0/0/0**. That retires #558, which we filed this morning as pre-existing.

**The `all_continuous` rider — measured, then scoped.** Capability census first: every mesh that reaches the gate in practice (`UnstructuredSimplexBox` 2-D/3-D, `StructuredQuadBox` 2-D/3-D, deformed quad, `Annulus`) is `"exact"`, where the gate cannot change anything; only a warped hex measures `"continuous"`. On a warped hex carrying a P1 and a P0, 1500 interior points:

| policy | P1 error | P1 vs old | P0 vs old |
|---|---:|---:|---:|
| old (gate never bound) | 8.04e-3 | — | — |
| gate over every mesh variable | 1.17e-1 | 62 differ | 62 differ, max 1.71 |
| **gate over the requested variables** (shipped) | 8.04e-3 | **0 differ** | 62 differ, max 1.71 |

Binding it mesh-wide would cost the *continuous* field a factor 15 in accuracy for no correctness gain — one discontinuous variable elsewhere de-authorises everything. Scoped to the requested variables it is what the surrounding comment always claimed, leaves continuous evaluation bit-identical, and moves only the discontinuous field, which is the O(jump) correction the gate exists to make. `test_0503_evaluate.py::test_evaluate_warped_hex_rbf_fallback_no_silent_values` was written for this gate, had been passing vacuously, and now exercises it.

**M2** — `test_a_point_a_local_cell_contains_is_never_rejected` with a brute-force oracle and a margin sweep: 0 false rejections at the shipped margin on every rank at np2/np4, and the 0.25 control fires everywhere (3-D `{2.0:0, 1.0:0, 0.5:4, 0.25:168}`). The vacuous serial assertion in the hint test is gone — serial now asserts that no hints are offered, MPI requires them and checks containment.

**M1** — documented rather than buried: the restated identity claim is in the PR body, a `.. note::` sits in the locator docstring, and `test_a_discontinuous_field_at_a_shared_vertex_takes_a_containing_cell` asserts the value belongs to *a* containing cell (never a specific one, which would over-pin).

**Minors**: m1 reach-invalidation now asserted in tests (ratio exactly 50.0); m2 `cell_hints` copied; m3 `_build_kd_tree_index_PIC` **and** `_build_kd_tree_index_DS` deleted (both set `_index` without the reach, both zero callers); m4 the two scales and their ~5e-6 crossover documented.

**The CI failure on `test_rotated_workspace_deform_invalidates`: not reproduced, and every path this PR touches is measurably inert in that test.** Not in `amr-dev`, not in a freshly built CI-toolchain `dev` environment, not in single/whole-file/CI-batch shape (137 passed in both environments), with or without CI's env vars; locally the error is exactly 0.0 with `|r| = 8.27e-12` bit-identical between the two solves. Specifically: the mesh is a `StructuredQuadBox` and the merge base already routed quad/hex through `_robust_owning_cells` (item 3 changed serial *simplex* only); the 8 locator calls, run with and without the rejection radius, give 0 cell differences over 2270 points; capability is `"exact"` on both sides of the deform so the continuity gate cannot bind; serial hands over no hints. Rather than guess a fix, the test now **attributes** its own failure — equal convergence reports with non-zero error means the operators differ, differing reports mean the linear solves did — and prints both solves' reason/iterations/residuals plus the reach before and after. If it fails again here, we get the diagnosis for free; if it passes, it belongs to #543 as a flake.

Gate 627 passed / 0 failed (622 before, +5 new). Fault set + 1018: 70 passed. Attribution oracle against the pre-#556 walk across 10 mesh cases: newly-lost 0 everywhere.

Issue #512, opened 2026-08-08
adapt/relax follow-ups from the #488 re-review: empty-rank eval_metric skip (3 sites), reconnect invariant guard, vacuous partition test


lmoresi:
Non-blocking findings from the #488 re-review (full detail in the review comment on the PR), tracked here so they survive the merge:

1. **Empty-rank `eval_metric` collective skip — latent hang** at three sites in `_adapt_nested` (`discretisation_mesh.py`: nvb-native `if cur_h.size:`, edge_split `if centroids.shape[0]:`, and the sbr path which `break`s out of the level loop entirely — the worst variant, that rank leaves for good). At np>1 with a field/expression metric, `eval_metric` is `global_evaluate` (collective); a cell-less rank skips it while peers enter. Needs a rank with zero cells to fire, hence latent. Fix pattern: the #488 collective-emptiness treatment (allreduce the predicate; all ranks call or none do).
2. **`reconnect.py:781`** rank-local invariant guard ("a shared point was deleted") — assert-class, fires only on internal contract violation, but would hang peers if it ever fires. Route through the allgather-verdict pattern.
3. **`test_pinned_set_is_partition_independent` is vacuous** — it allgathers the union on every rank then compares the union's size to the max over ranks of that same size; identical by construction, cannot fail. Rewrite as a genuine cross-np fingerprint (the moved-vertex test added in the #488 fix currently covers the defect it missed).

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
Closed by PR #596, with two corrections to what this issue says.

**Item 1's mechanism is not the one stated.** The issue attributes the hazard to
`eval_metric` being `global_evaluate` and therefore collective, with a cell-less
rank skipping it while peers enter. Measured at np=2, `global_evaluate` does not
wait for its peers: with one rank asleep for ten seconds, the other's call
returned in 0.06 s — for in-domain points, and for points stranded outside the
mesh entirely, which is the branch that allgathers. The rank-local
`if cur_h.size:` guards were not hanging anything.

We are recording this because working from the stated premise would send the
next person to "fix" the two engines that were fine, and past the one that was
not.

**Item 1 is nonetheless a live hang, in the sbr path, for a different reason.**
The collective in that loop is the DM refinement, not the metric. The break on
`refine.size == 0` is rank-local — this rank's cells are all fine enough, which
says nothing about its peers' — so the rank that runs out of work leaves the
[...]

PR #557, opened 2026-08-14
A rank with no cells must answer every global mesh query like its peers (#405)

### LINKED ISSUE #405: Empty-rank support, layer 2: evaluate/points_in_domain, radii/centroid reductions, gather_data NaN-stripping

lmoresi:
# Empty-rank support, layer 2: a rank with no cells gets the same global answer as its peers

Fixes #405.

## The defect

A rank that owns **zero cells** raised a rank-local `ValueError` from an unguarded
local reduction — `self._radii.min()` on an empty array — while its populated peers
sat in the matching collective. The job then hung, or aborted asymmetrically.

This is distinct from #314's collective-count divergence and fires *earlier*, before
the interpolation path is reached. It matters because `get_min_radius` feeds
`estimate_dt` and the penalty scaling at ~14 call sites and sits under `Swarm.migrate`
as well as under `evaluate`: a starved rank took down ordinary time-stepping, not just
point evaluation.

## The correctness argument

Every quantity here is **global** — the smallest cell *anywhere*, the domain bounding
box, the largest diffusivity. It happens to be computed from rank-local data, so it
must be reduced across ranks, and a rank with no cells must contribute the **identity
element** of that reduction (`+inf` for a MIN, `-inf` for a MAX, `0` for a SUM) rather
than raise. Every rank then calls the collective and every rank — starved or not —
returns the same value. That is the shape `get_mean_radius` already used; the fix
copies it.

Two supporting rules the diff follows throughout:

- **No rank-asymmetric raise around a collective.** Where a short-circuit is right
  (`points_in_domain` on an empty rank is honestly all-`False`), the collective is
  taken *first*, by every rank, and the short-circuit comes after.
- **A genuinely rank-local path may return early**, because it takes no collective with
  it. The two interpolation guards are of this kind, and say so.

## Sites fixed

From the issue's own list (all verified still present at `development` @ `e475246f`):

| Site | Was | Now |
|---|---|---|
| `discretisation_mesh.py` — `get_min_radius` / `get_max_radius` | `_radii.min()` / `.max()` on an empty array | MPI allreduce MIN/MAX; empty rank contributes `±inf` |
| `discretisation_mesh.py` — `points_in_domain` | rank-local short-circuit placed *before* the collective `get_max_radius`; then interrogated an empty cell set | collective first; then all-`False` on a zero-cell rank |
| `utilities/_utils.py` — `gather_data` | silently stripped NaN rows, so the table's row index no longer equalled rank | stripping is now `strip_nan=True`, off by default and documented |
| `function/functions_unit_system.py` — monotone kNN limiter | kd-tree over an empty source cloud, stencil gather on an empty array | returns the (empty) value unchanged; rank-local, no collective skipped |
| `discretisation/discretisation_mesh_variables.py` — `rbf_interpolate` | same, for the MeshVariable RBF rung | correctly-shaped zeros (matches the already-guarded SwarmVariable path) |
| `meshing/smoothing/metrics.py` — `mesh_metric_mismatch` | KNOWN LIMIT note said this would lift once (1) was fixed | measured: it does not — see below |

Found beyond the list, same class (a global quantity computed from rank-local data):

- **`Mesh.quality()`** chose between its simplex and volume-only branches from
  rank-local evidence. A starved rank collects no triangles, so it ran **three**
  reductions where its peers ran **eleven** — mismatched collective counts, i.e. a hang
  on any simplex mesh with an empty rank. The branch is now decided collectively and
  every reduction is empty-safe. (The quad-box fixtures escaped this by luck: both
  branches coincide there.)
- **`Mesh.physical_bounds` / `physical_extent`** reduced over *this rank's* nodes only,
  so every rank reported a different "domain size", and an empty rank raised. Both now
  share a new `_global_coord_bounds()` which allreduces the box. Kept behaviour-neutral
  otherwise: a suspected double application of the length scale in that path is flagged
  with a `TODO(BUG)` rather than changed here.
- **`estimate_dt`, three implementations**: `_global_max_diffusivity` (`.max()` of the
  diffusivity sampled at *this rank's* centroids), `SNES_NavierStokes.estimate_dt`
  (`.max()` of centroid velocity magnitudes), and
  `SNES_Stokes_SaddlePt.estimate_dt` in the `.pyx` (`.max()` of the local velocity
  DOFs). All three reduced *before* the allreduce, so a starved rank raised while its
  peers waited. These are the priority path named in the issue.

Assessed and deliberately left:

- `SNES_AdvectionDiffusion.estimate_dt` and `SNES_Stokes.estimate_dt` per-element
  reductions — already `len()`-guarded, verified empty-safe.
- `Swarm.estimate_dt` — already has a sanctioned `except (ValueError, IndexError)`
  contributing `0.0` to the MAX.
- `_get_domain_centroids` — already guarded (#399) with a finite sentinel. Only its
  comment changed, to record that `gather_data` no longer strips NaN but the sentinel
  is still required (a NaN row would poison the kd-tree).
- `quality()`'s percentiles and neighbour size-jump — rank-local estimates by
  documented design; they now return `NaN` on a rank with no local distribution rather
  than raising.
- `mesh._radii` per-cell loop in `_get_mesh_sizes`, `stats()` (PETSc `Vec.max()` is
  already global), the XDMF connectivity check (rank-0, file data).

## Item 2 does not lift — measured, not assumed

The issue expected `mesh_metric_mismatch`'s starved-rank limit on field-valued metrics
to lift automatically once the reduction layer was safe. It does not. Measured at np=4
on a starved mesh with this fix in place: `get_min_radius`, `get_max_radius`,
`get_mean_radius`, `points_in_domain`, `quality()` and `estimate_dt` all succeed and
agree, but any path that builds a mesh-variable sub-DM fails below UW3 with
`MPI_ERR_BUFFER` out of `DMCreateSubDM_Plex` → `DMClone` when a rank has no cells. The
KNOWN LIMIT comment now records that, names #314 as the remaining blocker, and says
what was measured.

## Fail-before evidence

Same test file, same fixture, on a build with the fix stashed out:

- **np=2** — `test_radius_accessors_are_global_on_a_zero_cell_rank` **FAILED on one
  rank only** with `ValueError: zero-size array to reduction operation minimum which
  has no identity` (raised at `discretisation_mesh.py:6512`, `get_min_radius`), the
  other rank did not fail, and the run then **HUNG** at the next test. Ended by the
  `mpirun --timeout` guard.
- **np=4** — **three ranks FAILED**, one did not, then **HUNG**. Ended by the timeout
  guard.

That asymmetry is the bug: one rank raises, the rest are still in the collective.

With the fix: 7 passed at np=2 and 7 passed at np=4.

## Tests

`tests/parallel/test_0774_empty_rank_reductions_mpi.py` (7 tests, `level_1` +
`tier_a`, `pytest.mark.timeout(120)` — the pre-fix failure mode is a hang, so every
test is timeout-guarded).

The fixture is `StructuredQuadBox(elementRes=(1, 2))`: two cells, so PETSc must leave a
rank empty at both np=2 (`[2, 0]`) and np=4 (`[0, 0, 0, 2]`).

Two tests exist to keep the others honest:

- `test_premise_some_rank_owns_no_cells` fails if the partition is not actually
  starved, so the suite cannot quietly stop testing the thing it is named for.
- `test_negative_control_rank_local_minimum_would_be_caught` (house negative-control
  rule) computes the rank-local minimum on this partition and asserts the ranks do
  **not** agree — proving the cross-rank agreement assertion is not true by
  construction, and would fire if `get_min_radius` ever returned a rank-local answer.

Oracle for the value: the reported characteristic length is the cell's
centroid-to-corner half-diagonal, `sqrt(0.5² + 0.25²)` for this fixture — analytic, not
a recorded number. Matches the np=1 value to 1e-9.

## Verification

| Run | Result |
|---|---|
| `test_0774_empty_rank_reductions_mpi.py` np=2 | 7 passed |
| `test_0774_empty_rank_reductions_mpi.py` np=4 | 7 passed |
| `ptest_0008_mesh_radii_accessors.py` np=2 / np=4 | OK, values unchanged from before the fix |
| `test_0700_basic_parallel_operations.py` np=2 / np=4 | 12 passed 2 skipped / 14 passed |
| `test_0750_global_statistics.py` np=2 / np=4 | 13 passed / 13 passed |
| `test_0755_swarm_global_stats.py` (a `gather_data` consumer) np=2 / np=4 | 11 passed / 11 passed |
| `tests/test_1018_rotated_freeslip.py` | 22 passed |
| `pytest tests -m "level_1 and tier_a" -q --ignore=tests/test_0050_utils.py` | **604 passed, 24 skipped, 1 xfailed, 0 failed** |

## Unblocks #314

#314 (`global_evaluate` np=4 hang on empty-interior ranks) is the same family seen from
the DMInterpolation side, and could not be worked while the layer beneath it raised
first: any starved-rank reproducer died in `get_min_radius` before reaching
`DMInterpolation`. That layer is now clean, and the measurement above localises what is
left — `DMCreateSubDM_Plex` → `DMClone` returning `MPI_ERR_BUFFER` on a zero-cell rank —
which is a concrete starting point for #314 rather than a symptom. #512's item 1 should
fall out of the same work.

### DIFF
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -9600,7 +9600,10 @@ class SNES_Stokes_SaddlePt(SolverBaseClass):
         if self.mesh.dim == 3:
             magvel_squared += vel[:, 2] ** 2
 
-        max_magvel = math.sqrt(magvel_squared.max())
+        # A rank owning no cells owns no velocity DOFs; it contributes the
+        # identity element of the MAX rather than raising on the empty array
+        # while its peers wait in the allreduce (issue #405).
+        max_magvel = math.sqrt(magvel_squared.max()) if magvel_squared.size else 0.0
 
         from mpi4py import MPI
 
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -1504,6 +1504,21 @@ class Mesh(Stateful, uw_object):
                     val, op=getattr(_MPI, op))
             return val
 
+        # A rank owning zero cells contributes the identity element of each
+        # reduction rather than raising on an empty array (issue #405).
+        def _reduce_min(arr):
+            return _reduce(float(arr.min()) if arr.size else float("inf"),
+                           "MIN")
+
+        def _reduce_max(arr):
+            return _reduce(float(arr.max()) if arr.size else float("-inf"),
+                           "MAX")
+
+        def _local_percentile(arr, pct):
+            # Rank-local estimate (see the docstring); a rank with no cells
+            # has no local distribution to take a percentile of.
+            return float(np.percentile(arr, pct)) if arr.size else float("nan")
+
         tri_vertex_lists = []
         is_simplex2d = cdim == 2
         if is_simplex2d:
@@ -1516,17 +1531,32 @@ class Mesh(Stateful, uw_object):
                     break
                 tri_vertex_lists.append(cell_vertices)
 
-        if not is_simplex2d or not tri_vertex_lists:
+        # Choose the branch COLLECTIVELY. A rank owning zero cells collects no
+        # triangles and would otherwise take the volume-only branch (three
+        # reductions) while its populated peers took the simplex branch
+        # (eleven) — mismatched collective counts, i.e. a hang (issue #405).
+        # A starved rank abstains from the vote instead.
+        has_cells = cEnd > cStart
+        n_simplex_ranks = _reduce(
+            int(has_cells and is_simplex2d and bool(tri_vertex_lists)), "SUM")
+        n_populated_ranks = _reduce(int(has_cells), "SUM")
+        is_simplex2d = (n_populated_ranks > 0
+                        and n_simplex_ranks == n_populated_ranks)
+
+        if not is_simplex2d:
             try:
                 volume = np.abs(np.array(
                     [dm.computeCellGeometryFVM(cell_id)[0]
                      for cell_id in range(cStart, cEnd)]))
             except Exception:
                 volume = np.array([1.0])
-            if not volume.size:
+            # A rank that owns cells but cannot compute their geometry keeps
+            # the unit-volume placeholder; a rank owning NO cells contributes
+            # nothing at all, so the global cell count stays honest.
+            if not volume.size and has_cells:
                 volume = np.array([1.0])
             n_cells = _reduce(int(volume.size), "SUM")
-            vol_min = _reduce(float(volume.min()), "MIN")
+            vol_min = _reduce_min(volume)
             vol_sum = _reduce(float(volume.sum()), "SUM")
             metrics = dict(
                 n_cells=n_cells, element="non-2D-simplex",
@@ -1538,7 +1568,9 @@ class Mesh(Stateful, uw_object):
                 metrics["per_cell"] = dict(volume=volume)
             return metrics
 
-        tri = np.asarray(tri_vertex_lists, dtype=np.int64)
+        # reshape(-1, 3) keeps the (0, 3) shape on a rank with no cells, where
+        # np.asarray([]) would be 1-D and the column indexing below would fail.
+        tri = np.asarray(tri_vertex_lists, dtype=np.int64).reshape(-1, 3)
         v0, v1, v2 = (vertex_coords[tri[:, 0]],
                       vertex_coords[tri[:, 1]],
                       vertex_coords[tri[:, 2]])
@@ -1566,7 +1598,9 @@ class Mesh(Stateful, uw_object):
              _angle_deg(edge_c, edge_a, edge_b)])
         longest_edge = np.maximum.reduce([edge_a, edge_b, edge_c])
         aspect = longest_edge * longest_edge / (2.0 * area)
-        rel_area = area / area.mean()
+        # rel_area only ever masks this rank's own cells, so the rank-local
+        # mean is the right scale — and an empty rank has no cells to mask.
+        rel_area = area / area.mean() if area.size else area
 
         # Neighbour size-jump: map each (undirected) edge to the triangles
         # sharing it; interior edges (exactly two triangles) contribute the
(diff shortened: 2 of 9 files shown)


lmoresi:
CI regression diagnosed and fixed in `6e97b076` — it was ours, and the mechanism is worth recording because it is a type contract that had only ever held by accident.

`np.squeeze(np.float64(0.5))` returns `np.float64(0.5)`; `np.squeeze(0.5)` returns `array(0.5)` — a 0-d ndarray. Every `estimate_dt` funnels its result through `np.squeeze`. Making `get_min_radius` return a plain `float` (what the allreduce naturally produces, and what its `-> float` annotation already promised) therefore promoted the timestep to a 0-d array, and `solve(timestep=dt)` evaluating `timestep != self.delta_t` against a UWexpression made numpy try to convert the sympy object to a sequence instead of deferring to it. Measured, merge base vs head: same value, different type (`np.float64(0.01586914…)` vs `array(0.01586914…)`, ndim 0).

Fixed at the contract boundary rather than by restoring the accident: `_apply_unit_aware_scaling` already documents its return as "float or UWQuantity", so it now collapses a 0-d array to a scalar (unit-bearing values pass through untouched — `.item()` would discard units), and the two `estimate_dt` implementations that return directly use the same helper. Returned numbers are bit-identical to the merge base. `tests/test_1007_estimate_dt_scalar_contract.py` pins the contract so it no longer depends on a caller's numeric type — and it sits in the `test_100[0-9]` batch CI actually runs.

Incidental catch: **`SNES_AdvectionDiffusion.estimate_dt` has been returning a 0-d array since before this branch** — a live trap, now closed by the same helper.

Re-ran the batches this diff can reach that the tier-A gate does not (they are the exposure via `solvers.py` and the `.pyx`): `test_100x` 28 passed (was 1 failed), `101x/102x` 137, `105x` 83, `1100` 5+1 xpassed, `1110/1120` 4, `1450` 3; plus `test_0774` 7 at np=2 and 7 at np=4, and the full tier-A gate 604 passed / 0 failed.

For a reviewer: the part of this PR to look at hardest is the radius accessors' **return type**, consumed at ~14 sites — that is what had the blast radius here.

PR #565, opened 2026-08-14
Mesh files: one directory, atomic writes, and a quick tier that is quick (#563)

### LINKED ISSUE #563: The mesh cache races: '.meshes' is hardcoded in 42 places, so two concurrent UW3 processes in one directory corrupt each other
## What happens

`Mesh._from_plexh5` fails with `petsc4py.PETSc.Error: error code 76` when two processes build the **same** mesh in the same working directory at the same time: one is still writing `.meshes/uw_....msh.h5` while the other opens it for reading.

```
underworld3/discretisation/discretisation_mesh.py:178: in _from_plexh5
    viewer = PETSc.ViewerHDF5().create(filename, "r", comm=comm)
petsc4py.PETSc.Error: error code 76
```

This is not only a testing problem. Any two UW3 runs sharing a working directory hit it — a user running two models side by side, or a parameter sweep launched as concurrent single-rank jobs. The cache filename is derived from the mesh parameters, so *identical* geometry is exactly the case that collides.

## Why it cannot currently be worked around

The cache directory is the string literal `".meshes"`, hardcoded **42 times across 5 modules** (`cartesian.py`, `spherical.py`, `annulus.py`, `segmented.py`, `geographic.py`), with no environment override and no parameter. There is no way to give a process its own cache.

Note this also violates the Style Charter's DRY clause (§5, "DRY after the 2nd occurrence") 42 times over.

## Measured impact on the test suite

Running the level-1 suite under `pytest-xdist` (`-n 4 --dist loadfile`, BLAS threads pinned) takes **2:55 against 9:45 serial — a 3.3x speedup** — but **6 tests fail and 3 error**, every one of them traceable to this race:

- `tests/test_1014_stokes_multigrid.py` (collection error)
- `tests/test_0850_faults.py` — three tests
- `tests/test_0007_snapshot_inmemory.py`, `tests/test_0101_kdtree.py`, `tests/test_0650_recursion_prevention_regression.py`
- `tests/parallel/test_0765_internal_boundary_integral_mpi.py`

So this defect is the single thing standing between us and a 3x faster test suite, locally and in CI.

## Proposed fix

1. One helper owning the cache location — `_mesh_cache_dir()` — reading `UW_MESH_CACHE_DIR` and defaulting to `.meshes`, replacing all 42 literals.
2. `tests/conftest.py` sets it per xdist worker from `PYTEST_XDIST_WORKER`, so workers cannot collide.
3. Write atomically regardless: generate to a unique temporary name and `os.replace()` into place, so a reader either sees the old complete file or the new complete file, never a partial one. This is what makes concurrent *user* processes safe, which (1) and (2) alone do not.

(3) is the real fix; (1) and (2) make it configurable and unblock the test suite.

lmoresi:
Fixes #563.

## The race

Two processes building the **same** mesh in one working directory raced. The generated file is named from the mesh *parameters*, so identical geometry is exactly the colliding case: one process opened `<name>.msh.h5` for reading while the other was still writing it, and PETSc raised error 76. A parameter sweep launched as concurrent single-rank jobs hits this, and so does any parallel test run.

**The fix is atomicity.** Both writes — gmsh's `.msh` and the `.msh.h5` PETSc converts it to — now land under a process-unique name and are renamed into place, so a reader sees a complete file or no file. `_scratch_name` keeps the extension, because gmsh chooses its output format from it.

The rename makes an MPI barrier before the read both **necessary** and **sufficient** — the other ranks must not go looking before rank 0 has renamed — so that barrier is now explicit rather than implied by the write happening to be fast.

`UW_MESH_CACHE_DIR` replaces the `.meshes` string literal that was hardcoded at every site across five meshing modules, and `tests/conftest.py` gives each xdist worker its own directory.

**Negative control, because this is the kind of claim that is easy to get wrong**: with every worker forced to *share* one directory, the run is equally green. So the atomicity is what fixes the race; the per-worker directory only stops four workers redoing the same gmsh work.

## Two generators wrote to the wrong place

`QuarterAnnulus` and `SegmentofAnnulus` created `.meshes/` and then wrote their `.msh` into the **working directory** — the prefix was simply missing from the name. That is where stray `.msh` files in run directories come from, and it made those two generators maximally exposed to the race.

## The quick tier was not quick

`./uw test` advertises "~2 minutes" and took **9:45**. The cause is pytest mark semantics: marks **merge**, so a module-level `pytestmark = pytest.mark.level_1` plus a per-test `@pytest.mark.level_2` leaves that test carrying **both**, `-m level_1` selects it, and the demotion the author wrote does nothing. Nine files rely on that demotion — 24 escalations in total — and the heaviest of their tests is a 96-second homotopy solve, 17% of the whole tier on its own.

A level now selects by excluding the levels above it (`-m "level_1 and not level_2 and not level_3"`). That needs **no change to any test file**; the alternative was adding `@pytest.mark.level_1` to 187 of 213 tests.

## The recursion tests assumed an empty stack

`setrecursionlimit(50)` conflates "this operation does not recurse without bound" with "the whole stack is under 50 frames". Under xdist the worker's own frames spend the budget before the test body starts. The limit is now measured from the current depth, so the tests assert what they mean. (`test_0600` keeps its absolute 100 — it has headroom and passes; left alone deliberately.)

## Measured

Level 1 on a 16-core box, threads pinned:

| | time | result |
|---|---|---|
| before | 9:45 | xdist impossible: 6 failed, 3 errors |
| after, serial | 7:26 | all green |
| after, `-n 4 --dist loadfile` | **2:17** | all green (1175 passed) |

Level 2 also runs green under `-n 4` (693 passed, 4:35), which is the regression check on the meshing change — every one of those tests builds meshes.

## Not in this PR

Turning xdist on by default in `./uw test` and in CI. That is the follow-up now that the race is dead; on CI's 4-vCPU runner it should take the 44-minute suite to roughly 13-15 minutes. Worth doing as its own change so the mechanism and the policy are reviewable separately.

### DIFF
--- a/scripts/test_levels.sh
+++ b/scripts/test_levels.sh
@@ -146,26 +146,34 @@ run_tests() {
 }
 
 # Functions for each test level using pytest markers
+#
+# A level selection EXCLUDES the levels above it, and has to: pytest MERGES
+# marks rather than overriding them, so a file whose module declares
+# `pytestmark = pytest.mark.level_1` and whose heavy test then carries
+# `@pytest.mark.level_2` leaves that test marked BOTH. A plain `-m level_1`
+# selects it, and the demotion the author wrote does nothing. Nine files rely
+# on that demotion; the heaviest of their tests is a 96-second homotopy solve
+# that was running in the "quick" tier because of it.
 run_level_1() {
     echo "⚡ Running LEVEL 1: Quick Tests (Core Functionality)"
-    echo "Using pytest marker: -m level_1"
+    echo "Using pytest marker: -m 'level_1 and not level_2 and not level_3'"
     echo "Expected runtime: ~2 minutes"
     echo ""
 
-    # Run all tests marked with level_1
+    # Tests marked level_1 and NOT demoted to a higher level (see above)
     run_tests "Level 1 tests (quick core functionality)" \
-        tests/ -m level_1
+        tests/ -m "level_1 and not level_2 and not level_3"
 }
 
 run_level_2() {
     echo "🔧 Running LEVEL 2: Intermediate Tests"
-    echo "Using pytest marker: -m level_2"
+    echo "Using pytest marker: -m 'level_2 and not level_3'"
     echo "Expected runtime: ~5 minutes"
     echo ""
 
-    # Run all tests marked with level_2
+    # Tests marked level_2 and NOT demoted to level_3 (see above)
     run_tests "Level 2 tests (units, integration, projections)" \
-        tests/ -m level_2
+        tests/ -m "level_2 and not level_3"
 
     # Parallel tests for global statistics (requires MPI)
     if [ $RUN_PARALLEL -eq 1 ]; then
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -143,9 +143,20 @@ def _from_gmsh(filename, comm=None, markVertices=False, useRegions=True, useMult
             plex_0.setName("uw_mesh")
             plex_0.markBoundaryFaces("All_Boundaries", 1001)
 
-            viewer = PETSc.ViewerHDF5().create(filename + ".h5", "w", comm=PETSc.COMM_SELF)
+            # Write aside and rename, so ``filename + ".h5"`` never names a
+            # half-written file. The name comes from the mesh PARAMETERS, so a
+            # second process building the same geometry in this directory picks
+            # the same one and would otherwise read what this one is still
+            # writing (issue #563).
+            # Imported here, not at module scope: the meshing package imports
+            # this module, so a top-level import would close a cycle.
+            from underworld3.meshing._mesh_files import _scratch_name
+
+            scratch = _scratch_name(filename + ".h5")
+            viewer = PETSc.ViewerHDF5().create(str(scratch), "w", comm=PETSc.COMM_SELF)
             viewer(plex_0)
             viewer.destroy()
+            os.replace(scratch, filename + ".h5")
     finally:
         # The gmsh import options are import-time scratch — meaningful only for
         # the createFromFile above. Clear the whole namespace so a value set by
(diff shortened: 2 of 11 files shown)


lmoresi:
## Adversarial review

We went after the claims rather than the code. Findings, with the evidence that survives them.

**1. "Atomicity fixes the race" is the load-bearing claim, and it is the one we controlled for.** Asserting it from the green run alone would be worthless — the per-worker directory would equally explain it. Forcing every worker to share one directory (`UW_MESH_CACHE_DIR=.meshes/shared`) is still green, so the rename is doing the work. That control is the difference between a fix and a coincidence, and it also tells users something real: a sweep sharing one directory is now safe, not merely tolerated.

**2. The barrier is a behaviour change and deserves its own line.** Before, non-zero ranks read a file rank 0 wrote in place; if they arrived early they saw a partial file and usually got away with it. With the rename they would instead see NO file and fail hard. The barrier converts "usually gets away with it" into "correct", but reviewers should note the failure mode moved from silent-corruption to loud-error before the barrier was added, which is why it is not optional.

**3. `_scratch_name` keeping the extension is not cosmetic.** gmsh selects its output format from the suffix, so the obvious `f"{path}.{pid}.tmp"` would have silently written something that is not a gmsh mesh — and the failure would have appeared much later, in `createFromFile`. Called out because the obvious refactor is wrong.

**4. What we did NOT verify.** True multi-rank behaviour of the new barrier is exercised only by the existing MPI suites, not by a test written for it; and we have not proven the race is gone by reproducing it deterministically — we observed 6 failures + 3 errors before and 0 after, across four runs. A deterministic reproduction (two processes, same geometry, one sleeping mid-write) would be a stronger regression guard than what we have. We think that is worth adding and did not add it here.

**5. The marker change alters what `./uw test` RUNS, not just how fast.** 69 tests leave the quick tier (1244 -> 1175 selected). That is the intent — they are the demoted ones — but it means the quick tier is genuinely a smaller promise than it was yesterday, and anyone reading a green `./uw test` should know it now excludes a 96-second homotopy solve that it previously included by accident. The tests still run at levels 2 and 3.

**6. Scope we deliberately left.** `test_0600`'s absolute `setrecursionlimit(100)` is the same latent defect and is untouched because it passes with headroom; it will bite at higher `-n`. The `os.makedirs(mesh_file_dir(), ...)` call remains repeated at each site — the directory identity is now single-sourced, which is the part that mattered, but the "ensure it exists" line is still copy-paste.

PR #566, opened 2026-08-15
Run the developer test loop across worker processes (CI held serial pending #567)

lmoresi:
Stacked on #565 — the mesh-file race had to die first, or this simply fails. Merge that one before this.

## What changes

Both test runners (`scripts/test_levels.sh` for developers, `scripts/test.sh` for CI) distribute across worker processes with `--dist loadfile`. That keeps every test in a file on one worker, which is the granularity this suite is safe at: tests within a file are written to follow each other, while global state (Model, units, PETSc contexts) does not cross between workers. It is the same reasoning the existing per-file isolation mode and the CI batching already rest on.

**Thread pools are pinned.** Each worker is a full PETSc/BLAS process; without `OMP_NUM_THREADS=1` (and the OpenBLAS/MKL equivalents) they each start their own pool and oversubscribe the machine. This is not a micro-optimisation — measured, an unpinned parallel run is *slower than serial*.

**MPI batches are untouched** — they run their own `mpirun` invocation.

## Why 8 workers and not `-n auto`

Measured on a 16-core box, level 1:

| workers | time | result |
|---|---|---|
| serial | 9:45 | green |
| 4 | 2:17 | green |
| **8** | **1:32** | green |
| 16 | 1:31 | **3 failed** |

Throughput saturates by 8, so the last doubling buys nothing — and it costs something. At one worker per core, three point-locator tests fail. That is **not** numerics degrading under load: it reproduces identically across two runs and is unaffected by `UW_ENABLE_TELEMETRY`. With `--dist loadfile`, changing the worker count changes which *files share a process*, so this is state left behind by whatever shared the worker. Filed as its own issue rather than papered over with a retry.

Default is therefore `min(cores, 8)`, overridable with `--workers N` / `-j N`, or `WORKERS` in the environment for CI.

## `--isolation` now means what it was for

It drops to **one** worker rather than selecting a different mode. Every run is per-file isolated already; `--isolation` additionally removes the concurrency, which is what you actually want when a test passes alone and fails in a full run.

## Result

`./uw test` end to end: **9:45 → 1:22**, 1175 passed.

On CI's 4-vCPU runner this should take the 44 minutes of pytest to roughly 12-15. I have not measured that — the runner is not this machine — so treat the CI number as a projection until the first run on this branch reports it. The projection is the point of the change, but it is a projection.

### DIFF
--- a/.github/workflows/build_uw3_and_test.yaml
+++ b/.github/workflows/build_uw3_and_test.yaml
@@ -28,6 +28,13 @@ env:
   # the UCX transports and retire the whole flake class.
   UCX_TLS: tcp,sm,self
 
+  # Worker processes for the serial batches. Set explicitly rather than
+  # detected: the first run of this change detected ONE core on the runner and
+  # so ran xdist with a single worker — a process spawn and a fresh
+  # underworld3 import per batch, parallelising nothing, for 52m37s against
+  # 49m43s without it. GitHub's standard ubuntu-latest runner has 4 vCPU.
+  WORKERS: 4
+
 jobs:
   test:
     runs-on: ubuntu-latest
--- a/.github/workflows/build_uw3_and_test.yaml
+++ b/.github/workflows/build_uw3_and_test.yaml
@@ -28,12 +28,16 @@ env:
   # the UCX transports and retire the whole flake class.
   UCX_TLS: tcp,sm,self
 
-  # Worker processes for the serial batches. Set explicitly rather than
-  # detected: the first run of this change detected ONE core on the runner and
-  # so ran xdist with a single worker — a process spawn and a fresh
-  # underworld3 import per batch, parallelising nothing, for 52m37s against
-  # 49m43s without it. GitHub's standard ubuntu-latest runner has 4 vCPU.
-  WORKERS: 4
+  # WORKERS is deliberately NOT set: the batches run in-process until #567 is
+  # fixed. Distributing them works and was measured at 4 workers (49m43s ->
+  # 30m53s) but changes which files share a process, which makes three
+  # point-locator tests answer in the wrong cell. Setting `WORKERS: 4` here is
+  # the whole change once #567 lands — and the only change, because
+  # scripts/test.sh already honours it.
+  #
+  # Note if you do set it: detect nothing. The first attempt derived the count
+  # inside the job, got 1, and ran xdist with a single worker — a spawn and a
+  # fresh underworld3 import per batch, parallelising nothing, for 52m37s.
 
 jobs:
   test:
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -64,12 +64,24 @@ export MKL_NUM_THREADS=1
 # cores, throughput saturates by 8, and at one worker per core the file
 # grouping shifts enough to expose test pollution.
 if [ -z "$WORKERS" ]; then
-    _cores=$( (command -v nproc >/dev/null && nproc) \
-              || sysctl -n hw.ncpu 2>/dev/null || echo 2 )
+    _cores=$(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || echo 0)
+    case "$_cores" in
+        ''|*[!0-9]*) _cores=0 ;;          # unparseable: fall through to 2
+    esac
+    [ "$_cores" -lt 1 ] && _cores=2
     WORKERS=$(( _cores < 8 ? _cores : 8 ))
 fi
-echo "Serial batches: $WORKERS worker process(es)"
-PYTEST="pytest --config-file=tests/pytest.ini --dist loadfile -n $WORKERS"
+echo "Serial batches: $WORKERS worker process(es) (detected ${_cores:-preset} core(s))"
+
+# One worker is worse than none: xdist would add a process spawn and a fresh
+# underworld3 import to every batch and parallelise nothing. Measured in CI,
+# where core detection returned 1: 52m37s against 49m43s without xdist.
+if [ "$WORKERS" -le 1 ]; then
+    echo "  (single worker — running in-process, xdist would be pure overhead)"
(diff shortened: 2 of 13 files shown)


lmoresi:
## Adversarial review

**1. The CI number in the description is a PROJECTION, and we said so.** 12-15 minutes on the 4-vCPU runner is extrapolated from a 16-core machine; the runner has different cores, different memory bandwidth, and a cold pixi cache. The first CI run on this branch measures it for real. We would rather publish the projection labelled as one than quietly let the local number stand in for it.

**2. The 8-worker cap rests on a failure we did not root-cause.** We chose `min(cores, 8)` for two reasons — saturation (1:32 vs 1:31 for twice the workers) and the three point-locator failures at 16. The saturation argument alone justifies the cap. The failure argument is weaker: we characterised it (deterministic, telemetry-independent, consistent with file-grouping pollution) and filed #567, but we did not find the polluting file. So the cap is currently also masking something. That is a deliberate trade — a green suite at 8 today, with the underlying defect tracked rather than hidden — but a reviewer should know the cap is doing two jobs.

**3. Thread pinning is the difference between a speedup and a regression.** Worth calling out because it is invisible in the diff's intent: an earlier unpinned run of a comparable selection took 4h44m and 34 CPU-hours against 38 minutes serial. If anyone later "simplifies" the three exports away, the suite does not get slightly slower, it collapses.

**4. What `--isolation` means has changed, and that is a user-visible semantic change.** It used to select per-file subprocessing versus in-process; now every run is per-file and `--isolation` means one worker. Anyone with muscle memory for "isolation = the reliable slow mode" still gets the reliable slow mode, so we think this is safe, but it is a redefinition rather than an addition.

**5. Not verified.** We ran level 1 (1175 tests) and level 2 (693) green under workers; we did NOT run level 3 that way, and the CI script's batches are exercised only by CI itself. If a level-3 physics suite carries cross-file coupling, this change will find it — on the first run rather than in review.

lmoresi:
## The first CI run of this branch was SLOWER, and the projection in the description was wrong

Reporting this against my own change rather than letting the local numbers stand.

**52m37s, against a 49m43s baseline.** The log says why:

```
Serial batches: 1 worker process(es)
```

Core detection returned **1** on the runner, so xdist ran with a single worker: a process spawn and a fresh `underworld3` import added to each of the fourteen batches, parallelising nothing. That is the extra three minutes. The projection of "12-15 minutes" in the description was not merely optimistic — the change as first written was a regression.

The cause is the over-clever detection I wrote, `(command -v nproc >/dev/null && nproc) || ...`. Two fixes are pushed, because either alone leaves the trap armed:

- **The workflow sets `WORKERS: 4`.** GitHub's standard ubuntu-latest has 4 vCPU; guessing it from inside the job is exactly what failed.
- **The script refuses xdist at one worker** and falls back to the plain in-process run. One worker is strictly worse than none, and nothing in the code said so.

Detection is now plain `nproc` → `sysctl` → default, and it **logs the core count it found**, so a wrong answer is visible in the log rather than silently costing three minutes.

The local numbers are unaffected — `./uw test` at 8 workers is measured, not projected. What was unmeasured, and wrong, was the CI extrapolation. The next run on this branch reports the real figure, and I will post it whether or not it is good.

lmoresi:
## Status: the speedup is real, the job is RED, and it should stay red

Second CI run, with the worker count set explicitly:

| | time | result |
|---|---|---|
| baseline (no xdist) | 49m43s | pass |
| first attempt (detected 1 worker) | 52m37s | pass, **slower** |
| **this run (4 workers)** | **30m53s** | **FAIL** |

So the parallelism does what it was supposed to — **1.6x on the real runner**, not the 3-4x I projected from a 16-core box, and that projection remains the thing I got wrong. But the job fails, and the failure is not incidental to this PR.

**Three point-locator tests fail, and they are pointing at something real** (#567):

```
2-D P1 evaluation at t=0.0 along cell edges is off by 1.890e+00
at 829 of 829 points — the query was answered in a cell that does not contain it (#432)
```

Every point, not a tolerance miss. Ruled out since the last comment: it is **not** the per-worker mesh directory (forcing a shared directory reproduces it exactly), not telemetry, not load (deterministic), and not any of the obvious neighbouring files run immediately before it.

**Why it appears now**: on `development` this same batch dies in 9 seconds on the mesh-file race, so xdist never got far enough to reach it. #565 makes it reachable. The defect is not new; the exposure is.

**We are not skipping or xfailing these to get the PR green.** A locator answering in the wrong cell for an entire query set is worth more than the merge is. This PR is blocked on #567, which now has a one-minute local reproduction.

If the speedup is wanted before #567 is understood, the honest interim is to enable workers in `test_levels.sh` (the developer loop, measured 9:45 -> 1:22, all green) and leave `scripts/test.sh` serial — that is a one-line split and I can push it on request. I would rather fix #567.

lmoresi:
## Split, so this can land green

Following the call to take the speedup where it is safe and not to buy the rest with an xfail:

- **`scripts/test_levels.sh` (the developer loop) uses workers.** `./uw test`: **9:45 -> 1:22**, measured, all green. This is the part that matters for people not skipping tests before they push.
- **`scripts/test.sh` (CI) runs in-process**, as it does today. The 4-worker run is real (49m43s -> 30m53s) but exposes #567, and three tests answering in the wrong cell for an entire query set is worth more than 19 minutes of CI.

`scripts/test.sh` still honours `WORKERS` if set, so the parallel run is one environment variable away for anyone bisecting #567 on the runner. Turning CI on after the fix is `WORKERS: 4` in the workflow, and nothing else — the script side is already done and the workflow carries a comment saying exactly that.

What this PR now delivers, all measured rather than projected:

| | before | after |
|---|---|---|
| `./uw test` | 9:45 | **1:22** |
| CI | 49m43s | unchanged (deliberately) |

#567 keeps its one-minute reproduction and stays open with everything already eliminated.

PR #571, opened 2026-08-15
The analytic solution suite: uw.analytic with the full Velic family, an oracle-free validation contract, and four fixed defects

### LINKED ISSUE #498: EnhancedMeshVariable.clone() drops its arguments, breaking six shipped examples
`EnhancedMeshVariable.clone` is declared as `clone(self)` and calls `self._base_var.clone()`, but the underlying `MeshVariable.clone` requires `(name, varsymbol)`:

```
src/underworld3/discretisation/enhanced_variables.py:466
    def clone(self):
        """Clone the variable."""
        return self._base_var.clone()

src/underworld3/discretisation/discretisation_mesh_variables.py:577
    def clone(self, name, varsymbol):
```

So it is broken in both directions: passing the arguments callers actually use raises `TypeError: clone() takes 1 positional argument but 3 were given`, and the no-argument form the wrapper offers would fail inside the base for a missing argument.

Every in-tree caller uses the two-argument form:

- `docs/examples/fluid_mechanics/advanced/Ex_Stokes_Cartesian_SolC.py:141,142`
- `docs/examples/utilities/advanced/Ex_Darcy_1D_benchmark.py:89,90`
- `docs/examples/utilities/advanced/Ex_Darcy_3D_flow_x_axis.py:90,91`
- `docs/examples/utilities/advanced/Ex_Darcy_3D_flow_z_axis.py:92,93`
- `docs/examples/utilities/advanced/Ex_Darcy_3D_Loop_Mesh_Fault.py:218,219`

Each of those examples aborts at that line, so none of them runs to completion. No test covers `clone` at all, which is how it stayed broken.

Found while replacing the Underworld2 shell-out in the SolC example with `uw.analytic.SolC` — the example turned out never to have run, which is also how a mismatched body force survived in it.

Fix is to forward the arguments and add a test. Doing that in the analytic-suite branch since it blocks verifying that work.

lmoresi:
# Land the analytic solution suite: the Velic family as `uw.analytic`, with an oracle-free validation contract

Branched from `development`, merging the long-stranded `bugfix/analytic-module-rename`.
That branch name is misleading — it long outgrew a rename. This is the suite
itself, plus a convention audit and four defect fixes found while landing it.

## What this adds

A single namespace, `uw.analytic`, holding nineteen exact solutions behind one
contract:

- **The Velic Stokes family** — `SolA`, `SolB`, `SolC`, `SolCx`, `SolDA`,
  `SolDB2d`, `SolDB3d`, `SolH`, `SolKx`, `SolKz`, `SolM`, `SolNL`.
- **An elliptical inclusion** (Schmid & Podladchikov 2003), derived rather than
  transcribed — the authors publish pressure, deviatoric stress and rotation
  rate but no velocity, so the Muskhelishvili potentials were recovered from the
  fields and the velocity built from those.
- **A cylindrical Stokes case** (Kramer et al.) on an annulus.
- **Transport and Richards solutions** — `Poisson1D`, `ErfcDiffusion`,
  `AdvectedFront`, `TwoLayerDarcy`, `GardnerSteady`, `GardnerTransient`.

Every one reads the same way:

```python
sol = uw.analytic.SolKz(mesh, B=..., n=3, m=2)
stokes.constitutive_model.Parameters.shear_viscosity_0 = sol.fn_viscosity
stokes.bodyforce = sol.fn_bodyforce
sol.apply_boundary_conditions(stokes)
stokes.solve()
rel = sol.error("velocity", stokes.u)
```

`available()` and `describe()` list the family without constructing it.
Published kernels are vendored under `analytic/_reference/` and **transcribed
into SymPy** rather than compiled, so the solutions are differentiable,
JIT-compilable, and usable as solver input rather than only as point oracles.

`SolKz` is the immediate motivation: a **smooth** viscosity variation with a
known solution, which is what makes a resolution-convergence claim checkable.
`development` currently has only `SolCx`, whose viscosity is a step.

## Validation posture — read this part

Most of the 11k lines are mathematics transcribed from published kernels, where
a sign error is invisible to a build and to any test that compares against the
same transcription. So the question that matters is not "do the tests pass" but
"can the tests fail".

**The oracle is a residual that consults nothing.** `momentum_residual` forms
$\nabla\cdot\sigma + \mathbf f$ by symbolically differentiating the solution's
own Cauchy stress and adding its own body force. It never touches the vendored
kernel and never touches the solver. The scalar families have counterparts
(`transport_residual`, `diffusion_residual`, `richards_residual`), and
`incompressibility_residual` and `strainrate_consistency` complete the set.

This is not tautological, and the design forbids it becoming so:
`AnalyticSolution.set_fields` takes the body force as an independently
transcribed input, and **refuses** a solution that supplies neither a stress nor
a strain rate — the one configuration in which the residual would be true by
construction.

**The negative control fires.** Flipping the sign of the body force moves the
momentum residual from $10^{-16}$ to between 0.5 and 2.0 for every solution that
has one. That is now asserted, not merely observed
(`test_flipping_the_body_force_breaks_the_momentum_balance`). `EllipticalInclusion`
is excluded by name because it is boundary-driven and has no body force to flip;
that exclusion is itself asserted.

**Per-solution strength.** All nineteen registered solutions are swept by
`test_1024_analytic_conformance.py`, which iterates `available()` so a solution
added later is covered the moment it is registered. Two exclusions exist and
both are *declared by the solution*, not listed by name: `symbolic = False` (an
external numeric oracle with nothing to differentiate — `CylindricalStokes`) and
`requires = "..."` for an uninstalled optional dependency. The sweep asserts
that every exclusion has one of those grounds, so a typo in a class attribute
fails the test rather than quietly shrinking coverage.

**Where the validation is weaker, stated plainly.** `reference_agreement`
compares a transcription against the kernel it came from. That tests the
transcription and not the mathematics: if the kernel is wrong, agreement is the
wrong answer. SolM is the worked example — see the errata.

**One gate was partly true by construction, and is now labelled.**
`test_stress_and_strain_rate_agree` asserts $\sigma + p\mathbf I = 2\eta\dot\varepsilon$.
For a solution that supplied only one of stress or strain rate, that is exactly
the identity `set_fields` used to derive the other — structural, not evidential.
Only `SolNL`, `SolDB2d` and `SolDB3d` supply both. The gate is kept (it would
catch `set_fields` regressing) but the docstring now says what it is worth, the
list of solutions publishing both is asserted against the sources rather than
commented, and the genuinely independent check — `strainrate_consistency`, which
compares $\tfrac12(\nabla\mathbf u + \nabla\mathbf u^T)$ against the strain rate
derived from the stress, two separate kernel outputs — has been promoted from a
single solution's file into the family-wide sweep.

## The convention audit

Louis asked for the sign and stress conventions to be unified before this lands,
on the timing argument that nothing depends on them yet. **The audit's finding is
that the exposed conventions are already uniform** — measured, not asserted — and
that the non-uniformity is entirely in the published sources, where it is
absorbed at a single declared boundary. The full table, the per-kernel
provenance, and the errata are in `analytic_errata.md`, written as prose for the
technical note. Summary:

UW3's own conventions win, and the suite already matches them:

| quantity | UW3 | where fixed | suite agrees |
|---|---|---|---|
| total stress | $\sigma = \tau - p\mathbf I$ | `SNES_Stokes.stress` | `set_fields` builds exactly this |
| pressure | positive in compression | same | measured: $\mathrm{tr}\,\sigma + d\,p = 0$ to $10^{-15}$, all solutions |
| momentum | $\nabla\cdot\sigma + \mathbf f = 0$ | `F0 = -bodyforce`, `F1 = stress` | `momentum_residual` forms precisely this |
| $\sigma_{nn}$ | along the domain's **outward** normal | `boundary_normal_traction` (post-#561) | see below |

The sources are *not* uniform: four of thirteen publish the deviator, the rest
the total; component order varies (`solKx.c` uses `[xx, xz, zz]` where every
sibling uses `[xx, zz, xz]`); and several label the vertical velocity `u1`. All
of that is absorbed by the `stress_is_deviatoric` declaration and an explicit
component mapping, honoured in one place. Two things *are* uniform across every
source and neither is documented in most of them — both had to be measured by
finite-differencing the kernels' own outputs: the momentum sign, and pressure
positive in compression.

### Errata (each recorded with its citation, not silently conformed)

1. **SolA's published $\sigma_{zz}$ is missing a factor of $Z$** (`solA.c:156`) —
   *new, and it was live in this suite*. See below.
2. **SolM's published stress is wrong** (`AnalyticSolM.hpp`) — viscosity is
   $(1+\cos)\eta_0 + 1$, stress is computed as $2(\eta-1)\dot\varepsilon$.
   Published stress → residual **0.2117**; derived from the kernel's own strain
   rate → **1.7e-16**. Re-verified independently while landing this: the
   published deviator differs from $2\eta\dot\varepsilon$ by 0.333 relative and
   from $2(\eta-1)\dot\varepsilon$ by **exactly zero**, which is what makes it a
   defect in the source rather than a slip in transcription.
3. **SolC publishes a density, not a body force** (`solC.c`) — as summed 1.848,
   negated 2.5e-16.
4. **SolKz's deviator/total boundary** — this one corrects a statement of *our
   own*. The subsystem doc said `solKz.c` writes its deviator into an array named
   `total_stress`. As a claim about the C function that is wrong: it converts
   explicitly (`u6 -= u5; /* get total stress */`) and says so. What is true is a
   claim about the *transcription's cut point* — the transcriber stops at the
   first mode accumulation, which precedes the conversion, so what we capture is
   the deviator. `stress_is_deviatoric = True` is correct but describes the cut,
   not the kernel. Corrected in the errata.

## Four defects fixed, all found by the oracle

Three were found by a single question: **do the gates still hold away from the
default parameters?** The conformance sweep builds every solution from a mesh
alone, which is right — the defaults are part of the interface — but it means a
coefficient that is unity by default multiplies a term nothing ever looks at.

### 1. SolA is wrong for any viscosity other than 1 (the significant one)

`solA.c:156` computes the $zz$ stress as `u3 = 2.0*kn*ss_z - pp`. The matching
line in `solB.c:140` reads `2.0*Z*kn*ss_z - pp`, and solA's *own* $xx$ stress on
the next line but one carries the `Z`. The shortfall is $\tau_{zz}(1-Z)/Z$,
identically zero at $Z=1$ — the only case the kernel's own disabled driver
exercised, and the default `eta` in our transcription.

| `eta` | momentum | deviator trace | $\dot\varepsilon$ vs $\nabla\mathbf u$ |
|---|---|---|---|
| 1.0 | 0 | 0 | 0 |
| 3.0 | **2.8e-1** | **6.7e-1** | **6.7e-1** |
| 0.25 | **6.4e-1** | **3.0e+0** | **7.5e-1** |

$|1-3|/3 = 0.667$ and $|1-0.25|/0.25 = 3.0$ — the predicted $(1-Z)/Z$ exactly.
Three independent gates, all silent at the default.

Repaired by restoring the missing factor on the term that lost it, declared per
solution (`_zz_stress_lost_the_viscosity`), leaving the vendored source verbatim.
Deriving $\sigma_{zz}$ from $\sigma_{xx}$ via tracelessness would also work and
was **rejected**: it would make the deviator traceless *by construction* and
silently retire one of the three gates that caught the defect. The test asserts
both halves — solA's published deviator is not traceless at $Z=3$, solB's is —
so the control says this is a correction to a defect, not a convention imposed
on the family.

### 2. `EllipticalInclusion` ignores `matrix_viscosity`

At `matrix_viscosity = 3.0` the momentum residual is 0.632 while tracelessness
and strain-rate consistency stay at machine precision — the combination that
localises it. The potentials are normalised to unit matrix viscosity; under
$\eta\to\lambda\eta$ at fixed boundary velocity the stress **and the pressure**
both scale by $\lambda$. `fn_viscosity` was scaled and `fn_pressure` was not, so
the two parts of $\sigma$ were in different units — invisible to any gate that
looks at only one of them. Now 3.9e-15.

Worth noting why it escaped: this is the only solution that assigns `fn_stress`,
`fn_pressure` and `fn_strainrate` directly instead of going through
`set_fields`, which is the one place the stress/pressure relationship is applied.

### 3. `SolNL(r=2.0)` raised `KeyError: 'ComplexInfinity'`

$r=2$ gives $\alpha = 1/r - 1 = -1/2$, at which the published pressure's
denominator (`AnalyticSolNL.c:52`) vanishes **identically** — for every
wavenumber, so it is a pole of the solution rather than a bad $(r,n)$ pairing.
It now raises a `ValueError` naming the cause; previously it produced
ComplexInfinity in three fields and surfaced as a `KeyError` from SymPy's code
printer, a long way from the cause.

### 4. `MeshVariable.clone` is still a `NameError` on `development`

Unrelated to the analytic suite, but it is in the diff and it is the one merge
conflict. `development` fixed #498 via the quickfix batch (#532) with
`newMeshVariable = MeshVariable(...)` inside
`discretisation_mesh_variables.py` — where the only class defined is
`_BaseMeshVariable` and `MeshVariable` **is not a name**. The branch's version
uses `uw.discretisation.MeshVariable(...)`, which works; the merge took it. The
`EnhancedMeshVariable` half now delegates to the base rather than duplicating
the constructor call, since duplicating it is what let the two halves of #498
drift apart.

## The rename, and the shim

`underworld3.function.analytic` → `underworld3.analytic`. The suite outgrew
`function`: it now carries boundary conditions, error norms and a registry, none
of which belong under "symbolic function evaluation".

**Decision: move every caller AND keep a deprecating shim.** The old path is a
public namespace and users' scripts live outside this repo, so a hard break is
not something we can audit. The shim:

- resolves names through `__getattr__`, so the `DeprecationWarning` fires when a
  name is *used*, not when the module is imported — ten test files import it at
  collection time and a warning there is noise the reader cannot act on;
- returns **the same objects**, not copies, so `isinstance` and pickling work
  across both paths;
- is a package *directory* rather than a module file on purpose: the old
  `function.analytic` was a compiled extension, and an orphaned `.so` from an
  earlier install would otherwise be imported in preference to the shim. Python
  checks for a package directory before an extension of the same name, so the
  directory always wins.

`test_1016_analytic_contract.py` tests the shim, including that orphaned-`.so`
case, and is therefore the one file that still imports the old path deliberately.

Callers moved: `tests/test_1015`, `test_1017`, `test_1018`, `test_0835`,
`test_0836`, `test_0839`, `test_1062`, `tests/parallel/test_1017`,
`tests/parallel/test_1064`, `docs/examples/snesfas_investigation/benchmark_3way.py`,
and `docs/examples/.../Ex_Stokes_Cartesian_SolNL.py` (rewritten to use
`uw.analytic.SolNL` rather than three low-level kernel bindings). Docs at
`docs/api/function.md`, `docs/api/analytic.md` and `docs/api/index.md` match.

One nuance: `test_1015` uses the low-level `AnalyticSolCx_*` bindings, which are
the **vendored kernel** and deliberately not part of the `uw.analytic` public
surface — that namespace exposes solutions, not the functions they were
transcribed from. Those imports now go to
`underworld3.analytic._reference._velic`, where they actually live, rather than
through the shim.

## Drift: 152 commits

The merge itself was one conflict (above). Everything else was checked by
running, not by reading:

- **#561 (rotated nodal normal, which reversed inner-boundary $\sigma_{nn}$)** —
  no impact, and checked specifically. `CylindricalStokes` is the only solution
  on a curved boundary and applies `add_rotated_freeslip_bc(0.0, wall)`; the
  datum is zero and $\mathbf u\cdot\hat n = 0$ is invariant under
  $\hat n \to -\hat n$. **Nothing in the suite reads `boundary_normal_traction`
  or `dynamic_topography` back**, so no error measure consumes the #561 sign.
  The one place the suite states a traction sign is `SolCx.topography_top`,
  returning $-\sigma_{zz}$ on the box top where the outward normal is $+\hat z$
  — the same sign as UW3's $h_\infty = -(\sigma_{nn}-\overline{\sigma_{nn}})/\rho g$.
  Consistent. One difference recorded: `boundary_normal_traction` returns its
  result mean-removed and `topography_top` does not, so a caller comparing them
  must remove the mean itself (`test_1018` does).
- **#556 (locator/evaluate), #557 (empty-rank reductions), #534 (solver
  options), #543 (rotated workspace cache)** — no breakage.

Nothing in the suite needed fixing for the drift. The defects fixed here are
older than the drift and were found by the audit, not caused by it.

## Test evidence

All runs in the `analytic-suite` worktree against its own pixi environment
(`amr-dev`), sequentially, on a machine also running other sessions.

**The suite's own tests** — `test_101[5-9]_analytic*` and `test_102[0-8]_analytic*`:

```
382 passed, 1 failed in 1063.60s (0:17:43)
```

The one failure was a guard added in this PR, and it was a genuine false
positive in the guard rather than in the code — see "Record what set_fields was
given" below. After the fix, the two affected tests pass (`14 passed, 84
deselected in 65.07s`). Every other test in the suite passed on the first run
after the merge, before any of the fixes here: **270 passed** across the eleven
pre-existing analytic files, which is the answer to "did a fortnight of API
change break it" — it did not.

**The tier-A gate** — `pytest -m "level_1 and tier_a" tests/`:

```
1021 passed, 26 skipped, 1554 deselected, 1 xfailed, 78 warnings in 1491.45s (0:24:51)
```

All 26 skips are `need --with-mpi option to run` plus two self-documenting
point-locator skips; the xfail is the tracked 1-manifold evaluate limitation.
Nothing related to this change.

**`scripts/test.sh`** — the full serial suite, end to end, because the tier-A
gate and what CI actually runs are not the same set:

```
129 passed                                   test_00[0-4]*
 43 passed                                   test_005[1-9] test_006[0-1]
168 passed                                   test_01*
 27 passed                                   test_02*
532 passed,  5 skipped,  9 xfailed, 1 xpassed  test_05* test_07*
605 passed, 11 skipped, 11 xfailed            test_08*   (units)
 30 passed                                   test_100[0-9]  (Poisson/Darcy)
522 passed                          20:30    test_101* test_102*   <- this PR
 83 passed                          05:59    test_105*
  5 passed,  1 xpassed                       test_1100*
  4 passed                                   test_1110* test_1120*
  3 passed                                   test_1450*
 74 passed,  1 skipped                       named (JIT, docstrings, projections)
-----
2225 passed, 17 skipped, 20 xfailed, 2 xpassed, 0 failed

✅ All tests passed!
SCRIPT_EXIT=0
```

The `test_101*/test_102*` batch is the one this PR changes most, and it is also
the batch that previously matched no CI glob (fixed under issue #504). It now
carries the whole analytic suite plus the new parameter sweep, at 20:30.

**`scripts/test.sh --parallel-only --p 2`** — run because this PR edits two files
under `tests/parallel/`:

```
38 passed,  1 skipped                          in 3.42s
20 passed,  2 xfailed, 5 xpassed               in 68.05s
0 FAILED / 0 ERROR

✅ All tests passed!
SCRIPT_EXIT=0
```

The xfail/xpass are the pre-existing #564 free-slip partition-dependence family
(`test_1063`, `test_1064`, `test_1066`), each carrying its own long explanation
in the test file recording that it is not caused by #560/#561 and that it passes
locally on macOS/arm64 (hence `strict=False`). Unrelated to this PR; noted here
so the counts are not read as new.

**The off-default parameter sweep**, run directly as a probe before and after
the fixes — this is the measurement that found three of the four defects:

| | before | after |
|---|---|---|
| SolA `eta=3.0` | momentum 2.84e-1, trace 6.67e-1, edot 6.67e-1 | 0, 0, 0 |
| SolA `eta=0.25` | momentum 6.35e-1, trace 3.00e+0, edot 7.50e-1 | 0, 0, 0 |
| SolA `eta=3, n=2, m=1.5` | momentum 2.86e-1 | 4.44e-16 |
| EllipticalInclusion `matrix_viscosity=3.0` | momentum 6.32e-1 | 3.91e-15 |
| SolNL `r=2.0` | `KeyError: 'ComplexInfinity'` | `ValueError` naming the pole |

38 other configurations across 13 solutions passed both before and after, so the
fixes are localised and did not perturb anything else.

### One thing worth flagging separately

`pytest -m "level_1 and tier_a"` **with no path argument** — the exact command in
`CLAUDE.md` under "Quick validation" — dies silently: exit 1, zero bytes of
output, no traceback. Adding `tests/` fixes it (pytest otherwise collects from
the repo root, including `.pixi/` and `petsc-custom/`). This is pre-existing and
unrelated to this PR, but it is the "silent pytest death" signature and it means
the documented quick-validation command has been reporting failure without
saying why. Worth its own issue.

## Not done

- `CylindricalStokes` remains `symbolic = False` — it is an external numeric
  oracle, so the residual gates cannot apply to it. That is declared by the
  class and asserted by the sweep, not silently skipped.
- The parameter sweep moves coefficients; it does not sweep *meshes* or
  resolutions. Convergence-rate testing is a separate concern.
- `docs/docstrings/review_queue.md` still lists the removed
  `src/underworld3/function/analytic.pyx`. It is generated by
  `scripts/docstring_sweep.py` and should be regenerated rather than hand-edited.


lmoresi:
The CI result was a **60-minute job timeout**, not a failure — zero FAILED lines, analytic files at 81% and passing. Measured from the log, the suite cost **26.6 minutes** against a run (#568, same day, without it) that already took 55m07s. Fixed in `01f542ae`.

**The profile overturned the obvious hypothesis.** There is exactly **one** Stokes solve in the entire suite; the cost is symbolic — every residual gate differentiates the solution's expressions and runs CSE over the result. Five solutions are 565s of 1010s (SolDA 187, SolKz 187, SolKx 72, SolH 52, SolC 35); the other eight total ~17s. SolC sums forty modes; SolKx/SolKz carry exponential viscosities.

**A real redundancy, removed**: `momentum_residual` sampled the *symbolic* sum of its terms and then sampled each term again for the scale — but the terms are sampled anyway, so the residual can sum **numerically** for free. Same numbers to round-off, eight orders of margin against the 1e-8 gate, negative control still exactly 2.000. Applied to the momentum, incompressibility, transport, diffusion and strain-rate gates. `test_1021`/`test_1023` were also rebuilding their solutions per test (SolKz 6s, SolDA 12s per build) — now cached.

**Split by file placement, not marker** — `scripts/test.sh` globs by file, so a `slow` marker alone would need every batch line to remember to deselect it. `tests/analytic_full/` is a subdirectory the globs do not recurse into, so it is excluded by construction, and the `level_2` mark additionally keeps it out of `pytest -m "level_1 and tier_a" tests/`, which does recurse. All three collection paths verified: 0, 0, 189.

Solutions declare their own tier (`expensive_to_validate` on the class), so the split has one source of truth, with two anti-drift guards: `test_1024` asserts every solution it skips is *named* in the full-family file, and that file asserts its own list matches the declarations.

**No guard weakened.** Momentum residual, incompressibility, tracelessness, `strainrate_consistency` and the body-force negative control run in **both** tiers, on every solution that tier covers — fewer solutions per run, never fewer checks per solution. Critically, every solution that caught a defect in this PR (SolA, SolM, SolNL, SolDB2d/3d) is in the cheap group. SolKx/SolKz keep one canonical parameter case per-PR; the three cases moved out are picked up exactly in the full sweep's table.

| | before | after |
|---|---|---|
| analytic files | 16m50s | **4m07s** (307 passed) |
| CI batch `test_101*`+`test_102*` | 20m30s | **7m30s** (446 passed) |
| full family (opt-in) | — | **8m34s** (189 passed, exit 0) |

Scaling by the observed CI/local factor (26.6 min CI vs 16m50s local, ≈1.6×), the suite's CI cost should land near 6.5 minutes. Running *both* tiers now costs less than the old single tier did.

Full family, one documented command (in the module docstring and `docs/developer/subsystems/analytic-solutions.md`):

```bash
pixi run -e amr-dev python -m pytest tests/analytic_full/ -v
```

PR #581, opened 2026-08-16
Add mesh-files-only internal spherical shell generation

gthyagi:
## Summary

- extract the existing atomic Gmsh-to-PETSc-HDF5 conversion into `_gmsh_to_h5()`;
- add `SphericalShellInternalBoundary(write_mesh_files_only=True)`;
- return the generated `.msh.h5` path without constructing an in-memory UW3 `Mesh`;
- preserve the existing full-`Mesh` behavior by default;
- add focused regression coverage for bypassing `Mesh` construction and reloading the generated HDF5 with `Lower`, `Internal`, and `Upper` labels.

## Motivation

`SphericalShellInternalBoundary()` currently writes the reusable `.msh.h5` file during Gmsh import, then reloads that file into a full UW3 mesh and materializes `Inner` and `Outer` cell-region labels. The region creation loops over every cell in Python.

That work is required when the caller needs the returned in-memory mesh, but it is unnecessary for workflows that only generate reusable mesh files for later MPI jobs. The custom region labels are created after the HDF5 write and therefore do not alter the already-written reusable file.

This became significant for the Zhong et al. spherical-shell benchmark on Gadi. Each `cellSize=1/64` mesh contains about 4.2 million cells. The approximately 710 MB HDF5 files appeared several minutes before the serial jobs exited; one measured job completed in 15:14 using 8.35 GB while remaining CPU-bound in post-conversion mesh setup.

## API

```python
h5_path = uw.meshing.SphericalShellInternalBoundary(
    radiusOuter=1.0,
    radiusInternal=0.775,
    radiusInner=0.55,
    cellSize=1.0 / 64.0,
    filename="shell.msh",
    write_mesh_files_only=True,
)
```

With `write_mesh_files_only=True`, the generator:

1. writes the Gmsh `.msh` file;
2. atomically converts it to PETSc `.msh.h5`;
3. preserves the imported physical boundary labels;
4. returns the HDF5 path as a string;
5. skips HDF5 reload, coordinate-system initialization, and cell-region classification.

The default remains `write_mesh_files_only=False`, so existing callers continue to receive a fully initialized UW3 `Mesh` with no behavior change.

## Validation

Validated on a fresh branch from `upstream/development` at `59982834`:

- `./uw build` completed successfully;
- file-generation regression passed;
- unchanged full-constructor boundary-area regression passed;
- focused result: `2 passed`;
- Python syntax and whitespace checks passed.

The same feature diff also passed the complete level-1 suite before being squashed for this PR: `1493 passed, 33 skipped, 2 xfailed`.

### DIFF
--- a/src/underworld3/discretisation/__init__.py
+++ b/src/underworld3/discretisation/__init__.py
@@ -23,3 +23,4 @@ from .discretisation_mesh import checkpoint_xdmf
 from .discretisation_mesh import meshVariable_lookup_by_symbol
 from .discretisation_mesh import petsc_dm_find_labeled_points_local
 from .discretisation_mesh import _from_gmsh
+from .discretisation_mesh import _gmsh_to_h5
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -96,18 +96,15 @@ def extend_enum(inherited):
 
 
 @timing.routine_timer_decorator
-def _from_gmsh(filename, comm=None, markVertices=False, useRegions=True, useMultipleTags=True):
-    """Read a Gmsh .msh file from `filename`.
+def _gmsh_to_h5(
+    filename,
+    markVertices=False,
+    useRegions=True,
+    useMultipleTags=True,
+):
+    """Convert a Gmsh file to PETSc HDF5 without loading the resulting DM."""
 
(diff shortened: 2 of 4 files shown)


gthyagi:
Superseded by #592 after renaming the fork branch from `codex/spherical-shell-mesh-files-only` to `feature/spherical-shell-mesh-files-only`. The commit, draft status, and PR description are unchanged.

Issue #555, opened 2026-08-13
Outcrop on curved boundaries: the wall descriptor is (axis, value) everywhere


lmoresi:
The outcrop machinery (3-D: #525/#526; 2-D: PR #552) is restricted to axis-aligned box walls, and the restriction is ours, not gmsh's. OCC's `intersect` clips against any tool solid — a disk, cylinder, or revolved shell works exactly like `addBox` — and gmsh can mesh cap patches on curved OCC surfaces or in a local chart. What blocks a curved outcrop (a fault daylighting on an annulus top, a spherical surface) is that the wall is represented as the pair `(axis, value)` threaded through everything:

- band identification is coordinate-equality against the plane/line value (`_split_skin_band`);
- the post-mesh snap is per-axis (`_occ_assembly_2d` / `_occ_assembly_3d`);
- the carve's open-wall tests are `X[:, axis] == value` (victim protection, ring checks, the 2-D splice, the 3-D bowl);
- the 3-D cap is meshed flat in the wall plane's uv coordinates.

Generalising means replacing the pair with a small wall descriptor — implicit function for membership, projection for the snap, chart for the cap — a bounded refactor: one concept, several touch points, each already isolated behind a named test in the 2-D path.

The annulus fault studies are the first customer. Corner/box-edge outcrop (a zone meeting two walls) is a separate, genuinely harder problem and stays refused regardless.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
Progress in #582: the assembly clip now cuts against the mesh's own boundary rebuilt as OCC geometry (2-D and 3-D), the outcrop band is identified as the trace on any boundary and labelled `<label>_trace`, and the 2-D carve/sew is fully general — a radial ribbon out of the annulus's outer boundary embeds with area conserved to 1e-12 and the P2 oracle exact through the zone, at np=1/2/4 with partition-independent counts.

Remaining for this issue: the general 3-D cap. A zone outcropping on a non-axis-aligned 3-D boundary is currently refused with the reason (`TODO(DESIGN)` at `_trace_wall_code`). The construction we intend: re-triangulate the bowl-minus-band collar per coplanar boundary region, in each region's own plane so the cap stays exactly on the faceted surface, with 1-D overlays where the band outline crosses creases. That is a spike-validated piece of work of its own.

Issue #553, opened 2026-08-13
fault_split: allow a slit that crosses the boundary transversally (the daylighting contact)


lmoresi:
Split out of #549, which the ribbon representation resolves (PR #552). The contact (slit) representation still refuses any fault touching the boundary:

```
fault_split: the fault touches the domain boundary. Only strictly interior
             faults, with both tips ...
```

The refusal's argument is about **tips** — the slit would reach the boundary, where the tip-clamping argument fails. That argument is right for a tip and over-broad for a crossing:

- **A tip lands on the boundary** — keep refusing. The clamping argument genuinely fails there.
- **The slit crosses the boundary transversally** — a daylighting fault's boundary end is not a tip. It is an edge where the boundary condition differs above and below the slit, which is exactly how Barr & Houseman (GRL 1992; GJI 1996) drive slip: velocity +u0/2 above the fault on the boundary, -u0/2 below.

The substantive work is the boundary labelling: the two sides of the slit at the wall must be **separately addressable** so a solver can set different conditions across it. That is a design decision about label identity (does the wall label split? does each side get a derived name, as fault_split already mints per-fault boundary values?), not carve plumbing.

The ribbon path went first because it was bounded and unblocks the science; this issue is the second half of the original ask, and the Barr & Houseman boundary-driven configuration is its acceptance test.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
## The distinction, located in the code

The refusal in `_fault_chain` walks **every** vertex of the fault chain and refuses if any has a domain-boundary facet in its star. It makes no distinction between the two cases, which is why it is over-broad rather than wrong.

The tip-clamping it protects is not in that check at all — it is in `_take_sides`:

```python
for i in range(1, len(chain) - 1):
```

**The endpoints are skipped.** A tip vertex is therefore never duplicated, both sides keep sharing it, and the slip is forced to zero there. That skip IS the clamping argument, expressed as three characters of a range.

So the three cases separate cleanly:

| case | what it is | what to do |
|---|---|---|
| endpoint strictly interior | a real tip | skip it — clamp, as today |
| **endpoint on the boundary** | **the fault leaves the domain** | **split it — no clamping** |
| interior chain vertex on the boundary | the chain runs ALONG the wall | keep refusing; degenerate |

## What makes it work, and what makes it awkward

[...]

PR #582, opened 2026-08-16
Clip fault assemblies against the mesh's own boundary; label the outcrop trace

lmoresi:
Addresses the curved-boundary side of #555 and builds the trace primitive agreed on #553.

## What this does

The thin-volume assembly clip (`place_thin_volume`, 2-D and 3-D) no longer intersects with an analytic axis-aligned box. It intersects with the domain the mesh actually has: the boundary facets (support 1 and unshared) are gathered collectively, stitched by exact coordinate identity, and rebuilt as OCC geometry — collinear runs and coplanar triangles compressed to their corners, so a box constructs the same four-sided / six-faced tool `addRectangle`/`addBox` did, and the boolean does not imprint the wall's mesh spacing onto the clipped faces. The cut lands on the mesh's own facets (measured 3.5e-17 from the chords, against the 8.6e-3 sagitta a cut on the smooth circle would leave), so no corefinement library is needed and the clip-then-mesh order is preserved: the ribbon still meshes at gmsh quality.

The outcrop band is now identified as the **trace** — membership of the domain's own boundary facets, defined on any boundary — and the band facets carry `<label>_trace` alongside the skin and wall labels, in both dimensions. The trace is the intersection itself; it never partitions the wall into named pieces, so an annulus (closed wall) and a 3-D footprint need no naming convention. `info["n_trace_facets"]` reports the count.

The 2-D carve/sew is fully general:

- the ring splice detects the wall span topologically (boundary-edge pairs) and orders coverage and chain orientation by arc projection onto the removed span's own polyline;
- boundary-vertex deletability: a vertex goes when the band re-provides its geometry, or when it lies on a band-touched straight segment where the splice's cap stays collinear; a compressed corner off the band is the domain's shape itself and is protected. On a box this reproduces the wall-plane rule exactly;
- a clipped corner landing within `0.1*size` of an imprinted domain vertex is merged into it, along the boundary segment — measured: a 0.18-degree sliver on the annulus radial outcrop restored to 29.4 degrees against the 29.6-degree interior twin, with domain shape and area untouched.

## What this does not do yet

The 3-D outcrop keeps the box frame: a zone outcropping on a **non-axis-aligned 3-D boundary** is refused with the reason (marked `TODO(DESIGN)` at `_trace_wall_code`). The general 3-D cap must re-triangulate the bowl-minus-band collar per coplanar boundary region, with 1-D overlays where the band outline crosses creases — a separate, spike-validated piece of work. The 3-D clip, trace identification and trace labelling in this PR are already general; only the carve/sew cap is not.

`place_sheet`'s discrete Sutherland–Hodgman clip is also untouched, deliberately — unifying it is follow-on work.

## Evidence

- Box regression oracle unchanged: `test_0854`, `test_0855`, `ptest_0854`, `ptest_0855` pass in serial and at np=2/np=4.
- New `test_0859`: the tool's OCC mass equals the mesh's own summed cell measure to round-off in both dimensions, and — the negative control — differs from the smooth-geometry measure by far more, proving the tool is the polygon, not the circle. A cube merges to exactly six faces.
- New annulus outcrop test: interior twin as negative control (no trace), area conserved to 1e-12, every boundary edge relabelled, trace coincides with the band, minimum angle floored at 15 degrees, P2 oracle exact through the zone.
- Parallel: annulus outcrop passes at np=2, 3, 4; counts identical at np=1/2/4 (trace=1, zone=44, skin=24, placed=60, removed=17) — the trace is a function of the mesh, not the partition.
- Full level-1 suite: 1499 passed. Style gates clean.

### DIFF
--- a/src/underworld3/utilities/place_surface.py
+++ b/src/underworld3/utilities/place_surface.py
@@ -4588,26 +4588,29 @@ def _place_thin_volume_2d(dm, polylines, width, label, label_value,
                     "volume must be interior, with clearance to spare")
             victim |= orphan
 
-            # Labels the deleted wall span carried, read before the
-            # rebuild forgets them — the splice segments and the band are
-            # relabelled with exactly these after sewing.
-            wall_pairs = []
+            # Labels the deleted wall span carried, read PER EDGE before
+            # the rebuild forgets them — a span across a domain corner
+            # changes label mid-way (Top on one side, Right on the other),
+            # so each new wall edge must take the labels of the old
+            # segment it lies on, not a set common to the whole span.
+            span_labels = []
+            span_segments = []
             if removed_wall:
-                edge_pts = [int(dm_work.getFullJoin(
-                    [int(a) + vS, int(b) + vS])[0])
-                    for a, b in removed_wall]
-                for i in range(dm_work.getNumLabels()):
-                    name = dm_work.getLabelName(i)
-                    if name in reconnect._TOPOLOGY_LABELS:
-                        continue
-                    lab = dm_work.getLabel(name)
-                    values = lab.getValueIS()
-                    if values is None:
-                        continue
-                    for val in values.getIndices():
-                        if all(lab.getValue(p) == int(val)
-                               for p in edge_pts):
-                            wall_pairs.append((name, int(val)))
+                names = [dm_work.getLabelName(i)
+                         for i in range(dm_work.getNumLabels())
+                         if dm_work.getLabelName(i)
+                         not in reconnect._TOPOLOGY_LABELS]
+                for a, b in removed_wall:
+                    p = int(dm_work.getFullJoin([int(a) + vS,
+                                                 int(b) + vS])[0])
+                    pairs_p = []
+                    for name in names:
+                        val = dm_work.getLabel(name).getValue(p)
+                        if val >= 0:
+                            pairs_p.append((name, int(val)))
+                    span_labels.append(pairs_p)
+                    span_segments.append((X[int(a)].copy(),
+                                          X[int(b)].copy()))
 
             Xall = np.vstack([X, asm_pts])
             holes = [[len(X) + int(v) for v in loop] for loop in hole_loops]
@@ -4634,7 +4637,7 @@ def _place_thin_volume_2d(dm, polylines, width, label, label_value,
                 # The splice ends: the surviving corner vertices and the
                 # chain ends they meet — the chain is the ring's tail, in
                 # the orientation the splice chose.
-                outcrop = (wall_pairs,
+                outcrop = (span_labels, span_segments,
                            int(removed_wall[0][0]),
                            int(removed_wall[-1][1]),
                            int(ring[-len(chain_asm)]) - len(X),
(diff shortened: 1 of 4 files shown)


lmoresi:
Adversarial review (self, pre-merge):

**Found and fixed during the work — reporting the negative results:**

1. The first annulus outcrop meshed a 0.18-degree sliver. Probe: the two offending nodes were an imprinted domain vertex (bit-exact match with the boundary loop) and the ribbon's clipped corner 7.3e-5 away on the same chord — the boolean forces both into the layer mesh. Fixed by `_collapse_boundary_imprints` (merge along the boundary segment, after the meshed-vs-CAD gate); the outcrop now meshes at 29.4 degrees against the interior twin's 29.6.
2. Our first parallel census reported 21 "bare" boundary edges at np=2. That was the probe, not the code: a partition-seam facet also has local support 1 on its owner side, and a leaves-only filter counts it as boundary. We made exactly the mistake `_true_wall_vertex_mask`'s docstring warns about; the probe now skips shared facets.

**Checked and held:**

- The snap is a pure normal-component move, so on an axis-aligned wall the off-axis coordinate is set exactly and the tangential coordinate is bit-untouched — box parity is by construction, and the box oracle did not move at any stage.
- `_split_skin_trace` requires every vertex AND the centroid on the boundary, so a skin facet spanning between two boundary contacts through the interior cannot classify as trace.
- Counts are identical at np=1/2/4 because `_domain_boundary_facets` sorts the gathered complex into canonical order before the rank-0 clip.

**Residual risks we have argued but not measured:**

- `_collapse_boundary_imprints` only merges into a node that exactly matches a domain vertex; a crossing near a compressed-away vertex leaves no second node and therefore no sliver — argued from the tool construction, not measured on a box corner case.
- `_occ_domain_3d` is measured only at test scale (a few hundred boundary triangles). A fine spherical shell builds one OCC face per triangle; cost on thousands-of-face tools is unmeasured. The reach restriction from the design note remains available if it bites.
- The trace membership tolerance (1e-9) is metric. The separation to the nearest non-trace facet is the layer mesh size, five orders larger, but it is a tolerance, not a topological identity.

**Deferred, explicitly:** the general 3-D cap (bowl-minus-band collar per coplanar region) — refused with the reason and marked `TODO(DESIGN)`; `place_sheet`'s discrete clip — untouched by design to keep the diff reviewable.

lmoresi:
Review addendum (04cca8ed): probing the question "is every 2-D outcrop case covered?" found one that was neither tested nor refused. The general carve/sew had silently subsumed the box-corner outcrop the old wall-code frame refused — a single ribbon out through the (1,1) corner embeds, area conserved to 1e-12 — but the wall-label harvest was wrong there: labels common to ALL removed span edges are empty across a corner (Top one side, Right the other), leaving four bare wall edges. The harvest is now per removed edge and each new wall edge takes the labels of the segment it lies on; the corner node recovers both walls' labels through the two closures. Pinned by a new test; uniform spans (one wall, the annulus loop) unchanged; two separate bands stay refused.

lmoresi:
## Adversarial review — fault-placement stack (#582, #589)

Reviewed together. Each already carries its own review; this adds the
relationship between them, which neither says.

**S1. #589 is stacked on #582.** `pr582` is an ancestor of `pr589`. Both edit
`src/underworld3/utilities/place_surface.py` and
`tests/parallel/ptest_0855_place_thin_volume_parallel.py`; #589 adds
`test_0854_place_sheet.py` and `test_0860_outcrop_general_boundary_3d.py` on top
of #582's `test_0855_place_thin_volume.py` and `test_0859_domain_boundary_tool.py`.

Consequences for the backlog:

- Merging #589 merges #582. Merging #582 first leaves #589 as the general 3-D
  cap alone, which is how the work was scoped (#555's remaining piece).
- The reviews are not independent evidence. Anything approved on #582's
  `place_surface.py` is being approved again as part of #589, and a finding
  raised on #582 and fixed on #589 will look unaddressed on #582.
- Whichever merges second needs no rebase, but #582 will show as fully merged
  rather than closing on its own commits.

Recommended order: #582, then #589.

**S2. Both touch `ptest_0855`, which is a `ptest_` file.** Those are not
collected by a default `pytest tests/` run — they are driven under `mpirun`
explicitly. Two PRs editing the same manually-driven parallel test is the case
where a regression is least likely to be noticed by CI. Worth confirming the
combined file runs at np=2 and np=4 after both land, rather than after each.

lmoresi:
Merging this first, ahead of #589, per S1 in the cluster review above: #589 is stacked on this branch and merging #589 would have merged this one as part of it. Taking them in order keeps #582's commits as #582's, and leaves #589 as the general 3-D cap on its own — which is how the work was scoped against #555.

#589 is not merged with it: its CI is currently failing.

Issue #544, opened 2026-08-13
What should the rheology be where two fault zones overlap at a high angle?


lmoresi:
**Rescoped 2026-08-16** after review. This was originally filed as three defects; two of them do not survive scrutiny and are withdrawn below. What remains is one open modelling question.

## The question

Where two fault zones overlap, the fused zone carries a single label by design — that ignorance is what makes `fuse` (#541) remove junction detection, and it is deliberate. So the rheology at an overlap comes entirely from the `Surface` lookup, which today gives each node the director of the **nearest** surface.

At a low-angle crossing or a tangential merge the two directors nearly agree, so the choice does not matter. At a **high-angle** crossing the overlap is weak in two directions, and a single director represents only one of them.

The open question is whether that matters in practice. The ownership boundary is a measure-zero surface with cells either side taking opposite directors, so the **net** response of the overlap region may already approximate the orthotropic answer without anyone constructing it. That is plausible and unmeasured.

## What would settle it

A **high-angle** crossing — the tangential geometries used so far cannot discriminate, since both directors agree there by construction. Compare, on one mesh:

- nearest-fault director (current behaviour)
- an explicit orthotropic treatment, e.g. summing the two weak planes' **compliance** tensors and inverting — which represents both weaknesses, is continuous, and degenerates correctly where only one surface is near

against the gmsh-union control. If they agree, the current rule is fine and this issue closes. If not, the gap is the error being carried today.

Note the connection to **#539**: an additive-compliance treatment needs the TI compliance tensor and the `eta -> 1/(4 eta)` structure that issue records. The two are related if this route is taken.

## Withdrawn from the original report

- **"Nearest vertex, not nearest surface."** Real as stated — a densely-sampled trace annexes territory from a sparsely-sampled one — but this is a **modelling-discipline requirement**, not a defect: a fault set that builds one model should be consistently sampled. Generalising the KDTree to exact distance-to-segment would make a coarse triangulation work well enough to keep, which is the wrong service. Refine or retriangulate the surface instead, which fixes the distance query and everything else that reads it.
- **"The director flips discontinuously across the partition boundary."** Poor wording on my part: "partition" here meant the geometric medial axis between the two traces, not the MPI decomposition. Nothing is decomposition-dependent — the KDTree is built from replicated `Surface` vertices and queried with rank-local coordinates, so a given point gets the same answer at any rank count. The parallel contract is intact.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

[...]

Issue #539, opened 2026-08-13
TI: expose the compliance (inverse) tensor — same structure, reciprocal scale


lmoresi:
## Request

Expose the inverse of the transversely isotropic viscosity tensor (the compliance) alongside `flux`, so strain rate can be inferred from a recovered stress rather than differentiated from velocity.

## Why it is nearly free

The Mühlhaus–Moresi tensor decomposes as

    C = 2 eta_0 I  -  2 (eta_0 - eta_1) P

with `P` the projector onto the on-plane shear modes (the bracketed n-terms in `TransverseIsotropicFlowModel`). `C` therefore has two eigenvalues — `2 eta_1` on that subspace, `2 eta_0` off it — and the inverse is the same decomposition with those reciprocated:

    C^-1 = (1 / 2 eta_0) I  +  (1 / 2 eta_1  -  1 / 2 eta_0) P

Same structure, different scale. So `_build_c_tensor` can build the compliance unchanged under the substitution

    eta_0 -> 1 / (4 eta_0)
    eta_1 -> 1 / (4 eta_1)

(the quarter absorbing the factor-of-2 convention in both directions).

## Why it is wanted

1. **Plasticity**: the return map is derived in compliance form and implemented in stiffness form; the inverse is what maps between them.
2. **Diagnostics**: traction (`sigma . n`) is continuous across a fault-zone boundary while the strain rate is not, so the sound route is to recover stress by projection and infer strain rate through the compliance. Differentiating the velocity instead gives a field that jumps at the zone edge and recovers badly. This came up while building stress diagnostics for a TI fault zone (`~/+Simulations/listric_extension/fuse_stress_p1.py`); the TI case is the one where it has not been done.

## Check before trusting it

The substitution assumes the bracketed term is exactly `2P` with `P` idempotent in the implemented form. Build `C` and `C^-1` numerically for several directors and assert `C : C^-1 == I` to round-off. That either confirms the substitution or exposes a convention factor — much cheaper to settle now than inside a return map.

Underworld development team with AI support from Claude Code

lmoresi:
**The check this issue asks for, done: the substitution is exact.** No convention factor, no director dependence.

Built `C` from the implemented `TransverseIsotropicFlowModel`, built the candidate compliance by re-running the same builder under `eta_0 -> 1/(4 eta_0)`, `eta_1 -> 1/(4 eta_1)`, and contracted:

```
[...]

PR #583, opened 2026-08-16
Teaching examples on the ribbon representation: crossing fault zones, and gouge width

lmoresi:
Two pages in `docs/advanced/`, built on the finite-width (ribbon) fault representation. Each tests a prediction rather than illustrating one. Scripts and full working notes live in `~/+Simulations/ribbon_network_2d/`.

## `crossing-fault-zones.md` — the measurement #544 asked for

**The algebra comes first, and it reframes the question.** In plane strain the deviatoric strain-rate space is two-dimensional, and rotating the material by θ rotates a state in it by 2θ. A weak plane is compliant in exactly one mode, so it occupies a single *line* in that space; two planes separated by Δ in the rock are separated by 2Δ there.

So Δ = 0 and Δ = 90 are the same case — shear on a plane with normal x̂ *is* shear on a plane with normal ŷ. The 90-degree X that "high angle" suggests is the one geometry where the two rules provably coincide, and the acceptance test as #544 proposed it would have measured nothing.

Checked numerically against the shipped Mühlhaus–Moresi tensor. Two by-products: #539's `eta -> 1/(4 eta)` substitution is confirmed to round-off, and the orthotropic tensor is itself a TI tensor in 2-D, so it needs no new constitutive machinery — read off an effective director and effective η₀, η₁.

**The measurement.** Two ribbons crossing at the box centre, fused into one zone, five rheologies per angle on one mesh, differing *only* in the overlap cells, against a no-fault control on that same mesh. Metric is dissipation weakening, which has no probe geometry in it.

| Δ | 10° | 20° | 30° | 45° | 60° | 75° | 90° |
|---|---|---|---|---|---|---|---|
| nearest / orthotropic | 70.0% | 81.6% | 86.2% | 93.0% | 97.1% | 99.2% | **100.0%** |

The Δ = 90 row is the experiment's negative control and it passes exactly — the two dissipations agree to seven significant figures. The algebra requires it, nothing was tuned to produce it, and it is what licenses the other six rows.

**The error is largest where the issue expected none.** The per-cell tensors differ most at 45°, and the curve shows no feature there. The overlap *area* goes as 1/sin Δ — 141 overlap cells at 10° against 24 at 90° — so the area wins. The geometry to worry about is the shallow crossing.

**Resolution.** The host mesh does not enter at all: 0.04, 0.03 and 0.02 give the same ratio to four significant figures, because the embedded zone mesh is identical in all three. The zone mesh does, and refining it always moves the nearest-fault rule closer to the orthotropic one — 70.0 → 73.8 → 75.1% at Δ = 10, converging near 76%. The sweep table is therefore a set of bounds, and the corrected headline is that the rule loses about a quarter of the network's weakening at the worst angle, not a third.

## `gouge-zones.md` — what the matched collapse loses

The collapse `eta_I = eta_1 / w` is checked rather than assumed: it reproduces slip to 1.7% at w = 0.002 and 12% at w = 0.02, behaving as the w → 0 limit it is.

It preserves the *ratio* η₁/w and destroys the pair. So the gouge's own temperature excess,

    Delta T = eta_1 V^2 / (8 kappa) = (eta_I V^2 / 8 kappa) * w

is proportional to the width at fixed interface law and is zero for a contact at every width. Matched to 1% at three widths with nothing fitted — η₁ and κ are set, V and ΔT are measured independently.

Separately and more sharply: **a transport equation posed on a split-node mesh sees a perfect insulator.** Plain conduction, uniform diffusivity, no fault properties anywhere, no mechanics — the split mesh loses 19% of the wall flux and develops a temperature discontinuity of 0.47 across the fault. The plain-mesh control reproduces the analytic flux exactly (−0.010000), so that is the mesh and not the solver.

## Things recorded so they are not re-derived

- `clearance` is **not monotone**: a 30° crossing builds only at 1.3, a 45° one at 1.0 but not 1.3. Walk a ladder and record which value was used.
- A fixed probe standoff is wrong at a shallow crossing — a probe at 0.75 w from one trace sits inside the *other* ribbon and read 0.006 against a peak of 0.18. Even walked clear of the union, the local probe cannot be trusted at the crossing itself, which is why the metric is a dissipation.
- The rotated contact path builds its own prefixed KSP that `stokes.petsc_options` never reaches; on its default the velocity sub-solve hit its 200-iteration cap and warned while the outer SNES still reported convergence. Use `_rotated_use_lu` at this size, and take the verdict from the constraint (the no-opening leak, 5.5e-17 here) rather than from a SNES reason that does not describe that solver.
- Do not take a dissipation profile through `uw.function.evaluate` — it L2-projects derivative composites (#491) and the smoothing lands on the structure being measured. Project to discontinuous degree 0, where the L2 projection is the cell average.

## Not included

Example 3 of the session plan (a large-scale single surface, where the contact should win outright) is blocked on #553 — the contact cannot daylight, and daylighting is the use case.


lmoresi:
## Adversarial review

Reviewed wide — the diff, the two sweeps behind it, #544, #539, #553, and the split-node and fault-network pages this one links to. Findings first, then what we could not fault.

### Fixed in 519743ec

**The central algebraic result is two-dimensional and the page did not say so.** The 90-degree null exists because the deviatoric space here has two dimensions, so two lines in it either coincide or do not. In 3-D that space has five dimensions, two weak planes at any angle occupy different subspaces, and there is no null. A reader taking "a 90-degree crossing is safe" into a 3-D model would be wrong, and the page as written invited exactly that. It now refuses the reading explicitly.

**We measured a distance to a choice and reported it as a distance to an answer.** There is no external reference for what an overlap should be, and there cannot be one without modelling the fabric — both orthotropic constructions are assumptions. The page said "neither is obviously right" in passing and then quoted percentages as if one were. What actually carries the argument is that the two constructions differ by a **factor of two** in overlap weakness and still bracket the answer tightly (61.8% against 70.0% at the worst angle; 92.3% against 93.0% at 45 degrees), so the conclusion is insensitive to the assumed part. That inference was available in the tables and left for the reader to make; it is now stated.

**The headline is a small difference of larger numbers.** At Δ = 10 the dissipations are 3.9521 and 3.9225 against a control of 4.0000 — the entire signal is about 0.03 between numbers agreeing to 1%. Now said next to the metric, with what makes it readable: the control hits the analytic 4.000000 exactly, and the refinement study moves the ratio in a controlled, contracting sequence.

### Standing limitations, not fixed

**One point along the fault carries the gouge oracle.** ΔT is measured at the fault midpoint only. The 1% agreement at three widths is strong, but nothing checks that it holds toward the tips where slip and therefore Φ fall away. The claim is about the peak, which is the quantity of interest, so this is scope rather than error — but it is untested scope.

**The gouge oracle's input is a probe.** ΔT = η₁V²/8κ takes V from a velocity-jump probe with the ribbon's standoff convention, and V enters squared. The 1% agreement is therefore also a check on that probe, not only on the physics. It cannot be disentangled from the data we have.

**The sweep confounds crossing angle with fault orientation** — unavoidable geometrically, since the pair is symmetric about the loading axes and two planes cannot change their mutual angle while both keep their orientation. Stated on the page; worth repeating that no absolute slip should be compared across angles.

### What survived

- The Δ = 90 null (3.243286 against 3.243286, seven significant figures) is required by the algebra, was predicted before it was run, and nothing in the setup was tuned toward it. It is the strongest thing here.
- The host-mesh independence (three sizes, same ratio to four figures) is a genuine check on the embedding contract rather than a convenience.
- The no-fault control landing on the analytic 4.000000 validates the dissipation integral independently.
- On the gouge side, the plain-mesh conduction control reproducing −0.010000 exactly is what makes the split-mesh insulation result attributable to the mesh and not the solver. Without it the 19% flux deficit would be an assertion.

### Related, unblocked by this PR

#544 has its measurement and a recommendation; it can close on a modelling decision rather than more work. #539 is confirmed incidentally — the `eta -> 1/(4 eta)` substitution reproduces the compliance to round-off, and the director round-trips exactly mod 90 degrees. #553 still gates the third teaching example; nothing here moves it.

lmoresi:
The scope of this PR grew with the live sessions it documents; a summary of what it now contains, in reading order:

1. **`crossing-fault-zones.md`** — overlap rheologies for fused zones in pure extension: the 2-D deviatoric algebra (90° crossings are degenerate; the discriminating angle is 45°), the measured sweep with its exact Δ=90° null, the resolution study making every quoted percentage a bound, and a postscript stating plainly that this test's symmetric, rotation-free frame flattered the constrained models.
2. **`gouge-zones.md`** — what the collapse to a surface preserves (slip, to its w→0 limit) and destroys (the interior: gouge ΔT = η₁V²/8κ matched to 1% with nothing fitted; a split-node mesh as a perfect insulator for any second equation).
3. **`fault-branching-junctions.md`** (this commit) — the shear-box study that decides: the junction as a slip exchanger the crack constraint cannot run, the exit-throat law and the corner singularity it implies, the matrix-stress pattern flip, the damage experiment with negative control, and the choosing rules that follow from "two of these are crack models and one is not".

The instruments changed under review as defects were found and fixed: all strain-rate work is P0 cell projections (local, validated to 7e-16 on a linear field); τ-P1 recovery is retained only where it is sound (isotropic media, traction diagnostics) with its failure across director interfaces documented; band structure (union-jack vs ladder) is verified per build, never assumed from the size parameter.

Issue #579, opened 2026-08-16
GAMG defaults deviate from PETSc's and cripple it: pc_mg_type=additive, and block size 1 on vector fields


lmoresi:
## Two deviations from PETSc's defaults, both costly

Underworld sets `pc_mg_type = "additive"` in every GAMG bundle
(`petsc_generic_snes_solvers.pyx` lines 3260, 4173, 5189, and 6073 for the
Stokes velocity block). **PETSc's default for GAMG is `multiplicative`.**
Additive multigrid is a strictly weaker cycle — the levels do not see each
other's corrections.

And the velocity operator reaches GAMG with **block size 1** on a
two-component field. GAMG then aggregates each scalar degree of freedom
independently rather than aggregating nodes, which is the standard way to
cripple algebraic multigrid on a vector problem.

## Measured

SolKz, unit box, constant viscosity, Taylor-Hood P2-P1, 11 727 unknowns,
refinement 2 — nothing hard is happening here:

| GAMG configuration | cycles per velocity solve | flops | wall |
|---|---|---|---|
| as Underworld configures it | 243–347 | 1.155e10 | 1.38 s |
| `pc_mg_type=multiplicative` | 74–118 | 6.892e9 | 0.90 s |
| block size 2 | 84–120 | 4.599e9 | 0.70 s |
| **both corrected** | **34–47** | **3.121e9** | **0.54 s** |

Seven times the cycles and 3.7 times the arithmetic, for two settings.

At larger sizes the uncorrected configuration is worse still: at 186 879
unknowns it needs up to 4 136 cycles per velocity solve, which overruns any
reasonable `ksp_max_it` and then silently reports a truncated solve.

## Why it matters beyond performance

GAMG is what every solver falls back to when the mesh carries no refinement
hierarchy, which is the common case. So this is the default path for most
users, and it is slow for reasons that have nothing to do with the method.

It also makes any geometric-versus-algebraic comparison meaningless: a
measurement against this configuration is measuring our own settings.

## Suggested fix

Remove `pc_mg_type = "additive"` and let PETSc choose. Deviating from a PETSc
default wants a recorded reason, and there is no comment giving one here.

Set the block size on the operator for vector fields so GAMG aggregates nodes.
The matrix is built by PETSc from the DM, so the natural place is wherever the
section is set up rather than after the fact.

Worth checking the other three call sites for the same reasoning, and whether
[...]

Issue #576, opened 2026-08-15
Stokes outer Krylov defaults to gmres while the fieldsplit sub-blocks are Krylov solves — must be fgmres


lmoresi:
## The default

On a plain `uw.systems.Stokes`, the solver configuration is:

| | type | preconditioner |
|---|---|---|
| outer | **`gmres`** | `fieldsplit` |
| velocity block | `fgmres` | `mg` |
| pressure block | `fgmres` | `gasm` |

The sub-blocks are flexible; the outer one is not.

## Why that is the wrong way round

The velocity sub-solve is not a passive preconditioner — inside the Schur
factorisation it **computes the search direction**. It is a Krylov method,
therefore inexact, therefore the operator effectively applied **differs between
outer iterations**. Plain GMRES assumes a fixed preconditioner; its residual
recurrence has no guarantee against an operator that varies. Flexibility is the
half of the design that tolerates inexact inner solves, and the inner tolerance
margins are the half that bounds them — they are two parts of one decision.

## Measured cost

On the notch problem at refinement 3, varying only the outer method:

| outer | inner rtol | velocity its/step | result |
|---|---|---|---|
| `gmres` | 3.3e-08 | 983 | DIVERGED_LINEAR_SOLVE |
| `gmres` | 3.3e-08, cap 2000 | 999 | DIVERGED_LINEAR_SOLVE (residual identical to the above) |
| `gmres` | 1e-03 | 915 | DIVERGED_LINEAR_SOLVE |
| **`fgmres`** | 3.3e-08 | **58** | no linear failure |
| **`fgmres`** | 1e-03 | **23** | no linear failure |

Seventeen times less work, and forty-three with a loosened inner tolerance.
Raising the velocity iteration cap is **inert** — byte-identical rows — so the
failure is inconsistency, not insufficient iterations. Loosening the inner
tolerance by five orders while the outer stayed non-flexible was worth 7%.

## Why it goes unnoticed

Easy problems hide it. On isoviscous SolKz at refinement 2 the outer converges
in 2 iterations and nothing looks wrong. The default only bites where the inner
solves are genuinely inexact — hard nonlinear rheologies, strong viscosity
structure — which is where a user is least able to tell a solver defect from a
modelling problem.

## Suggested fix

Default `ksp_type` to `fgmres` for the Stokes outer solve. The extra storage
over GMRES is one vector per Krylov direction, which is the normal price of
flexibility and is small against the cost of the block solves it wraps.

[...]

PR #584, opened 2026-08-16
Preconditioner defaults: PETSc's GAMG cycle, node aggregation, and a flexible outer Krylov

lmoresi:
Two preconditioner defaults that deviate from PETSc's without a recorded reason, and one consequence of the first.

## The GAMG cycle (#579)

`pc_mg_type` was `additive`. PETSc's default for GAMG is `multiplicative`, and additive is a strictly weaker cycle — the levels do not see each other's corrections.

Provenance, from `git log -S`: it arrived in December 2022 in example scripts, alongside `pc_mg_levels = 2`. With a single coarse level the two cycles very nearly coincide, so it cost nothing where it was written. It was promoted into the solver defaults, the hierarchies got deeper, and nobody revisited it. No comment ever gave a reason.

Measured on SolKz, Taylor–Hood P2–P1, 11 727 unknowns, cycles per velocity solve: **243–347 additive against 74–118 multiplicative**.

## The four open-coded copies

`utilities/multigrid_options.py` says of itself that "nobody writes a multigrid option value anywhere else", and that is the reason the module exists — hand-maintained copies are what let the smoother iteration count drift in #468. But four call sites in `petsc_generic_snes_solvers.pyx` open-coded the GAMG bundle key by key, which is how one value came to be wrong in four places at once.

Those four now read the bundle. The module's claim about itself is true again.

## The outer Krylov (#576)

The Stokes outer `ksp_type` was never set, so it inherited PETSc's `gmres`, while both fieldsplit sub-blocks are `fgmres` run to a tolerance.

That is the wrong way round. Inside the Schur factorisation the velocity sub-solve computes the search direction, so the operator the outer method applies differs between outer iterations, and GMRES's residual recurrence assumes it does not. Measured on the Spiegelman notch at refinement 3: 983 velocity iterations per step and `DIVERGED_LINEAR_SOLVE`, against 58 for `fgmres` and 23 with a loosened inner tolerance. Raising the velocity iteration cap is inert — byte-identical residuals — so the failure is inconsistency, not a shortage of iterations.

Set explicitly in `__init__` and in the `strategy` setter, which also resets the fieldsplit block.

## Node aggregation

The velocity operator reached GAMG with **block size 1** on a two-component field, so it aggregated scalars rather than nodes.

This is not an unconditional defect. PETSc infers the block size correctly wherever it can:

| BCs on velocity | velocity IS | fieldsplit sub-matrix |
|---|---|---|
| none | 2 | 2 |
| full vector Dirichlet | 2 | 2 |
| **component-wise (free slip)** | **1** | **1** |

It collapses to 1 under a component-wise Dirichlet condition because the field then genuinely carries one degree of freedom at a constrained wall node and two inside — block size 1 is the correct description. Component-wise Dirichlet is free slip, so the one case PETSc cannot infer is the standard geodynamics boundary condition.

Setting `mat_block_size` asks GAMG to aggregate on a pairing that does not align with nodes at the walls. That cannot change the answer — a preconditioner alters the route to the solution, not the solution — and measured on free-slip SolKz the two configurations agree to **2e-7 relative** on a 1e-6 solve. What it changes is the cost:

| free-slip SolKz, 11 727 unknowns | cycles/velocity solve | flops |
|---|---|---|
| block size 1 | 74–118 | 3.48e9 |
| block size 2 | 34–47 | 1.60e9 |

With the same contrast concentrated in a band rather than spread smoothly, a plain `stokes.solve()` goes from **2.61e11 flops / 32.7 s** to **4.44e9 / 1.8 s**.

PETSc honours `mat_block_size` as an option, so this needs no hook and no post-hoc PC surgery — only that it is set before the matrix is built. It is recorded on the solver as `_pc_block_size`, not just at the `__init__` call sites, because `_apply_preconditioner_options` re-applies the bundle on every build and a bundle built without it lists `mat_block_size` as a *stale* key and deletes what `__init__` set.

`SNES_MultiComponent` keeps block size 1: its unknown is not `mesh.dim` components in general, and claiming the wrong node size would cost rather than save.

## One test expectation changed

`test_0203_solver_wallclock_guard::test_first_solve_admits_its_count_is_a_lower_bound` asserted that a fresh solver's first velocity count *strictly* undercounts the second. The contract is that the first count is a **lower bound** and is flagged as one; how far short it falls depends on where `PCSetUp` lands relative to the first application of the block, which these defaults move. It is now tight on that problem (80 either way, still flagged incomplete), so the test asserts the bound rather than the size of the gap. Verified to pass on `development` before the change and to fail only because of it.

## Still unaudited

`pc_gamg_agg_nsmooths=2` (PETSc: 1), `pc_gamg_repartition=True` (PETSc: False), and the geometric bundle's `pc_mg_type="full"` (PETSc PCMG: `multiplicative`) remain deviations without recorded measurements.

## Verification

`./uw test`: **1493 passed**, 32 skipped, 2 xfailed. All three defaults confirmed live after `./uw build`, including that `strategy=` no longer reverts the outer.


lmoresi:

is semantic.

### Cluster

|---|---|---|---|

that branch or an equivalent.

### #584

open:

  an oversight.
  than to this follow-up batch.

### #548

lmoresi:
## The block-size failures, fixed, and the branch brought current

Pushed `8ed2558`. This branch was 57 commits behind `development`; it now merges
cleanly and carries the fix for both failure modes.

**The four PETSc error 75 failures.** `mat_block_size` is declared as a PETSc
*option*, so when `PCFieldSplit` extracts the velocity sub-matrix,
`MatSetFromOptions` hands it to `PetscLayoutSetBlockSize` — a hard error rather
than a hint it may decline:

    Arguments are incompatible
    Local size 67 not compatible with block size 2      (2-D)
    Local size 1384 not compatible with block size 3    (3-D)

The commentary in `_gamg_settings` anticipated the misalignment under
component-wise Dirichlet but not the refusal. The reason the SolKz measurements
never met it is narrower than "component-wise Dirichlet": divisibility is a
property of the particular COMBINATION of boundary conditions, not of their
kind. On a 3x3 P2 velocity field —

| velocity BCs | local size | |
|---|---|---|
| none | 98 | even |
| full vector Dirichlet, every wall | 50 | even |
| component-wise free slip | 70 | even |
| `(0, 0)` Bottom, `(0, None)` elsewhere | **67** | odd, refused |

Free slip is safe by an accident of counting, and it is the case the
measurements used. `test_1010_stokesCart`'s asymmetric mix is not.

`_withdraw_block_size_if_not_node_blocked` removes the option when the block is
not node-blocked, immediately before `setFromOptions` — the first point the
field decomposition exists — and on every build, since
`_apply_preconditioner_options` re-pushes the bundle each time.

**The decision is collective**, because divisibility is mixed across ranks. On a
6x6 box the per-rank velocity sizes are `[127, 150]` at np=2 and
`[72, 94, 111]` at np=3, so a rank-local test would leave ranks disagreeing
about the contents of the options DB.

**The fifth failure was a test defect.**
`test_1021_mg_option_bundle::test_bundles_clear_each_others_keys` built `owned`
from a bundle list that never included `gamg_bundle(block_size=2)` — the one
bundle that sets `mat_block_size` — so every bundle's stale list carried a key
`owned` lacked. The sibling is now in the list, which also brings that bundle
under the test.

**`test_1022_velocity_block_size_gate.py`** pins both directions. The second
case is the one that matters: a gate that withdrew the option unconditionally
would fix the crash and silently undo everything #579 was for, and the crash
test alone cannot tell them apart.

**Two things for the PR text.** The measured 74-118 to 34-47 cycle win is from
free-slip SolKz, which divides; the asymmetric mixes now run without node
aggregation, so they get the multiplicative cycle and the flexible outer but not
that factor. And `SNES_Vector` has the same exposure — the gate covers it, but

Verified: full `./uw test` 1556 passed, 32 skipped, 2 xfailed, against
`development` at `fe6b2e1`.

Issue #587, opened 2026-08-16
Eighteen test modules build state at import: retire the collection-guard ratchet list


lmoresi:
`tests/conftest.py` now fingerprints the process state around each test module's
import and fails the run when a module moves it (#575). Eighteen modules already
do, so they are exempted by name in `_KNOWN_COLLECTION_TIME_WORK` — a ratchet
that stops the practice spreading, not an approval of it. This issue is the list.

## Why module-level work is not merely untidy

pytest imports a module in order to collect it, so anything at module level runs
before any test, any fixture, and any isolation `conftest.py` provides. Two
defects reached `development` that way:

- **#567** — a module switched the units system on at import, and the first
  module-scoped fixture in that worker built its mesh under dimensional
  coordinates. Sampled 2.9 million times too far apart; it looked like a
  parallelism bug and blocked CI parallelism for days.
- **#505** — a module ran two Stokes solves at import, so `--collect-only` sat
  inside `SNESSolve` looking like a silent death.

A module that builds a mesh at import has not failed yet; it has built its mesh
under whatever global state the previously collected module left behind.

## The list

Fix first, because it is the only one that moves the **units** state
(`uw.use_strict_units(False)` at line 28, before three meshes and fourteen
variables — the #567 mechanism exactly):

- [ ] `tests/test_0601_mesh_vector_calc.py`

The rest build objects at import:

- [ ] `tests/parallel/test_0765_internal_boundary_integral_mpi.py`
- [ ] `tests/test_0004_pointwise_fns.py`
- [ ] `tests/test_0005_IndexSwarmVariable.py`
- [ ] `tests/test_0501_integrals.py`
- [ ] `tests/test_0502_boundary_integrals.py`
- [ ] `tests/test_0504_projections.py`
- [ ] `tests/test_0810_amr_swarm_migration_regression.py`
- [ ] `tests/test_0830_mesh_adapt_variable_transfer.py`
- [ ] `tests/test_1000_poissonCart.py`
- [ ] `tests/test_1000_poissonNaturalBC.py`
- [ ] `tests/test_1001_poissonSph.py`
- [ ] `tests/test_1004_DarcyCartesian.py`
- [ ] `tests/test_1010_stokesCart.py`
- [ ] `tests/test_1011_stokesSph.py`
- [ ] `tests/test_1014_stokes_multigrid.py`
- [ ] `tests/test_1014_stokes_shell_nullspace.py`
[...]

Issue #570, opened 2026-08-15
Audit scripts/test.sh against the test tree: four coverage gaps in two days, each of which hid a real defect


lmoresi:
CI does not run all the tests, and the gaps are not random — every one found in the last two days was hiding a defect that the missing test would have caught.

## The four

1. **The tier-A gate does not cover the solver batches.** `pytest -m "level_1 and tier_a"` was green while `test_1005_TransientDarcyCartesian` failed on CI (PR #557) — a `TypeError` from a 0-d ndarray reaching a sympy comparison. Local gate green is not CI green, and we treated it as if it were.
2. **`test_1018_*` matched no batch glob** — the entire rotated free-slip suite, including the #497 regression guard. Fixed as #504 (PR #532), which also found **56 orphaned files** and an inline comment that swallowed a batch's `|| status=1` so its failures could never fail CI.
3. **`tests/parallel/test_10*py` was commented out** in `scripts/test.sh`, so `test_1017` and `test_1062`–`test_1068` had executed at *no rank count* since they were written. Enabled in PR #561; it immediately surfaced seven partition-independence failures (#564), one of them a genuine 3.4% defect.
4. **`test_106*` is deliberately unbatched** — and that is where the serial default-normal consumers live. It is the one suite that would have caught the #568 review's B1: `mesh.boundary_normal()` returning normals wrong by ≈5.1° on a 3-D spherical shell **in serial**.

## The ask

A one-off audit of `scripts/test.sh` against the actual tree, and a mechanism so it cannot drift again:

- enumerate every `tests/**/test_*.py` and every `tests/parallel/**` file, and map it to the batch that runs it (or to an explicit, commented decision not to);
- for each file that is deliberately excluded, state the reason in the script itself — runtime, known-slow, requires a resource — so an exclusion is a decision rather than an accident;
- add a check (a test, or a step in the workflow) that FAILS when a test file matches no batch and carries no exclusion marker. Without that, this recurs — three of the four above were introduced by someone adding files and not noticing the globs;
- audit the batch lines for swallowed exit codes (the #504 instance) so a failing batch cannot pass silently.

## Why this is worth doing properly

Each of the four gaps cost a defect reaching development or surviving there — a silent-wrong-flux class, a 3.4% partition dependence, a 5° normal error on spherical shells, and a crash. The tests to catch all four already existed. They just never ran.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

Issue #567, opened 2026-08-15
Three point-locator tests fail when xdist worker count changes the file grouping


lmoresi:
Found while measuring parallel test runs (#566).

## What happens

Running level 1 under `pytest-xdist --dist loadfile` at **16 workers** on a 16-core box:

```
FAILED tests/test_0761_point_locator.py::test_p1_field_on_a_shared_edge_matches_the_closed_form[2d-0.0]
FAILED tests/test_0761_point_locator.py::test_p1_field_on_a_shared_edge_matches_the_closed_form[2d-0.5]
FAILED tests/test_0761_point_locator.py::test_p1_field_on_a_shared_edge_matches_the_closed_form[2d-0.25]
```

The same selection is **green at 4 and at 8 workers**, and green serially.

## It is not load, and not telemetry

- **Deterministic**: identical three failures across two runs.
- **Not telemetry**: `UW_ENABLE_TELEMETRY=0` (which `test_levels.sh` sets specifically to avoid a kdtree race) makes no difference — same three failures.
- The failure is an accuracy assertion in `uw.function.evaluate` at points on cell edges, not a crash or a resource error.

With `--dist loadfile`, the worker count determines **which files share a process**. So the most likely reading is test pollution: some other file landing in the same worker leaves state that changes what the locator returns. The test itself is explicit about what it is discriminating — "a locator that returns any cell CONTAINING the point gives the exact answer, and a locator that returns a cell merely NEAR it does not" — so whatever the leaked state is, it is plausibly steering cell selection.

## Why it matters beyond the test

If global state can change which cell a point-location query resolves to, that is worth understanding even for serial users: it means an evaluation's accuracy depends on what the process did earlier. It may equally turn out to be a fixture or a cached mesh in the test file. Either answer is worth having.

## Reproducing

```bash
OMP_NUM_THREADS=1 OPENBLAS_NUM_THREADS=1 MKL_NUM_THREADS=1 \
  pytest --config-file=tests/pytest.ini tests/ \
  -m "level_1 and not level_2 and not level_3" -q -n 16 --dist loadfile
```

`--isolation` (one worker, still per-file) is the tool for bisecting which preceding file is responsible.

Related: the locator suite is recent (#556, part of the #551 umbrella), and these tests had no prior parallel exposure.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
## Now reproducing at FOUR workers, in CI, and it is total rather than marginal

[...]

PR #588, opened 2026-08-16
Test modules must not build state while being collected

### LINKED ISSUE #505: test_0050_utils.py runs two Stokes solves at module scope — pytest COLLECTION executes them and whole-directory runs hang silently
Found during the #500 review gate. `tests/test_0050_utils.py` performs two full Stokes solves at module import time, so any pytest invocation that collects the directory (`pytest -m "level_1 and tier_a"`, `pytest --collect-only`, an IDE test explorer) executes them during collection. We sampled a collect-only process 20+ minutes deep inside `SNESSolve → PCApply_FieldSplit_Schur` at 99.7% CPU — to the caller it looks like pytest died silently (cf. the import-death class in the CI notes). `scripts/test.sh` already comments the file out ("disable auditor test for now"), which hides rather than fixes it.

Fix: move the solves into fixtures/test bodies (or guard with `pytest.importorskip`-style laziness). Until then every whole-directory pytest needs `--ignore=tests/test_0050_utils.py` to survive.

### LINKED ISSUE #575: Nothing prevents a test module from mutating global state at import — add a collection-time guard
Found via #567 (PR #574). `tests/test_0741_expression_arithmetic_units.py` reached `development` as a **converted script with no test functions at all**, whose entire body ran at module level — including `uw.Model().set_reference_quantities(length=2900 km, time=1 Myr)`, which switches the units system on **process-wide, during pytest collection**.

The consequence was subtle and expensive: a function-scoped reset cannot protect a module-scoped fixture (pytest builds higher-scoped fixtures first), so any test file whose module-scoped fixture ran *first in its worker* built its mesh under dimensional coordinates. `test_0761_point_locator`'s field was sampled 2.9 million times too far apart — worst error 1.6747e+00 against a clean 2.9e-15. It only showed up under `pytest-xdist` because `--dist loadfile` decides which file is first in a process, which is why it looked like a parallelism bug and blocked enabling CI parallelism for days.

PR #574 fixes the instance (module-scoped autouse reset, and 0741 rewritten as real tests) and pins the *consequence* with `test_0742`. **It does not pin the practice.**

## The ask

A collection-time guard that fails when a test module mutates global state at import. Candidates, cheapest first:

- a `conftest.py` hook that snapshots the relevant globals (the default `uw.Model`, the units/scaling registry, any process-wide PETSc options UW3 sets) before and after collection, and fails with the offending module named;
- a check that every `tests/test_*.py` contains at least one test function — a file with none is either dead or a script, and both are worth knowing about (this alone would have caught 0741);
- the same snapshot applied per-module during collection, which localises the culprit rather than reporting a global diff.

Related instances of the same family, so this is not a one-off: `tests/test_0050_utils.py` runs **two Stokes solves at module scope** (#505), which is why whole-directory collection hangs and why `pytest -m "level_1 and tier_a"` without a path dies silently. Both files predate any of today's work.

Related: #567/#574 (the instance), #505 (the sibling), #570 (the coverage audit — a file with no tests is invisible to it too).

lmoresi:
Closes #505. Closes #575. Closes one of the gaps in #570.

pytest imports a test module in order to collect it, so anything a module does at
module level runs before any test, any fixture, and any isolation
`tests/conftest.py` provides. Two defects have reached `development` that way:
#567, where a module switched the units system on at import and the first
module-scoped fixture in that worker built its mesh under dimensional
coordinates; and #505, where a module ran two Stokes solves at import, so
`--collect-only` sat inside `SNESSolve` looking like a silent death.

PR #574 fixed the first instance. This fixes the practice.

## The three files that had no tests in them

`test_0050_utils.py`, `test_0120_data_property_access.py` and
`test_0130_field_creation.py` were converted scripts. Between them they built two
meshes, a swarm, two Stokes solvers, and solved both — all at module level — and
contained no test function at all. 0120 and 0130 wrapped every statement in a
`try/except` that printed the exception, so a broken `.data` property or a
duplicate `field_id` printed a cross and the run stayed green. 0050's only
assertion was disabled by the name `dont_test_auditor`.

Each is now written as fixtures and tests, and each keeps its original subject:

- **0050** — the installation auditor, and a second Stokes solver built over the
  fields the first already owns. The auditor check is a delta rather than the
  `uw_object_count == 7` it used to assert: the counter is process-wide and
  monotonic, so an absolute count is only true in a fresh process running that
  file alone, which is why the assertion could not survive being enabled. Its
  control is that reading the auditor does not itself move the count.
- **0120** — a write through a swarm variable's `.data` is read back and compared
  with what was written, twice, which is what exercises the cache.
- **0130** — three variables on one mesh get three distinct `field_id`s, and
  `.array` is shaped by the variable's own component count.

Importing `test_0050_utils.py` cost 2.0 s of the 5.2 s import with warm JIT and
mesh caches; the 20+ minutes recorded on #505 was that work on a cold cache.

## The guard

`tests/conftest.py` fingerprints three pieces of process-global state —
the count of `uw_object`s created, the active model's reference quantities, and
the strict-units flag — before and after each module is collected, and reports
the difference as a **collection error against that module**. The fingerprint is
skipped when underworld3 is not importable, since conftest is loaded before the
package is necessarily installed in CI.

A collection error rather than a session abort, because the first form aborted
from `pytest_collection_finish` and that leaves an xdist worker part-collected:
the controller reports `INTERNALERROR> assert not crashitem` at return code 3,
which is the mode CI runs in. `pytest.exit(returncode=4)` behaves the same. The
check now lives in a `pytest_make_collect_report` wrapper, which sees the report
before it is dispatched. Measured on a module that builds a mesh at import:

| | offending module | clean tree |
|---|---|---|
| serial | RC=2, `ERROR tests/test_offender.py` | RC=0 |
| `-n 2` | RC=1, `10 passed, 1 error` | RC=0 |

`tests/test_0051_collection_state_guard.py` runs pytest on a module that offends
and on one that does not, loading the hooks out of the real `conftest.py` by
path. The second run is the control: without it the first proves only that
something failed.

`UW_TEST_COLLECTION_GUARD=off` turns the guard off. `test_0742` needs it: that
test copies the live conftest into a pytester sub-run together with a module
that leaks units at import, because what it pins is that the module-scoped reset
survives exactly that. The guard refused the module and ended the sub-run before
it reached its assertion — correct behaviour on a deliberately generated
offender. `test_0742` now sets the variable for its sub-run only, and
`test_0051` covers both settings so the off switch cannot become the default
without a test failing.

## The ratchet

Eighteen modules already move that state, so they are exempted by name in
`_KNOWN_COLLECTION_TIME_WORK`. The guard therefore stops the practice spreading
today rather than after eighteen files have been rewritten. The list is #587,
with `test_0601_mesh_vector_calc.py` named first because it is the only entry
that moves the units state — `use_strict_units(False)` at import, which is the
#567 mechanism itself, reaching every module collected after it.

`scripts/test.sh` disables `tests/test_06*py` wholesale, which is why 0601's
import-time flip has never been noticed by our own CI: it is collected by any
whole-directory run and executed by none of ours.

## scripts/test.sh

`tests/test_0050*py` is no longer commented out. It was disabled because the
auditor assertion could not pass, which the delta form fixes.

## Verified

- `pytest tests/ --collect-only` exits 0 with 2741 tests collected; before, it
  reported eighteen modules and, with the three files above, ran two Stokes
  solves.
- The guard was confirmed to fire by adding a module-level mesh to a scratch
  file: it named that file and no other.
- Full `./uw test`: 1504 passed, 32 skipped, 2 xfailed.

The fingerprint is not a general leak detector. It sees objects built and the
units state; a module that writes an environment variable, a PETSc option, or a
monkeypatched attribute at import passes it. The rule it enforces is the
stricter and simpler one — build nothing while being collected.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -83,7 +83,7 @@ if [ $PARALLEL_ONLY -eq 0 ]; then
 
   # Run simple tests (0000-0299: basic functionality, imports, simple operations)
   $PYTEST tests/test_00[0-4]*py || status=1
-  #$PYTEST tests/test_0050*py    || status=1 # disable auditor test for now
+  $PYTEST tests/test_0050*py || status=1
   $PYTEST tests/test_005[1-9]*py tests/test_006[0-1]*py || status=1
   $PYTEST tests/test_01*py || status=1
   $PYTEST tests/test_02*py || status=1
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -47,6 +47,129 @@ if _xdist_worker:
     os.environ.setdefault("UW_MESH_CACHE_DIR", f".meshes/{_xdist_worker}")
 
 
+# ==============================================================================
+# COLLECTION-TIME GLOBAL STATE GUARD (#575)
+# ==============================================================================
+# pytest imports a test module in order to collect it, so anything a module does
+# at import time runs BEFORE any test, any fixture and any isolation the
+# fixtures below provide. Two defects have reached `development` that way:
+#
+#   #567  a module switched the units system on process-wide at import, and the
+#         first module-scoped fixture in that worker built its mesh under
+#         dimensional coordinates;
+#   #505  a module ran two Stokes solves at import, so `--collect-only` sat
+#         inside SNESSolve for 20+ minutes looking like a silent death.
+#
+# The fixtures below fix the consequence. This guard fixes the practice: it
+# fingerprints the process state around each module's import and fails the run,
+# naming the module and what moved, when a module writes to it.
+#
+# The check is skipped when underworld3 is not importable — conftest.py is
+# loaded before the package is necessarily installed in CI.
+
+
+def _global_state_fingerprint():
+    """Process-global state that importing a test module must leave alone."""
+
+    try:
+        import underworld3 as uw
+        from underworld3 import model as _model
+        from underworld3.utilities._api_tools import uw_object
+    except ImportError:
+        return None
+
+    active_model = _model._default_model
+    reference_quantities = ()
+    if active_model is not None:
+        reference_quantities = tuple(
+            sorted(getattr(active_model, "_reference_quantities", None) or {})
+        )
+
+    return {
+        "uw objects created": uw_object.uw_object_counter(),
+        "units reference quantities": reference_quantities,
+        "strict units": uw.is_strict_units_active(),
+    }
+
+
+# Modules that already do this, measured on `development` at the time the guard
+# was written. They are exempted so the guard can be turned on today; the list
+# is a ratchet, not an approval — nothing may be added to it, and each entry is
+# a module whose module-level work belongs in a fixture (#587).
+#
+# `test_0601_mesh_vector_calc.py` is the one to fix first: alone in this list it
+# moves the units state (`use_strict_units(False)` at import), which is the #567
+# mechanism itself and reaches every module collected after it.
+_KNOWN_COLLECTION_TIME_WORK = (
+    "parallel/test_0765_internal_boundary_integral_mpi.py",
+    "test_0004_pointwise_fns.py",
+    "test_0005_IndexSwarmVariable.py",
+    "test_0501_integrals.py",
+    "test_0502_boundary_integrals.py",
+    "test_0504_projections.py",
+    "test_0601_mesh_vector_calc.py",
+    "test_0810_amr_swarm_migration_regression.py",
+    "test_0830_mesh_adapt_variable_transfer.py",
+    "test_1000_poissonCart.py",
+    "test_1000_poissonNaturalBC.py",
+    "test_1001_poissonSph.py",
+    "test_1004_DarcyCartesian.py",
+    "test_1010_stokesCart.py",
+    "test_1011_stokesSph.py",
+    "test_1014_stokes_multigrid.py",
+    "test_1014_stokes_shell_nullspace.py",
+    "test_1050_VEstokesCart.py",
+)
+
+
+def _is_known_offender(nodeid):
+    path = nodeid.replace(os.sep, "/")
+    return any(path.endswith(known) for known in _KNOWN_COLLECTION_TIME_WORK)
+
+
+_state_before_collect = {}
+_collection_offenders = []
+
+
+def pytest_collectstart(collector):
+    if isinstance(collector, pytest.Module):
+        _state_before_collect[collector.nodeid] = _global_state_fingerprint()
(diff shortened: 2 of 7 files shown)


lmoresi:
## Adversarial review

Reviewed against the branch head (`18e3833`). Four findings, none blocking; the
fifth item is a property we checked rather than a finding.

**1. The ratchet matches by suffix, so it exempts more than the eighteen files.**
`_is_known_offender` compares `nodeid.endswith(known)`. Any path ending in a
listed basename is exempt, in any directory — a future
`tests/parallel/test_1010_stokesCart.py` inherits the exemption without anyone
deciding it should. The reason it is a suffix match is that `pytest.ini` lives
in `tests/`, so a run under `scripts/test.sh` produces nodeids without the
`tests/` prefix while a run from the repo root produces them with it, and an
exact match would have to know which. We are accepting this because the list is
closed and only ever shrinks (#587), but it is a hole while it exists.

**2. The auditor delta could pass for the wrong reason.** `after - before >= 2`
is a lower bound over a mesh and a variable, and mesh construction alone may
account for both. If `MeshVariable` stopped registering as a `uw_object` the
assertion would very likely still hold. We chose the bound over an exact count
deliberately — an exact count is what made the old `== 7` unrunnable — and the
control that reading the auditor does not itself move the count is what stops
the test being vacuous. It is a weaker test than the file used to intend.

**3. `test_second_solver_over_the_same_fields` does not check either answer.**
It asserts both fields are finite, the first is non-zero, and the two differ.
A regression that made the second solve wrong in a way that differed from the
first would pass. That is the scope we want here — Stokes accuracy is
`test_1010`'s subject, and this file's is that a second solver over the same
`MeshVariable`s is not served the first one's setup — but the assertion is
weaker than it reads.

**4. The off switch is process-wide and unqualified.** `UW_TEST_COLLECTION_GUARD=off`
exported in a shell disables the guard for every run from that shell, silently.
Nothing prints when the guard is off. `test_0742` needs the switch and sets it
with `monkeypatch` for its sub-run only, which is the right scope, but the
mechanism does not enforce that scope on anyone else.

**5. Checked: the fingerprint does nothing collective.** It reads
`uw_object._obj_count`, `model._default_model` and `_STRICT_UNITS_MODE` — all
rank-local. This matters because the fingerprint runs twice per module inside a
collection hook, and the auditor's own `get_runtime_data` does an `MPI.bcast`;
calling that here would have put a collective in the collection path of every
`mpirun` test job. `get_runtime_data` is used only inside `test_0050`'s own
test body.

PR #589, opened 2026-08-16
The general 3-D outcrop cap: a zone may meet any boundary

lmoresi:
Stacked on #582 (base is its branch); only the last three commits are this PR.

A fault zone specified past the mesh boundary now embeds whatever that boundary is: a tilted or rotated wall, a box edge, a faceted sphere. The `NotImplementedError` at the old `_trace_wall_code` — "the zone outcrops on a non-axis-aligned boundary" — is gone, and the function with it.

## What the cap is now

The cap over the outcrop bowl is re-triangulated per **coplanar boundary region**, each piece meshed flat in its region's own plane, so the cap sits exactly on the faceted surface and volume conservation survives to 1e-12. Where the band crosses a crease between regions, the two sides segment the crease line differently — the bowl's nodes are mesh vertices, the band's crossing nodes are assembly nodes — so a 1-D overlay merges both node sets by arc coordinate and keeps each elementary sub-segment for the side(s) whose bowl covers it and the band does not. A sub-segment covered from one side only must be a whole mesh edge, because a cavity-shell face matches it in the fill's closed surface; anything else is a refusal, not a crack.

The frame rule (`_outcrop_frame_3d`) separates two notions the third dimension forces apart. Where the cavity may **open** is by smooth wall: coplanar regions grouped across low-dihedral creases (under 45 degrees), so a faceted sphere is one wall and its bowl legitimately spills onto facets beside the band's own, while a box's other walls stay refused across their sharp edges. What the carve may **delete** is by shape: band-covered vertices, single-region interiors, and interior vertices of straight creases between touched regions. A vertex where three or more regions meet — a box corner, every vertex of a faceted sphere — is the domain's shape and is never deleted off the band. A protected vertex whose whole cell star drops is not stranded: it is a collar node, and the gap fill's tets reference it. That is how a curved boundary keeps its faceting through the surgery.

Wall labels are restored per removed face (the 2-D per-segment rule of the corner fix, one level up), so a bowl spanning Top and Front restores each wall's own labels. The Euler gate now demands conservation of the input's Euler number rather than the value 1: a spherical shell is S²×I, Euler 2, and was refused before for its topology rather than for any defect. The same assumption remains in `place_sheet` and `remove_embedded`, marked `TODO(BUG)`.

## Evidence

- The box oracle did not move: the full placement suite (0853–0856, 0859) passes unchanged, 62 tests.
- New `test_0860_outcrop_general_boundary_3d`: a rotated box (no wall axis-aligned), a band across the Top/Front box edge (trace on both walls, each wall's labels restored on its side), and a spherical shell outer surface. Each runs its negative control first — the identical census on an interior twin counts no trace — and each ends in the P2 Poisson oracle, exact through the zone.
- Parallel: the rotated-box outcrop added to `ptest_0855` passes at np=2 and np=4; the trace count (28 facets) is identical at np=1, 2 and 4 with the info dict identical on every rank.

## Known limits, stated as refusals

- A band running **along** a domain crease (rather than across it) refuses with the reason.
- A band outline node landing within rounding of a kept mesh vertex refuses: the 2-D imprint collapse (`_collapse_boundary_imprints`) has no 3-D counterpart yet.
- `place_sheet`'s outcrop stays box-framed, per the plan's own advice — unify after, not during.

### DIFF
--- a/src/underworld3/utilities/place_surface.py
+++ b/src/underworld3/utilities/place_surface.py
@@ -5711,27 +5711,30 @@ def place_thin_volume(dm, patches, width, label=ZONE_LABEL, label_value=1,
         old_pts = np.vstack([tri for tri, _p in removed_wall])
         old_tris = np.arange(3 * len(removed_wall)).reshape(-1, 3)
         d_old, at_old = _nearest_facet(centres, old_pts, old_tris)
+        # A raise here is rank-local — a hang at np>=2 — so every
+        # refusal in this block goes through the collective failure.
         if (d_old > 1e-9).any():
-            raise RuntimeError(
-                "an outcrop wall triangle lies on no removed wall face; "
-                "the collar and the bowl disagree")
-        # The cap's collar first, then the band — the TRACE, the
-        # intersection itself, labelled as such so the model can form
-        # whatever unions it needs (never a partition of the wall into
-        # named pieces).
-        out_trace = new.getLabel(trace_label)
-        for k, ids in enumerate(wall_tris):
-            joined = new.getFullJoin(ids)
-            if len(joined) != 1:
-                failure = ("an outcrop wall triangle is not a face of the "
-                           "sewn mesh; the cap or band was not sewn.")
-                break
-            for q in new.getTransitiveClosure(int(joined[0]))[0]:
-                for name, val in removed_wall[int(at_old[k])][1]:
-                    new.getLabel(name).setValue(int(q), int(val))
-                if k >= n_cap_tris:
-                    out_trace.setValue(int(q), int(label_value))
-            n_wall_local += 1
+            failure = ("an outcrop wall triangle lies on no removed wall "
+                       "face; the collar and the bowl disagree")
+        else:
+            # The cap's collar first, then the band — the TRACE, the
+            # intersection itself, labelled as such so the model can form
+            # whatever unions it needs (never a partition of the wall
+            # into named pieces).
+            out_trace = new.getLabel(trace_label)
+            for k, ids in enumerate(wall_tris):
+                joined = new.getFullJoin(ids)
+                if len(joined) != 1:
+                    failure = ("an outcrop wall triangle is not a face of "
+                               "the sewn mesh; the cap or band was not "
+                               "sewn.")
+                    break
+                for q in new.getTransitiveClosure(int(joined[0]))[0]:
+                    for name, val in removed_wall[int(at_old[k])][1]:
+                        new.getLabel(name).setValue(int(q), int(val))
+                    if k >= n_cap_tris:
+                        out_trace.setValue(int(q), int(label_value))
+                n_wall_local += 1
     failures = comm.allgather(failure)
     real = [f for f in failures if f]
     if real:
--- a/src/underworld3/utilities/place_surface.py
+++ b/src/underworld3/utilities/place_surface.py
@@ -3141,12 +3141,29 @@ def _occ_assembly_3d(patches, width, size, domain=None, assembly="fuse"):
                                          dtype=np.int64).reshape(-1, 4))
         if not tets:
             raise RuntimeError("the assembly meshed to no tetrahedra")
+        tets = np.vstack(tets)
+        # The meshed-vs-CAD gate runs BEFORE the snap (the 2-D collapse
+        # precedent): it tests gmsh's fidelity to the boolean, while the
+        # snap is our own exact bookkeeping whose volume effect is gated
+        # by the caller's domain-conservation check. The reference, OCC's
+        # boolean mass, is itself only good to ~5e-7 relative on a
+        # many-faceted tool, hence the tolerance.
+        P = xyz[tets]
+        v6 = np.einsum("ij,ij->i",
+                       np.cross(P[:, 1] - P[:, 0], P[:, 2] - P[:, 0]),
+                       P[:, 3] - P[:, 0])
+        mesh_vol = float(np.abs(v6).sum() / 6.0)
+        if abs(mesh_vol - cad_volume) > 1e-6 * cad_volume:
+            raise RuntimeError(
+                f"the assembly meshed to volume {mesh_vol:.12e} against "
+                f"CAD {cad_volume:.12e}; the layer mesh does not fill its "
+                "own solids.")
         if planes is not None:
             # OCC's clipped faces sit within rounding of the boundary; the
             # band logic and the cap's node sharing need EXACT membership,
             # so snap defensively.
             xyz = _snap_to_boundary_3d(xyz, dom_verts, dom_tris, planes)
-        return xyz, np.vstack(tets), float(cad_volume)
+        return xyz, tets, float(cad_volume)
     finally:
         gmsh.finalize()
 
@@ -3240,6 +3257,103 @@ def _collapse_boundary_imprints(asm_pts, asm_tris, loops, delta):
     compact = np.full(len(asm_pts), -1, dtype=np.int64)
     compact[used] = np.arange(len(used))
     return asm_pts[used], compact[tris]
+
+
+def _collapse_boundary_imprints_3d(asm_pts, asm_tets, dom_verts, dom_tris,
+                                   delta):
+    """Route the band outline THROUGH any boundary vertex it grazes.
+
+    The 2-D collapse one dimension up. The clip lands the band's outline
+    where geometry puts it, and over a long band it passes arbitrarily
+    close to some of the wall's own vertices; the collar then meshes a
+    sliver between the vertex and the outline (measured: a 2.4 m gap on
+    a 1000 km megathrust, refused by the 2-D fill as moved nodes). Where
+    a kept-able boundary vertex lies within ``delta`` of the outline,
+    the outline is made to pass THROUGH it: the nearest outline node
+    moves onto the vertex when one is close enough, otherwise the
+    outline edge is split at the vertex — every tetrahedron on that edge
+    bisects, shape-safe because the new node sits within ``delta`` of
+    the edge it splits. The vertex is then band-covered exactly, the
+    frame rule deletes it, and the outline node re-provides its
+    position; the band still tiles the same faceted surface, so the
+    domain's shape is untouched. The band's side faces tilt by under
+    ``delta`` across their last cell — the end-snap contract, one level
+    up.
+    """
+    from collections import Counter
+
+    skin_xyz, skin_local, node_ids = _assembly_skin(asm_pts, asm_tets)
+    _interior, band_idx = _split_skin_trace(skin_xyz, skin_local,
+                                            dom_verts, dom_tris)
+    if not len(band_idx):
+        return asm_pts, asm_tets
+    cnt = Counter()
+    for t in skin_local[band_idx]:
+        a, b, c = sorted(int(v) for v in t)
+        for e in ((a, b), (a, c), (b, c)):
+            cnt[e] += 1
+    outline = [(int(node_ids[a]), int(node_ids[b]))
+               for (a, b), k in cnt.items() if k == 1]
+
+    pts = list(asm_pts)
+    tets = [list(t) for t in asm_tets]
+    onode_xyz = asm_pts[np.unique(np.asarray(outline).ravel())]
+    lo = onode_xyz.min(axis=0) - 2.0 * delta
+    hi = onode_xyz.max(axis=0) + 2.0 * delta
+    candidates = [int(v) for v in np.unique(dom_tris)
+                  if (lo <= dom_verts[v]).all()
+                  and (dom_verts[v] <= hi).all()]
+
+    def v6_of(row, at=None, coord=None):
+        P = [coord if at is not None and x == at else np.asarray(pts[x])
+             for x in row]
+        return float(np.cross(P[1] - P[0], P[2] - P[0]) @ (P[3] - P[0]))
+
+    for v in candidates:
+        V = dom_verts[v]
+        best_d, best_k = np.inf, -1
+        for k, (a, b) in enumerate(outline):
+            A, B = np.asarray(pts[a]), np.asarray(pts[b])
+            e = B - A
+            u = float(np.clip((V - A) @ e / float(e @ e), 0.0, 1.0))
+            d = float(np.linalg.norm(V - (A + u * e)))
+            if d < best_d:
+                best_d, best_k = d, k
+        if best_d >= delta or best_d < 1e-12:
+            continue
+        a, b = outline[best_k]
+        da = float(np.linalg.norm(np.asarray(pts[a]) - V))
+        db = float(np.linalg.norm(np.asarray(pts[b]) - V))
+        # A move or split may not flatten any incident tetrahedron — two
+        # grazes competing for the same neighbourhood would otherwise
+        # stack nodes and degenerate cells. A skipped vertex keeps its
+        # sliver, which is the status quo, not a defect.
+        if min(da, db) < 1.5 * delta:
+            n = a if da <= db else b
+            star = [row for row in tets if n in row]
+            if all(abs(v6_of(row, at=n, coord=V))
+                   > 0.2 * abs(v6_of(row)) for row in star):
+                pts[n] = V.copy()
+            continue
+        edge_rows = [row for row in tets if a in row and b in row]
+        m = len(pts)
+        ok = all(min(abs(v6_of([m if x == a else x for x in row],
+                              at=m, coord=V)),
+                     abs(v6_of([m if x == b else x for x in row],
+                              at=m, coord=V)))
+                 > 0.05 * abs(v6_of(row)) for row in edge_rows)
+        if not ok:
+            continue
+        pts.append(V.copy())
+        fresh = []
+        for row in edge_rows:
+            two = [m if x == b else x for x in row]
+            row[:] = [m if x == a else x for x in row]
+            fresh.append(two)
+        tets.extend(fresh)
+        outline[best_k] = (a, m)
+        outline.append((m, b))
+    return np.asarray(pts, dtype=float), np.asarray(tets, dtype=np.int64)
 
 
 def _refuse_multiple_bands(band_facets):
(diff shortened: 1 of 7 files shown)


lmoresi:
## Adversarial review (self)

We reviewed the three commits against the plan's gates and our own claims. Findings, most severe first.

**1. A rank-local raise shipped in the first commit and would hang at np≥2.** The collar-vs-bowl consistency check raised on the surgery rank inside the relabel block, which is outside any try — exactly the defect class the file's own comments warn about. Found by re-reading the diff, not by any test: the parallel tests pass because the check never fires on healthy geometry. Fixed in the third commit by routing through the block's collective failure. A test that forces this refusal at np≥2 does not exist and would need deliberately inconsistent geometry; we did not construct one.

**2. The general outcrops are 2–4× more slivery than the box outcrop.** Minimum dihedral angles, same mesh scale (cellSize 0.12, width 0.05): interior zone 9.5°, axis-aligned box outcrop 13.8°, rotated box 6.4°, box-edge crossing 3.6°, spherical shell 3.1°. The mechanism is the one the 2-D corner work already met: the band outline lands where it lands, and a collar polygon between an outline node and a nearby kept mesh vertex is thin. The 2-D cure (`_collapse_boundary_imprints`) has no 3-D counterpart; until it does, near-coincidence refuses at rounding distance but produces slivers just above it. All fills pass the conformity gates and the P2 oracle is exact — the slivers are a quality cost, not a correctness one.

**3. A dead end we built and removed.** Our first treatment of stranded protected wall vertices kept one cell of the vertex's star ("rescue"). It produced floating kept islands the fill wrapped as bubbles (global Euler defect, measured on the sphere) and was replaced wholesale: a protected vertex needs no kept cell, because the collar re-triangulation references it and the gap fill's tets give it cells. The final code is simpler than what it replaced; the connectivity gate we wrote for the islands went with the rescue that caused them.

**4. The Euler gate was wrong before this PR, for domains nobody could reach.** `!= 1` encodes a ball; a spherical shell is Euler 2. The general clip made shells reachable, so the thin-volume gate now demands conservation of the input's number. The identical assumption remains live in `place_sheet` and `remove_embedded` — a sheet placed inside a spherical shell refuses today for its topology. Marked `TODO(BUG)` at both sites rather than fixed, per scope discipline.

**5. Untested paths we know about.** The polygon-nesting logic in `_outcrop_collar_3d` is exercised only at depth ≤ 1 (outer ring plus hole); depth-2 nesting (an island inside a hole) is written but unreached. The three refusal guards — band along a crease, outline node within rounding of a kept vertex, a band node splitting a one-side crease edge — fire on geometry none of our cases construct; they are read-verified only. The 45° smooth-wall threshold means a curved boundary faceted coarser than ~45° facet-to-facet splits into separate walls and refuses; documented at the constant, not tested.

**6. Partition independence is asserted, not proven, for the sphere.** The np=1/2/4 identity (trace 28) is measured on the rotated box only. The sphere runs serial-only; its parallel behaviour rests on the same gather-first mechanism the box exercises, which is an argument, not a measurement.

lmoresi:
Review finding 4 is resolved in 41eb60ab rather than left as a TODO: `place_sheet` and `remove_embedded` now gate on conservation of the input's Euler number, the same fix `place_thin_volume` carries. A sheet embeds mid-gap in a spherical shell and removes again, pinned by a regression test in test_0854. Every placement primitive now works in non-ball domains.

lmoresi:
## Review addendum: what a 1000 km outcrop band taught us

We drove the general cap at Earth scale — a 1000 km, 4 km-wide megathrust band against a spherical shell, background mesh ~350 km — and it found four defects the 300 km tests could not see. Three are fixed on this branch, one is characterised and open.

**Fixed.** (1) OCC's boolean mass on a many-faceted tool is only accurate to ~5e-7 relative, so the 1e-9 meshed-vs-CAD gate refused correct assemblies; the gate now runs before the snap and allows 1e-6 (807a9129, restructured in db688649). (2) OCC leaves clipped nodes up to ~4e-7 off the tool's own planes; the snap now covers that noise, masked to the boundary facets so infinite-plane projection cannot reach interior nodes (34ea1928). (3) The band outline grazes the wall's own vertices — measured, a 2.4 m gap the collar meshed as a sliver — and `_collapse_boundary_imprints_3d` now reroutes the outline through such vertices, with a volume guard so a move or split can never flatten a cell; on-crease outline edges are reassigned to the far collar rather than refused (db688649). All 31 placement oracles pass throughout.

**Open, characterised.** Refining a spherical boundary projects new facet vertices onto the true sphere; the creases between the resulting nearly-coplanar sub-facets are laterally fuzzy at (placement noise)/sin(theta) ~ 1e-4 — wider than a 4 km band's own feature spacing — and the crease overlay cannot classify crossings against them. A fuzz-aware tolerance was tried and withdrawn: it traded the sliver for misclassified chains on the honest-crease tests. Until the crease representation is rethought, outcrops on adapt-refined spherical boundaries refuse at the fill's own gates; the plain-shell 1000 km case also still trips a skin/cap facet overlap that is not yet diagnosed. Negative results reported as found; the collar refusal now names its pinch (gap and node kinds), which is the diagnostic that separated all four causes.

lmoresi:
## Adversarial review — fault-placement stack (#582, #589)

Reviewed together. Each already carries its own review; this adds the
relationship between them, which neither says.

**S1. #589 is stacked on #582.** `pr582` is an ancestor of `pr589`. Both edit
`src/underworld3/utilities/place_surface.py` and
`tests/parallel/ptest_0855_place_thin_volume_parallel.py`; #589 adds
`test_0854_place_sheet.py` and `test_0860_outcrop_general_boundary_3d.py` on top
of #582's `test_0855_place_thin_volume.py` and `test_0859_domain_boundary_tool.py`.

Consequences for the backlog:

- Merging #589 merges #582. Merging #582 first leaves #589 as the general 3-D
  cap alone, which is how the work was scoped (#555's remaining piece).
- The reviews are not independent evidence. Anything approved on #582's
  `place_surface.py` is being approved again as part of #589, and a finding
  raised on #582 and fixed on #589 will look unaddressed on #582.
- Whichever merges second needs no rebase, but #582 will show as fully merged
  rather than closing on its own commits.

Recommended order: #582, then #589.

**S2. Both touch `ptest_0855`, which is a `ptest_` file.** Those are not
collected by a default `pytest tests/` run — they are driven under `mpirun`
explicitly. Two PRs editing the same manually-driven parallel test is the case
where a regression is least likely to be noticed by CI. Worth confirming the
combined file runs at np=2 and np=4 after both land, rather than after each.

lmoresi:
## The sheet path joins the general machinery (111ae096..24cbd564)

`place_sheet` now clips, carves and caps against the mesh's own boundary — the Sutherland–Hodgman box clip, the wall-code frame and the single-loop rim cap are deleted. The discrete clip cuts the authored triangulation directly (no OCC): per-component outward orientation, signed-distance classification, sequential cuts by the coplanar regions' planes, cut nodes interned by (original-edge, region) identity. The carve takes the frame's two masks like the volume's carve; the collar embeds the trace chain through its pieces instead of cutting a hole, with crossings on the creases via the existing 1-D overlay. The trace chain's edges carry `<label>_trace` through to the wall — the splittable trace the split machinery needs for a daylighting fault (test_0861 holds the kinematic acceptance criterion, gated on the split side's refusal until feature/fault-split-node lands).

Evidence: box oracle unchanged (test_0854 8/8; clip area matches Sutherland–Hodgman bit-for-bit); the spherical sheet outcrop that previously could not run at all embeds with the trace labelled, volume conserved to 1e-12 and the P2 oracle exact (test_0860, 4/4 with interior negative control); trace counts identical at np=1/2/4; full suite 1505 passed.

## Adversarial review (our own)

Defects found and fixed before the push:
- **Coplanar facets tore the sheet.** Cutting per facet plane keyed the two triangles sharing a cut edge by different (coplanar) facets of one wall — a duplicated node on the box top wall. Cuts now key and compute by the coplanar REGION's plane.
- **Sub-segment interpolation tore the sphere.** A side truncated by an earlier plane interpolated its next cut from the sub-segment, its neighbour from the full edge: ~1e-17-apart duplicate pairs. Side cuts now re-derive from the original sheet edge's endpoints.
- **The collar's overlay shadowed the new `chain` parameter** (`for chain in chains.values()`), so the VOLUME path read a stale crease chain as a sheet trace and indexed with tuples — caught by test_0860's box-edge case, renamed `crease_chain`.
- **petsc4py segfault in our own census**: `getStratumIS` on an empty stratum hands back an IS whose `getIndices` SEGVs (PETSc traps it, so it presents as a silent pytest death, exit 59). Guarded by `getStratumSize` first — the same trap class as the Null_Boundary lesson.
- **Pre-existing red test**: 34ea1928 widened the snap tolerance to 1e-6 and left test_0859's untouched-control at 2e-9, inside it. A branch push does not trigger CI, which is how it shipped red. Fixed (24cbd564).

Known limitations, deliberately loud rather than fixed here:
- A crossing that can touch a locally **concave** crease (an inner boundary) refuses — the polyline cut is not built. On adapted spheres with fuzzy creases the refusal may now fire at the clip rather than the fill's gates; same posture as the crease-classification redesign note, earlier trigger.
- Tolerances are absolute (clip 1e-12, membership 1e-9), consistent with the subsystem; at 1000 km-in-metres scale the rounding floor approaches them. Campaign-wide, not new here.
- The bbox near-set is a superset heuristic: an over-keep is caught by the every-node-inside exit gate (refusal); an over-clip cannot happen where the boundary is locally convex; a torn configuration surfaces at the fill's PLC as a loud error. No silent path found.
- The shell sheet outcrop's min cell volume is ~1e-10 — the sliver follow-on already recorded for general outcrops. Sheets have no 3-D graze collapse; the collar's near-coincidence refusal is the guard.

https://claude.ai/code/session_01WDKP73LD3dCFQSNgEXbbw9

lmoresi:
Merging as the second half of the #582 stack (#582 landed as `87232f0`). Its CI was red when we reviewed; it re-ran against a `development` that now contains #582 and is green — the failure was the stack, not the change.

Per S2 in the cluster review, both PRs edit `ptest_0855_place_thin_volume_parallel.py`, which is a `ptest_` file and so is not collected by a default `pytest tests/` run. The combined file is worth a run at np=2 and np=4 now that both halves are in, rather than after each.

lmoresi:
## The daylighting ruling lands (aa79ec55, ae7f9f6e)

Maintainer ruling (2026-08-18): split-node faults do not split through the surface. A fault stops an element or two below, blind, with a damaged region above carrying the deformation to the surface. This supersedes the split-through-the-wall acceptance test_0861 first carried; the body gated on a daylighting split is removed.

**The kinematic acceptance is now validated by running, not gated.** The listric handover rule (a contact tip inside weak material is free; a tip at the weak zone's edge is pinned) turned vertical against a FREE surface: a blind frictionless split fault two cells under the top, driven in antisymmetric dip-slip, solved bare / damage-abutting / damage-enclosing. Measured (probe in `~/+Simulations/blind_fault_surface_expression/`): near-tip slip 0.45 / 0.65 / 0.80 of peak, surface localization 37% / 44% / 50% — strict ordering on both, enclosing the tip worth 1.77x over bare pinning, and the abutting seam (the hybrid-seam defect) measurably worse than the enclosed tip. The overlap margin is the deliberate design parameter.

**`place_sheet(..., setback=d)`** does the deliberate part: the job is still to figure out the intersection so the fault can stop before it hits it — it just never meshes or splits all the way there. The sheet clips against the boundary offset inward by `d`, arrives blind with a strictly interior rim, and `split_along_label_3d` takes it as it stands; the would-be intersection with the true boundary comes back as `info['surface_trace']` — the damage region's locator. Verified end-to-end in test_0861: place with setback, trace located on the true wall, shallowest sheet vertex exactly a setback below it, split succeeds with the Plus side carrying every placed face.

Adversarial notes on our own additions: an all-rim corner face (two side-rim originals + cut nodes) admits no triangulation the split accepts, at any sheet density — fixed by longest-interior-edge midpoint splits. A centroid split was tried first and rejected on measurement: inside a sliver corner face it drove the gap fill to 1e-23 cell volumes and a P2 error of 0.64 on the shell outcrop, caught by the P2 oracle after every topological gate passed — the #520 lesson holding again. The blind-fault probe's margins are modest at unit scale (the far field dominates a 1x1 box); the committed thresholds hold generous slack under the measured values.

https://claude.ai/code/session_01WDKP73LD3dCFQSNgEXbbw9

lmoresi:
## Physics-first blind depth, verified on an adapted shell (de6f5ba3)

Ruling refinement: the blind depth is chosen from the physics (km-scale) and the resolution adapts to meet it — never a coarse mesh dictating the setback. Verified end to end at Earth scale: a 19,768-cell shell adapts to 288,525 cells and the megathrust places blind at 15 km = 2.5 local elements, trace located, all gates green (probe `~/+Simulations/spherical_slab_outcrop/megathrust_blind_adapted.py`).

Two library changes carried it, both replacing an existence test with a measurement:
- The clip's concave refusal now weighs the crease DEPTH against the setback (over-cut bound ≤ depth): a snapped graded sphere's resolution-transition concavity is metres — admissible under a km-scale setback; an inner boundary's is the facet size — refused; setback zero still refuses any concavity (outcrop exactness).
- `surface_trace` is computed by marching triangles (contouring the boundary signed distance over the sheet's own edges): exact chaining by construction on any boundary. We tried direct tri-tri intersection first and it fragmented a 300 km trace into 148 components at crease endpoints — rejected on measurement.

Findings against our own work, recorded for the next user: the carve's wall clearance requires PROPORTIONAL grading (h ≤ 0.5·distance-to-sheet) — a linear ramp left 570 km deep-rim tets spanning to the wall and the carve rightly refused; NVB's volume-proxy marking leaves cell DIAMETERS ~3× coarser than target while the carve reads diameters — `engine="edge_split"` is the one that meets a setback budget; and the adapt metric must be closed-form (it re-evaluates every bisection generation — a triangle-loop distance blew a 7-level adapt past 10 minutes). Min cell volume of the placed adapted-shell mesh is ~1.5e-18 (a ~7 m sliver): placement gates pass; solvability at that sliver level is the standing follow-on.

https://claude.ai/code/session_01WDKP73LD3dCFQSNgEXbbw9

lmoresi:
## The sheet's resolution is a mesh choice (925b34f3)

Ruling: the authored fault triangulation is data; the embedded fault's resolution must match the mesh it cuts. `place_sheet(size=...)` now re-triangulates the clipped sheet in its own plane at the embedding resolution — the counterpart of `place_thin_volume`'s `size`, closing the asymmetry the unification left. Rim corners are kept exactly (collinear runs compress, so a too-fine authored rim also coarsens), gmsh meshes the interior, and the all-rim split-safety pass (now factored as `_split_safe_triangulation`) runs on the result. Planar sheets, blind/interior placements only — an outcropping sheet's trace must stay verbatim on the boundary complex, and a curved surface needs a parametric remesh; both refuse with the reason.

Measured at Earth scale (the blind megathrust on the 288k-cell adapted shell, identical configuration): a 50-triangle sheet authored at 25 km, resampled to the 6 km local mesh size, embeds as 1744 gmsh-quality faces and the worst cell in the placed mesh improves from 1.5e-18 to 6.8e-14 — four and a half orders of magnitude — with the blind rim still exactly on the offset plane and the patch still splitting as it stands. The remaining worst cell sits in the gap fill, not the sheet triangulation; that residue is the standing gradation follow-on. Box workflow test in test_0861 (coarse 3x3 authored sheet -> min face quality > 0.3, split intact); suites 22/22.

https://claude.ai/code/session_01WDKP73LD3dCFQSNgEXbbw9

lmoresi:
Adversarial review, commits 4b8b0cd4..f8183b21 (the Mesh-level 3-D conforming sheet).

**What we tried to break, and what held:**

- *The refactor changes 2-D behaviour.* `_adopt_cut_child` is extracted verbatim from `add_conforming_surface` (one comment generalised); test_0844 passes complete, including the tail-composition and subsample assertions, so the 2-D contract is unchanged by construction and by test.
- *The 3-D tail is decorative.* test_0862 asserts the child's tail is the base's own level objects (coordinates equal to `dm_hierarchy[-2]`), that no level carries the label, that the cut REPLACES the base finest rather than stacking on it, and that a Poisson solve with a per-cell contrast across the sheet runs on the cut child as it stands.
- *Chaining breaks the tail.* Two stacked sheets on a refinement=1 base leave exactly the base L0 in the tail — each same-resolution cut replaced its parent, per the subsample rule. Asserted.
- *The zone was silently empty.* It was — in any 3-D use (#620): `cells_supporting` walked `_cells_on_edge` from FACE points, which lands on nothing. Fixed by height dispatch; the test asserts zone == union of the faces' support cells. This fix predates the branch conceptually (the method was 2-D-only in practice) and is a candidate for extraction to development.
- *The fault_network rewire changes the place route's output.* Label VALUES now come from `_boundaries_with` (sequential) rather than the local max+4 rule; `split_fault` resolves by NAME, and the end-to-end smoke (place → split → MainPlus/MainMinus present) passes. The route's known solve pathology (3850 s vs 125 s embed) predates the handoff and should be re-measured now the cut mesh actually carries a tail.

**What remains open, honestly:**
- The split at the end of the chain still drops the tail (`add_fault`'s documented contract) — carrying geometric MG through duplicated DOFs is future work; the essential-BC warning in `add_conforming_surface` applies to the sheet unchanged.
- test geometry margins: the carve gates are tight on coarse boxes (cellSize 0.25 + refinement=1 refuses; 0.2 places). The test fixtures sit inside the measured window; a platform with a different gmsh could conceivably land outside it. If 0862 flakes on CI, that is where to look.
- Serial-only coverage here; the parallel gate for the underlying chain remains ptest_0852, which does not go through the new wrapper.

Suites: test_0862 6/6; guards 0844/0845/0850/0851/0853/0854 81 passed, 1 pre-existing skip; full level_1 1574 passed.

lmoresi:
Adversarial review, commits 5f07bc32..1afaa789 (the solver-side pair: custom-P re-install guard, fgmres outer default).

**What we tried to break, and what held:**

- *The guard skips a NEEDED re-install.* The verdict comes from the live PC (managed block exists, is PCMG, level count matches), not from the marker alone; a rebuilt SNES, a preconditioner change, or an unreachable fieldsplit all answer False and re-install. The correctness case is measured, not assumed: perturbing the coefficient data and re-solving under the guard converges (mg:4its) and moves the answer — PETSc's re-Galerkin at PCSetUp carries the new operator values.
- *The guard's win was overstated by a benchmarking artifact — and we caught our own error.* A repeat solve with unchanged data warm-starts from the converged solution and is a near-no-op, so the first "60 s -> 5.5 s" claim compared trivia. The honest arena (genuine solve, perturbed operator) still favours the pair of changes: 543.9 -> 284.6 s.
- *fgmres changes answers.* vrms agrees to the converged tolerance across all arms (0.14723x); the level_1 sweep is green (1574 passed).
- *fgmres masks a real diagnostic.* The one test that failed asserted a STRICT first-solve undercount in solver_health — measured to be an artifact of left-preconditioned gmres applying the preconditioner once before its first monitor fire. Right-preconditioned fgmres forms the unpreconditioned residual first, the bound goes tight, and the test now asserts the contract (lower bound + honesty flag) rather than the artifact. The flag itself still reads False on a first solve, which remains honest: the instrumentation cannot know it caught everything.
- *A user's explicit ksp_type is clobbered.* No: pushed via _push_managed_option in both defaults sites (constructor and strategy setter), so a latched user choice wins — the same doctrine as the #584-era option management.

**Open, honestly:** the genuine-solve bottleneck is now the pressure sub-solve sitting AT its 200-iteration gasm cap, silently unconverged (#625) — it bounds what any velocity-block improvement can show (custom-P vs GAMG is 284.6 vs 307.7 s for an 11x iteration advantage). The hybrid cut-crossing transfer (#622) and the place-route sizing (#621) remain open. All measurements: ~/+Simulations/place_route_health/.

PR #593, opened 2026-08-16
Avoid replicated 3D boundary-flux recovery

gthyagi:
## Summary

Fix the MPI memory scaling of 3D boundary-flux recovery without changing its reaction, trace-mass, gauge-removal, or local-return mathematics.

## Problem

The existing 3D path uses `MPI.allgather` for nodal reactions and P1/P2 trace elements. Every rank therefore retains the complete global boundary mesh, assembles the same sparse surface-mass matrix, and runs the same consistent-mass solve. Memory grows with both global boundary size and rank count.

This became the limiting post-processing stage for high-resolution spherical rotated-free-slip models: the Stokes solve completed, but consistent P2 normal-traction recovery exhausted aggregate job memory.

## Change

- gather reactions, trace elements, and each rank requested coordinate keys to rank zero;
- assemble and solve the global 3D boundary-mass system once;
- scatter only the recovered values requested by each rank, preserving local `xs` order;
- preallocate sparse COO arrays instead of appending Python scalar entries;
- broadcast root-side errors before scatter so all MPI ranks fail collectively;
- retain the existing 2D path unchanged.

The behavior of `mass={"auto","lumped","consistent","p1"}`, partial-reaction sum/overwrite semantics, mean removal, and P2 midpoint interpolation is unchanged.

## Validation

- `tests/test_1019_boundary_flux.py -k 3d`: 3 passed.
- Four-rank `tests/parallel/test_1065_boundary_flux_parallel.py -k 3d`: 2 passed on every rank.
- Previous production validation of this exact commit completed the Zhong `cellsize=1/64`, 192-rank post-processing path in 7:29 with 301.77 GB peak memory; the replicated path had exhausted 576 GB.

## Scope

This PR contains only the general `boundary_flux` gather/solve/scatter correction and its `rotated_bc` documentation update. Zhong geoid mathematics and benchmark-facing APIs remain isolated in #591.

### DIFF
--- a/src/underworld3/utilities/boundary_flux.py
+++ b/src/underworld3/utilities/boundary_flux.py
@@ -12,8 +12,10 @@ gathering, the boundary-mass de-smear (lumped / consistent), and the field hand-
 The equation-specific bit — extracting the nodal reaction — is the solver method
 ``_assemble_volume_reaction``, which returns each rank's RAW (per-rank) volume FEM
 residual; the complete reaction at a boundary node shared across a partition cut is then
-assembled here in ``_desmear`` by SUMMING each rank's partial contribution by coordinate
-(the same rock-solid gather used for the boundary mass — no hand-rolled global assembly).
+assembled here in ``_desmear`` by SUMMING each rank's partial contribution by coordinate.
+In 3D, the complete boundary mass is assembled and solved once on rank zero; only the
+recovered values needed by each rank are scattered back. This avoids replicating the
+global P2 surface mesh and sparse solve on every rank.
 
 ``mass="auto"`` (default) uses a diagonal lumped mass where lumping is pointwise
 sound (the 2D P1/P2 line traces and the 3D P1 triangle trace), and the consistent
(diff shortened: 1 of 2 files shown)


lmoresi:

them.

### Cluster

offer.

### #594 — Zhong 2008

momentum-residual row.

two limits now binds.

reach every rank.

inside test functions.

### What we did not do

lmoresi:
Merging. Two findings from the cluster review are carried forward rather than resolved here:

- **The memory bound moves rather than falls** (finding 2). Gathering to rank zero removes the ×nranks replication #408 asks for; rank zero still holds the whole boundary and now solves it alone. That is the right trade at the sizes motivating #408 and not at the size where rank zero is itself the constraint.
- **No test starves a rank of boundary facets** (finding 3). The recovery is collective by construction now, and a rank owning no part of the boundary still has to enter the gather and the scatter. This is the same shape as the defect fixed in #596; worth a parallel test with a boundary that does not reach every rank.

PR #594, opened 2026-08-16
Add Zhong 2008 propagator-matrix analytic response

gthyagi:
## Dependency

This is a stacked PR on top of #591, which adds the generic
`uw.postprocessing.geoid` coefficient API. The Zhong oracle calls those pure
functions rather than carrying a second geoid/self-gravity implementation.

Until #591 merges, this comparison also shows its three commits. After #591
merges into `development`, the remaining diff is the two commits
`95b34701` and `fb9ce95b`.

## What this adds

`uw.analytic.Zhong2008` is a mesh-independent semi-analytical reference for the
spherical-shell delta-load Stokes benchmarks in Zhong et al. (2008).

For one harmonic degree `l`, it:

- propagates the four-component Hager--O'Connell poloidal state in
  `v = log(r)` with an exact matrix exponential in each constant-viscosity
  layer;
- accepts arbitrary piecewise-constant radial viscosity profiles;
- applies the radial delta load as the traction-state jump
  `(0, 0, -rint * load, 0)`;
- solves the two remaining CMB state coefficients from impermeable free slip
  (`radial velocity = tangential traction = 0`) at both shell boundaries;
- returns no-self-gravity topography/geoid, signed characteristic horizontal
  velocity, and horizontal velocity divergence at the surface and CMB;
- returns self-gravity-corrected topography/geoid through the generic API from
  #591.

This is intentionally a numeric response oracle rather than an
`AnalyticSolution(mesh)`: the paper publishes boundary response coefficients,
not pointwise symbolic velocity/pressure fields. It is exported from
`uw.analytic` but is not registered by `available()`, whose contract requires a
mesh-constructible symbolic solution.

## Radial system

With `L = l(l + 1)` and state

```text
u = (y1, y2, r sigma_rr / eta0, r sigma_r_perp / eta0)^T,
```

the constant-layer system is

```text
du/dv = A u

    [ -2          L          0   0     ]
    [ -1          1          0   1/eta ]
A = [ 12 eta     -6 L eta    1   L     ]
    [ -6 eta  2(2L-1) eta   -1  -2     ]
```

and `u(rb) = exp(A log(rb / ra)) u(ra)`.

## API example

```python
reference = uw.analytic.Zhong2008(
    harmonic_degree=2,
    radius_inner=0.55,
    radius_outer=1.0,
    internal_load_radius=0.775,
).response()

reference.surface_characteristic_velocity
reference.self_gravity.surface_topography
```

The layered Table 3 case is selected with:

```python
viscosity_interfaces=(0.971875,)
viscosities=(1.0, 1.0e4)
```

## Published-reference validation

The focused test checks every parenthesized propagator entry in Zhong et al.
(2008) Tables 2 and 3:

| Cases | Depths | Degrees | Quantities per case | Published values |
| --- | --- | --- | ---: | ---: |
| Isoviscous | `0.25d`, `0.5d`, `0.75d` | `2`, `5`, `8`, `15` | 8 | 96 |
| `10^4` viscosity lid | `0.25d`, `0.5d`, `0.75d` | `2`, `5`, `8`, `15` | 8 | 96 |

All 192 values agree within the precision printed in the paper. Additional
tests cover both free-slip boundary states, response linearity in load
amplitude, no-self-gravity recovery, input validation, and public API behavior.

Representative degree-2, mid-mantle responses:

| Case | `s` | `b` | `h_t` | `h_b` | `U_t` | `U_b` |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Isoviscous oracle | `0.499849` | `0.931260` | `0.0448629` | `0.0546149` | `-0.0100641` | `0.0118606` |
| Paper | `0.4998` | `0.9313` | `0.04486` | `0.05461` | `-0.01006` | `0.01186` |
| Layered oracle | `0.610374` | `0.435376` | `0.0578926` | `0.00675446` | `-1.25811e-05` | `0.00555605` |
| Paper | `0.6104` | `0.4354` | `0.05789` | `0.006754` | `-1.258e-05` | `0.005556` |

## Checks

- `python -m pytest tests/test_1029_analytic_zhong2008.py -v`:
  **34 passed**
- analytic contract/optional/oracle group: **66 passed, 15 skipped**
- `./uw test` Level 1: **1476 passed, 49 skipped, 2 xfailed**
- documentation audit still exits on the repository-wide baseline of 138
  undocumented API items; it reports no new missing Zhong API item.

## Review focus

Please check:

1. the four-state matrix and traction scaling against the intended
   Hager--O'Connell convention;
2. the sign and `rint` factor in the delta-load jump;
3. whether a mesh-independent response oracle should remain directly exported
   but outside the symbolic `available()` registry;
4. the constructor representation of radial viscosity interfaces.

## References

- Hager & O'Connell (1981), JGR 86, 4843--4878,
  doi:10.1029/JB086iB06p04843.
- Zhong et al. (2008), G3 9, Q10017, doi:10.1029/2008GC002048.

### DIFF
--- a/docs/api/analytic.md
+++ b/docs/api/analytic.md
@@ -17,6 +17,18 @@ whichever one you use.
    :show-inheritance:
 ```
 
+## Zhong spherical-shell response oracle
+
+`Zhong2008` is a mesh-independent propagator-matrix oracle. It returns the
+boundary response coefficients used by the Zhong et al. (2008) benchmark; it
+does not implement the symbolic mesh-field contract above.
+
+```{eval-rst}
+.. automodule:: underworld3.analytic.zhong2008
+   :members:
+   :show-inheritance:
+```
+
 ## See also
 
 - {doc}`solvers` — the solvers these solutions validate.
--- a/docs/developer/CHANGELOG.md
+++ b/docs/developer/CHANGELOG.md
@@ -273,10 +273,11 @@ component exactly — correct on curved, tilted, and deformed boundaries (#293).
   seam sensitivity is unchanged (#404).
 - Recorded as the preferred free-slip BC in the project guidance (#300);
   conda PETSc floor raised to ≥ 3.25 for FMG/rotation API consistency (#304).
-- `uw.postprocessing.zhong2008_response_from_rotated_stokes(...)` projects the
-  existing rotated-free-slip boundary traction onto the benchmark harmonic and
-  applies one Appendix A operator for geoid and self-gravity response. The pure
-  operator functions also accept precomputed topography coefficients.
+- `uw.postprocessing.geoid` provides generic spherical-shell geoid and
+  self-gravity coefficient functions. Its rotated-Stokes adapter projects the
+  existing boundary traction onto an axisymmetric harmonic; the pure functions
+  also accept coefficients recovered by other methods and an optional internal
+  load.
 
 ### Generalized Geometric Multigrid via Custom Prolongation (July 2026)
 
--- a/docs/developer/CHANGELOG.md
+++ b/docs/developer/CHANGELOG.md
@@ -278,6 +278,11 @@ component exactly — correct on curved, tilted, and deformed boundaries (#293).
   existing boundary traction onto an axisymmetric harmonic; the pure functions
   also accept coefficients recovered by other methods and an optional internal
   load.
+- `uw.analytic.Zhong2008` implements the Hager--O'Connell propagator-matrix
+  oracle used for the Zhong et al. spherical-shell response benchmark. It
+  supports piecewise-constant radial viscosity and reproduces every analytical
+  response printed in Zhong Tables 2 and 3; geoid and self-gravity are delegated
+  to the generic postprocessing functions above.
 
 ### Generalized Geometric Multigrid via Custom Prolongation (July 2026)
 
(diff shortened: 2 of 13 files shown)


lmoresi:

them.

### Cluster

offer.

### #594 — Zhong 2008

momentum-residual row.

two limits now binds.

reach every rank.

inside test functions.

### What we did not do

lmoresi:
Merging after #591 (`cdcec34`), which this was stacked on — the geoid work is recorded there rather than here.

Two things from the cluster review that are carried forward rather than fixed by this merge:

- **`tests/test_1029_` is now used twice** (C2): `test_1029_analytic_faulted_medium.py` from #550 and `test_1029_analytic_zhong2008.py` from here. Both are present after the merge and both run; the number is what collides. Tracked for a renumber to 1030.
- **`Zhong2008` and `Zhong2008Response` are exported but not registered** (finding 1). That is correct — neither subclasses `AnalyticSolution`, so `uw.analytic.available()`, `describe()`, the conformance suite and `convention_audit.py` all pass them by. It is also invisible from the namespace, where everything else carries the validation contract.

The `FaultedMedium` removal shown in this PR's diff is an artefact of diffing against the merge base; we resolved the merge before merging and confirmed the result carries `FaultedMedium` and `Zhong2008` side by side.

Issue #546, opened 2026-08-13
Rotated workspace cache follow-ups from the #543 review: config latch into the cached KSP (m1), enumeration reuse by coincidence (m3)


lmoresi:
Two non-blocking findings from the #543 adversarial review of the rotated-workspace reuse (full detail in the review comment there):

1. **m1** — `solver.tolerance` and a custom-FMG registration latch into the cached KSP at ctx creation. A later tightening is still honoured on the true residual (measured 1e-4 → 1e-10, err 1.8e-11) but via extra outer increments, with the inner-margin discipline silently violated. The cache key (or the ctx) should observe the tolerance/FMG configuration and rebuild the KSP when it changes.
2. **m3** — the coefficient-variable enumeration is reused on geometry match alone; today this is safe only because every dependency-changing route forces a full `_build` teardown — a coincidence inherited from aux-layout invalidation, not enforced by the cache. Re-derive the enumeration on JIT-key mismatch so the safety is by construction.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

PR #597, opened 2026-08-17
An owned option's user latch dies with the option

### LINKED ISSUE #490: yield_anchor follow-ups from the #475 review: snes_max_it latch resurrect, side-attribution docstring, small validation gaps
Follow-ups from the #475 adversarial review (none blocked the merge; all probed at the PR head):

1. **`_snes_max_it_user` latch never clears** (`petsc_generic_snes_solvers.pyx` ~6428-6443): set `snes_max_it=200` → delete the option → the next solve correctly uses the default, but the one after resurrects the stale 200. The latch should clear when the key is absent from the options DB. (Still strictly better than the pre-#475 always-clobber, hence follow-up not blocker.)
2. **Docstring attributes the Min side to the smoother family** (~8810): "powermean (default, approaches the yield surface from below) or sqrt (from above)" — the side belongs to `yield_anchor`, not the family; under the default onset anchor both families sit below Min near yield. Also `anchor` is forwardable via `homotopy_options` (`**options` passthrough) but missing from the documented key list.
3. **Uncommented `except Exception: return default`** at the options read (~6433) — Charter swallowed-exception rule: state what is being swallowed and why.
4. **`yield_anchor` setter orphans a pre-built `YieldHomotopyControl`'s δ atom** (`constitutive_models.py:1030-1039`) — diagnostics-only exposure since `yield_continuation` refuses `anchor=` with a ready-made control; either rebind or document.
5. **Test name**: `test_yield_anchor_keeps_onset_stress_exact` exercises `anchor="yield"`; "onset" is the other anchor. Rename. A march-level test driving `anchor="yield"` through JIT/solve would also pin what is currently only probe-verified.
6. **`viscosity_min_rounding` setter unvalidated** (`constitutive_models.py:1192-1195`) — accepts negative/wrong-sign values silently.

lmoresi:
Closes #490.

Five of the six follow-ups from the #475 review. One is a live defect; the rest
are a swallowed exception, two documentation corrections, a test rename and a
setter that accepted a value with no reading.

## The latch outlived the option

`_snes_max_it_user`, which the issue names, no longer exists — it became the
general `_resolve_owned_option`. The defect survived the refactor.

`solve()` re-pushes the keys the solver owns, so the resolver has to tell its own
previous push from a value the user set, and latches the latter. Deleting the key
made the next solve fall back correctly, but that solve pushed the default; the
one after read the default back, found `current == pushed`, and returned the
latched value instead. Measured on `development`:

| | resolved |
|---|---|
| user sets `snes_max_it = 200` | 200 |
| user deletes the key | 50 |
| next solve | **200** — resurrected |
| and every solve after | 200 |

and on this branch, 200 / 50 / 50 / 50.

The latch is now cleared when the key is absent. `test_0204_owned_option_latch.py`
drives the resolve-then-push pair directly rather than through `solve()` — that is
the pair `solve()` calls, it needs no solve, and the test then says what it means.
It carries the control (a value still present keeps being honoured, three solves
running) and the case where the user moves the value rather than deleting it.

## The rest

- **Item 3.** The bare `except Exception: return default` at that read now says
  what it swallows — a stored value that will not convert to the option's type,
  such as a key set as a bare flag and holding `None` — and why taking the
  default is the right response (Charter §  on swallowed exceptions).
- **Item 2.** The `homotopy_options` docstring attributed the side of `Min` to
  the smoother family: *"powermean (default, approaches the yield surface from
  below) or sqrt (from above)"*. The side belongs to `yield_anchor` — under the
  default onset anchor both families sit below `Min` near yield. `anchor` is also
  added to the documented key list, which it was forwardable through and missing
  from.
- **Item 5.** `test_yield_anchor_keeps_onset_stress_exact` drives
  `anchor="yield"`; "onset" is the other anchor, the one the test shows missing.
  Renamed to `test_the_yield_anchor_keeps_stress_exact_at_nominal_yield`, with
  the reason in the docstring. No other reference to the old name exists.
- **Item 6.** `viscosity_min_rounding` is a rounding WIDTH — zero is the exact
  hard `Max`, positive rounds the corner — so a negative value has no reading. It
  reached the tangent only through the compiled expression, so the symptom of
  setting one was a solver that would not converge rather than anything naming
  the property. Now rejected at the setter, sympy values passed through
  untouched.
- **Item 4.** The `yield_anchor` setter discards the softness atom and so orphans
  the δ held by a `YieldHomotopyControl` built before the call. Documented rather
  than rebound, following the issue's own reading: `yield_continuation` refuses
  `anchor=` together with a ready-made control, so the only route to it is
  building a control by hand, setting the anchor after, and reading δ back for
  diagnostics. The comment says to set the anchor first.

## Not in this PR

#546 (the rotated workspace cache follow-ups) was scoped alongside this one and
is deliberately left out. Its two items change the cache key and the coefficient
enumeration in a subsystem where a wrong answer is quiet, and they want their own
measurement rather than a place at the end of a docs-and-latch batch.

## Verified

Full `./uw test`: 1512 passed, 32 skipped, 2 xfailed.

### DIFF
--- a/src/underworld3/constitutive_models.py
+++ b/src/underworld3/constitutive_models.py
@@ -1034,6 +1034,15 @@ class ViscousFlowModel(Constitutive_Model):
         # rebuilt, not merely revalued — and _get_yield_softness only builds it
         # alongside δ. (The power mean carries the anchor in _combine_yield itself, so
         # for that family the _reset is the whole of the update.)
+        #
+        # Discarding the softness atom ORPHANS the δ held by any
+        # YieldHomotopyControl built before this call: the control still ramps
+        # the atom it captured, and the model now reads a fresh one. That cannot
+        # bite a march, because `yield_continuation` refuses `anchor=` together
+        # with a ready-made control — so the only way to reach it is to build a
+        # control by hand, set the anchor afterwards, and read δ back off the
+        # control for diagnostics. Set the anchor BEFORE building the control
+        # (#490).
         self._yield_offset_expr = None
         self._yield_softness_expr = None
         self._reset()
@@ -1191,6 +1200,18 @@ class ViscousFlowModel(Constitutive_Model):
 
     @viscosity_min_rounding.setter
     def viscosity_min_rounding(self, value):
+        # A rounding scale is a width, so it is non-negative by construction:
+        # zero is the exact hard Max, positive rounds the corner. A negative
+        # value has no reading — it would sharpen the corner past Max — and
+        # since it reaches the tangent only through the compiled expression,
+        # the symptom of setting one is a solver that will not converge rather
+        # than anything naming this property (#490).
+        if value is not None and not isinstance(value, sympy.Basic):
+            if float(value) < 0.0:
+                raise ValueError(
+                    f"viscosity_min_rounding is a rounding WIDTH and cannot be "
+                    f"negative; got {value}. Use 0 for the exact Max.")
+
         self._viscosity_min_rounding = value
         self._reset()
 
--- a/src/underworld3/cython/petsc_generic_snes_solvers.pyx
+++ b/src/underworld3/cython/petsc_generic_snes_solvers.pyx
@@ -6783,19 +6783,36 @@ class SNES_Stokes_SaddlePt(SolverBaseClass):
         """
         pushed = self._owned_option_pushes.get(key)
         user = self._owned_option_user.get(key)
-        if self.petsc_options.hasName(key):
-            try:
-                current = type(default)(self.petsc_options.getString(key))
-            except Exception:
-                return default
-            if pushed is None or current != pushed:
-                # Never pushed by us, or the user has moved it since. Latch: from here on
-                # the option is theirs, because the next solve will read back OUR push of
-                # THEIR value and would otherwise mistake it for our own default.
-                self._owned_option_user[key] = current
-                return current
-            if user is not None:
-                return user
+        if not self.petsc_options.hasName(key):
+            # The key is gone from the options DB, so any value latched from a
+            # previous solve is gone with it. Without this the latch outlives
+            # the option: set snes_max_it=200, delete it, and the next solve
+            # correctly uses the default — but the one after that reads back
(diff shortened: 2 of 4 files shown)


lmoresi:
## Adversarial review

Reviewed at `3e96cc5`. Three findings.

**1. The latch fix is tested through the private pair, so it does not pin the
route users take.** `test_0204` calls `_resolve_snes_max_it` / `_push_snes_max_it`
directly. That is what `solve()` calls, and it makes the test a second rather
than three solves — but it assumes the call pattern, and a refactor that changed
where `solve()` resolves owned options would leave the test green over a
reintroduced defect. The end-to-end version costs three Stokes solves and would
pin `snes.getIterationNumber()` instead; we judged the exchange worth it, and are
flagging it because the judgement is the reviewable part.

**2. Item 6's guard admits sympy values without checking them.** A sympy
expression is passed through on the grounds that its sign is not knowable at set
time. That is right for a free expression, but `sympy.Float(-1.0)` has a knowable
sign and slips through — `is_negative` would catch the constant case. Left as is
because the property is documented as taking a number and the sympy path exists
for δ-style atoms, but it is a hole in an otherwise complete check.

**3. Item 4 is documented, not fixed, and the comment is the only thing stopping
it.** Setting `yield_anchor` after building a `YieldHomotopyControl` still leaves
the control ramping an atom the model has discarded. The reasoning that this is
unreachable rests on `yield_continuation` refusing `anchor=` alongside a
ready-made control — a guard in a different module. If that guard is relaxed, the
orphaning becomes reachable with nothing to signal it. A `_yield_softness_expr`
rebind, or the setter raising when a control is attached, would close it by
construction.

**Checked and clean:** the negative control for the latch was run against
`development`'s build in the main checkout rather than argued — 200 / 50 / 200 /
200 there, 200 / 50 / 50 / 50 here. No other reference to the renamed test exists
in `tests/`, `docs/`, `scripts/` or `src/`.

lmoresi:

is semantic.

### Cluster

|---|---|---|---|

that branch or an equivalent.

### #584

open:

  an oversight.
  than to this follow-up batch.

### #548

PR #612, opened 2026-08-19
Stokes JIT: rank-0 compile + CSE before printing + configurable flags (#547)

### LINKED ISSUE #547: Stokes pointwise-function JIT: cache=False bypasses the rank-0 compile gate (every MPI rank compiles), and -O3 OOM-kills gcc on large expressions
### Summary
On cold start under `mpirun -np N`, a Stokes solve triggers **N simultaneous JIT compilations** of the same pointwise-function module, compiled at `-O3`. Two current-dev behaviours combine to break large models on HPC:

1. `SNES_Stokes_SaddlePt._setup_pointwise_functions` is the only `getext` call site passing `cache=False`, which bypasses the rank-0-only compile + cross-rank disk handoff added in the April 2026 JIT-cache refactor. Every MPI rank compiles its own copy of the module — and because `cache=False` also bypasses the in-process `_ext_dict`, the module is recompiled on **every** `_setup_pointwise_functions` call, not just cold start.
2. JIT compiles use `extra_compile_args=['-std=c99','-O3']`. For large generated expressions (e.g. a multi-phase harmonic/geometric viscosity mixing law with strain weakening, pressure-dependent yield, temperature dependence, and viscosity clipping, plus its Jacobian), a single `gcc -O3` compile is slow and memory-hungry; multiplied across N ranks it exhausts the node.

### Observed behaviour
Under `mpirun -np N`, the first build of the pointwise functions for a Stokes solve produces N independent compile attempts (N distinct temp dirs), each grinding for minutes or being OOM-killed, followed by a per-rank `RuntimeError: The Underworld extension module does not appear to have been built successfully.` The repeated failed attempts burn the wall clock and the scheduler eventually cancels the job (`DUE TO TIME LIMIT`).

Full slurm output (36 MB, includes the verbose JIT dump): https://gist.github.com/bknight1/de63790873ee5abbc7fab93ef2c56753

Key evidence from that log:
- `gcc: fatal error: Killed signal terminated program cc1` — OOM kill of the compiler (repeated on many ranks, each in its own `/tmp/fn_ptr_ext_<hash>_<pid>_...` build dir).
- `mpicc -DNDEBUG -g -fwrapv -O3 -Wall -fPIC ... -std=c99 -O3` — JIT compile runs at `-O3`.
- One rank's compile **did** eventually succeed, but only at the job's 60-minute wall-clock limit — i.e. a single un-contended `-O3` compile of this expression takes ~55-60 min, and 20 concurrent compiles' combined memory is enough to OOM the node (estimate derived from the node spec; no RSS measurement available).

### Root cause

**Part 1 — every rank compiles.** `getext()` (`src/underworld3/utilities/_jitextension.py`, ~lines 690-749) gates the rank-0 compile on:

```python
multi_rank = underworld3.mpi.size > 1
disk_enabled = cache and _jc.get_cache_dir() is not None
if multi_rank and disk_enabled:   # rank 0 compiles, others Barrier+load
else:                             # every rank compiles
```

`SNES_Stokes_SaddlePt._setup_pointwise_functions` (`src/underworld3/cython/petsc_generic_snes_solvers.pyx`, ~line 7990) calls:

```python
_getext_result = getext(..., cache=False)
```

`cache=False` -> `disk_enabled=False` -> the `else` branch -> every rank compiles. The other `getext` call sites (scalar/vector solvers) use the default `cache=True`, so this is Stokes-specific.

History: `cache=False` was introduced at 2d3a0895 (2024-11-27) and touched by 9a808420 (2026-03-25), both *before* the disk-cache refactor (44c9a71c / 17e0a894, 2026-04-22), which **extended** the `cache` flag to also gate the disk cache and rank-0 compile. The Stokes call site was never updated, so it now bypasses both.

**Part 2 — `-O3` memory/time.** The flags are hardcoded in `generate_c_source` (`_jitextension.py`, line 1151) and have been `-O3` since the only commit that touched the line (7c1ac129, 2024-03-22). Optimisation passes on a single expression tree with a very large number of nodes consume large amounts of memory. The generated pointwise kernels are dominated by transcendental function evaluation, so `-O1` typically costs little at solve time while drastically cutting compile-time memory.

### Suggested fixes
1. Use the default `cache=True` (or drop the kwarg) in `SNES_Stokes_SaddlePt._setup_pointwise_functions`. **Interaction to note:** this eliminates the OOM (1 compile vs 20) but does *not* reduce the ~55-60 min single-compile wall clock — a larger `--time` budget or fix 2 is still needed to prevent the time-limit kill. The disk cache keys on the canonical-C `source_hash`, so there is no stale-`.so` reuse risk across models; it requires a cache dir visible to all ranks (`~/.cache/underworld3/jit` by default, `UW_JIT_CACHE_DIR` to relocate, `UW_JIT_CACHE=0` to disable).
2. Change the JIT default from `-O3` to `-O1` (or `-O2` as a middle ground — please weigh runtime vs compile-time memory) and/or expose the flags via an environment variable (e.g. `UW3_JIT_CFLAGS`, default `-O1`). **Blocking requirement:** benchmark runtime impact on a representative nonlinear Stokes solve before landing.

### Verification
- `test_0004_pointwise_fns` still passes (it parses the `Location of compiled module:` prefix, printed on rank 0 in both branches). Grep the test suite for the `JIT compiling new module` wording, since the rank-0 branch prints `JIT compiling new module on rank 0 ...`.
- A multi-rank Stokes solve shows a single `JIT compiling new module on rank 0 ...` line and a cache hit on re-run.
- Benchmark runtime impact of `-O1`/`-O2` on a representative nonlinear Stokes solve (blocker for fix 2).
- Consider a regression test asserting no rank > 0 enters `compile_and_load` when the disk cache is enabled.

### Environment
- 20 MPI ranks on 1 node, Python 3.11, mpicc (gcc 12.3), UW3 `development` branch (current), 60-min wall clock.

bknight1:
## Summary

Fixes the HPC compiler OOM and wall-clock timeout during runtime JIT compilation of Stokes pointwise functions (e.g. `gcc: fatal error: Killed signal terminated program cc1` under `mpirun -np N`) described in #547.

This PR implements coordinated changes across Stokes pointwise JIT compilation:

1. **Rank-0 Compile and Collective Gate**: `SNES_Stokes_SaddlePt._setup_pointwise_functions` now calls `getext` with `cache=True` (the only call site that previously opted out). Under MPI, this activates the rank-0-only compile gate and cross-rank disk handoff, preventing N concurrent compiler invocations per node. All compile decisions and cache viability checks are allreduced across communicator ranks (`needs_compile = comm.allreduce(..., op=MPI.LOR)`, `disk_enabled = comm.allreduce(..., op=MPI.LAND)`) to prevent rank-divergence deadlocks on high-latency shared filesystems.

2. **Configurable SymPy Common Subexpression Elimination (CSE)**: `generate_c_source` in `_jitextension.py` supports CSE before C printing. Shared subexpressions become intermediate `double xN = ...;` statements evaluated in topological dependency order. This shrinks generated C headers by up to ~63x on complex rheologies (1.45 MB → 23 KB) and lowers peak `gcc` RSS from ~1.32 GB to ~62–74 MB. **CSE is opt-in via `UW_JIT_CSE=1` (default is off to preserve original code generation behavior).**

3. **Configurable JIT CFLAGS**: Defaults to `-O3 -g0` to eliminate debug symbol bloat from Python `sysconfig`. Customizable via `UW3_JIT_CFLAGS` (e.g. `-O1 -g0`) for memory-constrained HPC nodes.

4. **Matrix-Level SymPy Differentiation**: Replaces per-entry derivative loops with `sympy.diff` across whole block matrices (~1.7x faster Jacobian generation, PETSc `[fc, gc, df, dg]` layout preserved).

Fixes #547

## Verification

- Rank-0 compile + collective barrier gate and disk-cache reuse confirmed under `mpirun -n 4` (deadlock divergence reproduction script `jit_divergence.py`, `rc=0`) and `mpirun -np 2` (`ptest_jit_cache.py`).
- Jacobian layout validated against the finite-difference oracle (`test_1066_stokes_jacobian_layout.py`).
- JIT determinism and constants manifest verified (`test_jit_cache.py`, `test_jit_deterministic_ordering.py`, `test_0103_jit_rampable_constants.py`).
- Core, Stokes, and constants test suites passing (33 tests); both default (CSE off) and opt-in (`UW_JIT_CSE=1`) paths verified.


lmoresi:
## Adversarial review

Reviewed against the PR head. CI is green and the change does what it says; one
finding is worth acting on before it lands, because this PR makes a latent
deadlock live on the main solver path.

### 1. The `Barrier()` is reached conditionally, on a rank-local predicate

In `_jitextension.getext` the cold-compile block is entered per rank:

```python
module = None
if cache:
    module = _jc.load_module(...)     # "Disk lookup is cheap and rank-local
                                      #  — every rank checks independently"
if module is None:                    # <-- rank-local decision
    ...
    if multi_rank and disk_enabled:
        if rank == 0: compile + store
        underworld3.mpi.comm.Barrier()   # <-- COLLECTIVE, inside the branch
        if rank != 0: load
```

The comment above the disk lookup states it is rank-local. The `Barrier` two
lines below is not. If any rank finds the published module on disk and any other
does not, the second enters the barrier alone and the job hangs.

Two ways they diverge, both realistic on the systems #547 is about:

- **Shared-filesystem visibility.** NFS attribute caching, Lustre metadata lag,
  or a partially-visible write means rank 3 sees the `.so` and rank 5 does not.
  This is the same class of hazard the barrier exists to manage, and the barrier
  sits inside the branch it is meant to protect.
- **`disk_enabled` differing.** `disk_enabled = cache and _jc.get_cache_dir() is
  not None`. A rank whose cache directory cannot be created or resolved takes the
  `else` branch — local compile, **no barrier** — while its peers take the
  barrier branch.

**Why this PR changes the exposure.** Before it, Stokes was the one call site
passing `cache=False`, so `disk_enabled` was False and Stokes always took the
no-barrier branch. Item 1 switches it to `cache=True`, which is the point of the
change — and puts the main solver path onto the branch containing a conditional
collective, on precisely the shared high-latency filesystems the PR targets.

The fix is small: decide collectively. Allreduce the "does anyone need a
compile?" predicate before branching so every rank takes the same branch, or lift
the barrier out of the conditional. Either makes the discipline true by
construction rather than by every rank happening to see the same directory
contents at the same instant.

### 2. The gate's verification is in a file nothing runs

`tests/parallel/ptest_jit_cache.py` is named in the verification list.
`tests/pytest.ini` sets `python_files = test_*.py`, so `ptest_*` is never
collected, and neither `scripts/test.sh` nor `scripts/test_levels.sh` mentions
`ptest` at all. The rank-0 compile gate is therefore covered by a manual
invocation and by nothing automatic. Given finding 1, that is the coverage worth
having. (#615 fixes the directory globs; it does not make `ptest_` files
collectible, which is a separate decision.)

### 3. "Bit-identical" is claimed, and the evidence is a finite-difference oracle

Item 4 says the matrix-level `sympy.diff` produces "bit-identical entries", and
the verification for it is `test_1066_stokes_jacobian_layout.py` against a
finite-difference oracle. An FD comparison validates layout and approximate
values; it cannot establish bit-identity. If the claim matters — and it is the
reason to believe a 1.7x codegen speedup is free — the check is a comparison of
the generated C, or of evaluated entries, against the per-entry path.

### 4. Checked and clean: the CSE is bit-safe

Worth stating because it is the part that looks riskiest. Naming a repeated
subexpression and reusing it yields the same bits as recomputing it, and
`sympy.cse` with the default `optimizations=None` does no algebraic rewriting, so
this is structural rather than a simplification. The coordinate patch is also the
right call: `cse` minting `BaseScalar` instances without `_ccodestr` is a known
trap, and `_patch_coords` is applied to both the replacement expressions and the
reduced one.

lmoresi:
## The collective, demonstrated — and a branch you can take

Following finding 1 above, we reproduced it and fixed it rather than leaving it
as an argument. The fix is one commit on top of this branch, at
`bugfix/jit-collective-gate` (`aaba141`).

### The deadlock is real

Monkeypatching the disk lookup so rank 0 sees the cache and the other ranks do
not — the shape NFS attribute caching or Lustre metadata lag produces just after
rank 0 publishes — then running an ordinary Stokes solve at np=4:

| | ranks entering the solve | ranks returning | result |
|---|---|---|---|
| this branch as it stands | 4 | **0** | `mpirun` timeout, rc=241 |
| with the reduction | 4 | 4 | rc=0, four consecutive runs |

Nothing exotic is needed to provoke it: the divergence is in which branch each
rank takes, and the barrier is inside the branch.

### What the fix does

Both predicates are reduced before they are used to choose a branch:

- `needs_compile` becomes a global **OR** of `module is None`, so if any rank
  lacks the module every rank enters the branch and reaches the `Barrier`.
- `disk_enabled` becomes a global **AND**, so a rank whose `get_cache_dir()`
  returns `None` makes every rank fall back together to local compiles rather
  than some waiting in a barrier the others never enter. Wasteful, and the safe
  direction.

Ranks that already hold the module keep it — the rank-0 compile, the
post-barrier load and the no-disk fallback are each guarded on `module is None`,
so nothing recompiles or reloads what it already has.

### Scope

The gate predates this PR; what this PR changes is the exposure. Stokes was the
one call site passing `cache=False`, so it always took the no-barrier branch.
Item 1 switching it to `cache=True` is the point of the change, and it puts the
main solver path onto the branch with the conditional collective — on exactly
the shared filesystems #547 is about.

Verified: full `./uw test` 1556 passed, 32 skipped, 2 xfailed; the four JIT test
files 14 passed.

@bknight1 — the branch is there to cherry-pick or ignore as you prefer; say the
word and we can push it onto this one instead. Findings 2 and 3 above (the
`ptest_` file nothing collects, and "bit-identical" evidenced by a
finite-difference oracle) are unaddressed and are not blockers.

lmoresi:
## How to take the fix — pick one, all three are fine

The commit is `aaba141` on `bugfix/jit-collective-gate`, sitting directly on top
of this PR's head (`8978ad8`), so it applies cleanly with no rebase.

**Option A — cherry-pick onto this branch (recommended).**

```bash
git fetch origin bugfix/jit-collective-gate
git checkout bugfix/stokes-jit-oom-cse-cache
git cherry-pick aaba1417b7d9ecd359d5407ceedd12987a406cf5
./uw build
git push origin bugfix/stokes-jit-oom-cse-cache
```

**Option B — merge the branch in.** Same result, keeps the fix as its own
commit with the merge recorded:

```bash
git fetch origin bugfix/jit-collective-gate
git checkout bugfix/stokes-jit-oom-cse-cache
git merge origin/bugfix/jit-collective-gate
./uw build
git push origin bugfix/stokes-jit-oom-cse-cache
```

**Option C — say so here and we will push it onto this branch for you.** It adds
one commit and changes nothing else; we have not done it unasked because it is
your PR.

### Verifying it yourself

The reproduction is a monkeypatch, not a real filesystem race, so it is
deterministic and takes a few seconds. Save as `jit_divergence.py`:

```python
import sympy
import underworld3 as uw
from underworld3.utilities import _jit_cache as _jc

_real = _jc.load_module
# Rank 0 sees the published module, the others do not — what NFS attribute
# caching or Lustre metadata lag produces just after rank 0 publishes.
_jc.load_module = lambda *a, **k: _real(*a, **k) if uw.mpi.rank == 0 else None

mesh = uw.meshing.StructuredQuadBox(elementRes=(4, 4))
v = uw.discretisation.MeshVariable("Vj", mesh, mesh.dim, degree=2)
p = uw.discretisation.MeshVariable("Pj", mesh, 1, degree=1)
s = uw.systems.Stokes(mesh, velocityField=v, pressureField=p)
s.constitutive_model.Parameters.shear_viscosity_0 = 1
s.bodyforce = sympy.Matrix([0, -1])
s.add_dirichlet_bc((0.0, 0.0), "Bottom")
s.add_dirichlet_bc((0.0, 0.0), "Top")

print(f"[{uw.mpi.rank}] entering solve", flush=True)
s.solve()
print(f"[{uw.mpi.rank}] SOLVE RETURNED", flush=True)
```

```bash
mpirun --timeout 120 -n 4 python -u jit_divergence.py ; echo "rc=$?"
```

`rc=241` is the deadlock (mpirun's timeout). `rc=0` is the fix. **Read the exit
code, not the printed lines** — mpirun drops a rank's final stdout at teardown,
so the "SOLVE RETURNED" count reads 3 or 4 on a healthy run at np=4 and is not a
reliable signal.

Clear the JIT cache between runs if you want a genuinely cold start
(`_jit_cache.get_cache_dir()` tells you where it is).

### After applying

Nothing else in the PR needs to change. `./uw test` was 1556 passed with the fix
applied, and the four JIT test files 14 passed.

Findings 2 and 3 in the review above — `ptest_jit_cache.py` being in a file
nothing collects, and "bit-identical" resting on a finite-difference oracle — are
not blockers and can be follow-ups.

bknight1:
Thanks @lmoresi — great catch on the rank-local predicate / collective barrier divergence hazard on shared filesystems.

### Updates applied:

1. **Collective Gate Applied (Option A):** Cherry-picked `aaba141` (`9b8b3cd7` on this branch) to allreduce `needs_compile` and `disk_enabled` before the barrier branch. Verified locally with the `jit_divergence.py` reproduction script under `mpirun -n 4` (`rc=0`) and `ptest_jit_cache.py` under `mpirun -np 2`.
2. **CSE Made Opt-In (Default Off):** To preserve original codegen behavior and exact output by default, CSE is now opt-in via `UW_JIT_CSE=1` (commit `630ed8d2`). Users on memory-constrained HPC nodes with large composite rheologies can activate CSE with `export UW_JIT_CSE=1`. Both default and CSE-enabled paths pass all test suites.
3. **Findings 2 & 3 noted:** We will track automatic harness collection of `ptest_*` parallel suites in a follow-up test harness cleanup PR.

Pushed and ready for CI.

Issue #611, opened 2026-08-19
test_global_evaluate_after_migration hangs at np=4, and its file is in neither parallel-test glob


lmoresi:
`tests/parallel/test_0760_swarm_cache_migration.py::test_global_evaluate_after_migration`
hangs at np=4 on current `development` (`8b7c8b9`). It passes at np=2.

## Reproduction

```
mpirun --timeout 90 -n 4 python -m pytest --with-mpi \
    tests/parallel/test_0760_swarm_cache_migration.py::test_global_evaluate_after_migration -q
```

`mpirun` reaches its timeout with no test result. The other two tests in the
file pass at np=4:

| test | np=2 | np=4 |
|---|---|---|
| `test_swarm_cache_valid_after_migration` | pass | pass |
| `test_global_evaluate_after_migration` | pass | **hang** |
| `test_global_evaluate_displaced_nodes` | pass | pass |

Not introduced by anything in flight — reproduced on `development` itself, and
the same run on `bugfix/shrinking-working-set` (#609) hangs identically, which is
how it was found.

## Why nobody has seen it

`scripts/test.sh` runs exactly two globs under `mpirun`:

```
mpirun -n $PARALLEL_RANKS python -m pytest --with-mpi tests/parallel/test_075*py
mpirun -n $PARALLEL_RANKS python -m pytest --with-mpi tests/parallel/test_10*py
```

`test_0760_...` matches neither — `test_075*` stops at 0759 and `test_10*` starts
at 1000. The file has never been run in parallel by our own script, at any rank
count. `PARALLEL_RANKS` also defaults to 0, so `./uw test` runs no parallel tests
unless `--p` is passed.

This is the #570 class exactly: a coverage gap in the glob, hiding a real defect.
The gap is the reason the defect is old and unnoticed rather than new.

## Why it matters now

The hanging test is `global_evaluate` after a swarm migration, which is the
subsystem under active work: #551 (locator anti-scaling), #604 (a located point
returning NaN), #606 (rank-local answers in parallel), and PR #602, which adds a
regression test **to this same file** and records its validation as
`mpirun -n 2 ... 4 passed per rank`. At np=2 everything here passes, so that
validation would not have caught this.

## Suggested handling

1. Fix the globs so `tests/parallel/` is covered by pattern rather than by two
   ranges that happen to have a hole between them — the hole will recur
   otherwise.
2. Run the parallel selection at np=4 as well as np=2. Every test that passes at
[...]

Issue #573, opened 2026-08-15
CI wall-clock tracks total test time: split the batches across parallel jobs (the 60-minute cap was reached, 120 is a stop-gap)


lmoresi:
`scripts/test.sh` runs every batch sequentially inside one GitHub Actions job, so CI wall-clock is the *sum* of all test time. Measured 2026-08-15:

| | |
|---|---|
| serial phase | ~55 min (06:47→07:42 on run 31870162095) |
| three individual batches | 14–15 min **each** |
| parallel phase | runs LAST, so it is what gets cut |
| a run with no test additions (PR #568) | 55m07s |

Two PRs were cancelled at exactly 1h00m and surfaced as red X's when nothing had failed — #571 twice. PR #572 raises the cap to 120 minutes as a stop-gap so work can land, but that treats the symptom: the suite will keep growing, and the next person to add a batch's worth of tests hits it again.

## The durable fix

Split the batches across parallel jobs (an Actions matrix), so wall-clock tracks the *longest* batch rather than the sum. The batches are already enumerated in `scripts/test.sh`, so the matrix axis largely exists:

- one job per serial batch group (or a few groups balanced by measured runtime — the 14–15 min batches should not share a job);
- the parallel (`mpirun`) phase as its **own** job, so it can never again be the thing that gets truncated;
- the analytic full-family tier (`tests/analytic_full/`, added in #571) as an opt-in job that need not run on every PR.

Worth doing together with #570 (the coverage audit), since both concern `scripts/test.sh`'s batch structure: #570 asks which files are covered, this asks how the covered files are scheduled. A matrix built from a single declarative list of batches would answer both — each batch becomes a job, and any file matching no batch fails the audit.

## Why it matters beyond speed

A cancelled job reports as a failure indistinguishable from a real one. #571's first timeout cost a full diagnostic round hunting a test failure that did not exist, and the parallel phase being last means the tests most likely to be truncated are the parallel-correctness ones — exactly the class where today's defects (#564, #560) were found.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

PR #615, opened 2026-08-19
Run the whole parallel directory, and at four ranks as well as two

lmoresi:
Addresses part of #570, and is how #611 was found.

## Two scripts, two different holes

| script | glob | files named | missed |
|---|---|---|---|
| `scripts/test.sh` (CI runs this) | `test_075*py` + `test_10*py` | 14 of 32 | 0005, 0700, 0760–0790, 0855, 0873 |
| `scripts/test_levels.sh` (`./uw test`) | `test_07*py` | 20 of 32 | 0005, 0855, 0873, **all of `test_10*`** |

Three files ran at **no rank count in either script**: `test_0005_xdmf_viz_topology_mpi`,
`test_0855_mesh_smoothing_parallel`, and `test_0873_adapt_collective_stop_mpi` —
the last added two days ago in #596 specifically to guard against a parallel
hang, and never once run in parallel by our own tooling.

Both scripts now name the directory. A glob that names ranges grows holes as
files are added between them; `test_075*` stops at 0759 and `test_10*` starts at
1000, so everything added in between fell through. That is the #570 class, and it
is why #611 sat undetected.

## Four ranks, opt-in

Run unconditionally the four-rank pass took the test job past the 120-minute cap
— recent runs here take 26 to 40 minutes, and on a two-core runner np=4 is
oversubscribed and costs far more than the ~3 minutes it takes on a workstation.
It is behind `--full-parallel`, off by default, which is the shape
`test_levels.sh` already used. CI keeps the directory fix, which is what closes
the hole; the four-rank pass belongs in a nightly or a separate job (#573).

The reason to run it at all:

Two ranks is a special case. The defect this suite exists to catch is a
collective entered by some ranks and not others, and with two ranks the
mismatched pair often still meets — the skipped branch does not arise on a
two-way partition, or one rank's internal barrier is satisfied by the other's
next collective. Every instance found recently passed at np=2 and hung at np=4:
the conditional collective in #609, and #611.

`test_levels.sh` already had a `--full-parallel` flag for exactly this and was
pointing it at the narrower glob.

## Measured

Machine otherwise idle, since contention distorts these badly:

| | result | time |
|---|---|---|
| `tests/parallel/` at np=2 | 128 passed, 11 skipped | 156 s |
| `tests/parallel/` at np=4, 1 deselected | 135 passed, 3 skipped | 170 s |
| file-by-file at np=4 | 31 of 32 pass | — |
| end to end, `scripts/test.sh --p 2` | 135 passed, 3 skipped, 1 deselected, rc=0 | 198 s for the parallel section |

The skip counts differ between rank counts because some tests require four ranks
and skip at two — a second reason to run both, independent of the collective
argument.

Cost to CI: the directory pass at the requested rank count only. The four-rank
pass is opt-in for the reason above; #573 is open on the total.

## The one deselection

`parallel/test_0760_swarm_cache_migration.py::test_global_evaluate_after_migration`
passes at np=2 and hangs at np=4 on `development` — that is #611, filed
separately, not introduced here.

Note the node id has no `tests/` prefix: `tests/pytest.ini` puts rootdir at
`tests/`. A deselect that does not match any test is ignored in silence, so a
mistyped node id leaves the test running and the pass hanging, with nothing said
about it. Confirm `1 deselected` in the output when changing it.


lmoresi:
## Adversarial review

Reviewed at `ad43614`. Four findings.

**1. The deselection is a silent-failure mechanism guarding a silent failure.**
`--deselect` with a node id that does not match is ignored without complaint —
that is precisely what produced two wrong diagnoses while measuring this PR. So
the guard against #611 is itself the kind of thing that can stop working without
saying so: rename the test, move the file, or change `pytest.ini`'s rootdir and
the deselect quietly stops matching and the np=4 pass hangs again. A
`--strict-markers`-style equivalent does not exist for deselect. Marking the test
`@pytest.mark.skipif` on rank count, inside the file, would fail loudly if the
name drifted; it also puts the exclusion where a reader of the test will see it.

**2. Turning on 18 previously-unrun files rests on one machine.** They pass here
at np=2 and np=4, on macOS/arm64 with Open MPI 5.0.10 and `--oversubscribe`
above the core count. CI is ubuntu-latest with a different MPI and different core
count, and several of these files have never run in that environment at all. The
first CI run of this PR is the real test, and it may surface failures that are
about the runner rather than about the tests. That is a reason to watch it land,
not a reason to keep the hole.

**3. The np=4 pass is hard-coded to 4.** `if [ "$PARALLEL_RANKS" -ne 4 ]` then
`mpirun -n 4`, so asking for `--p 8` gives 8 and 4, while `--p 4` gives 4 alone.
That is defensible — 4 is the cheapest rank count that exposes the mismatched-
collective class — but it is a convention baked into two scripts rather than a
variable, and someone wanting 2-and-8 has to edit both.

**4. `test_levels.sh` still has two paths where `test.sh` has one.** Its
`--full-parallel` flag now duplicates what `test.sh` does unconditionally, so the
same directory is described in two places with slightly different flags. The
duplication predates this PR and is not made worse by it, but the two scripts
having diverged into different globs is exactly what caused the three-file hole,
and they will diverge again.

**Checked.** `bash -n` on both scripts; the full `scripts/test.sh --p 2` path end
to end at rc=0; and the deselect verified to actually match this time by
confirming `1 deselected` in the output, which is the check that was missing
when it silently did not.

lmoresi:
Holding this, and the reason is a point in its favour rather than against it.

Rebased onto current `development` and re-run. The job went red at **2h0m20s**,
and none of that time was the extra coverage being paid for:

```
01:11:04.49  tests/parallel/test_0855_mesh_smoothing_parallel.py ......F
01:11:04.82  tests/parallel/test_0873_adapt_collective_stop_mpi.py F
02:27:58     ##[error]The operation was canceled.
             Terminate orphan process: pid (45533) (mpirun)
```

Two failures and then 76 minutes of silence with a live `mpirun` — a rank stuck
in a collective after another had already left it.

Both files are among the three this PR identifies as matching no glob in either
script. This is the first time CI has ever run them, and they do not pass. That
is precisely the hole this PR exists to close, and `test_0873_adapt_collective_stop_mpi`
is the one added in #596 to guard a rank-divergent collective stop — it has been
dead weight since it landed and it fails the moment it runs.

Filed as #675 with the evidence, including what does **not** reproduce: on macOS
at np=2 both files pass alone, the failing batch of six passes alone, and
batches 1-4 run in sequence in one working directory pass 42 / 19 / 23 / 14. So
cross-batch contamination — the obvious suspect — is ruled out, and it is
Linux-specific or runner-specific. No traceback was captured because the job was
cancelled before pytest could summarise.

Not merging until #675 is fixed: as it stands this would hand every CI run a
two-hour hang. Nothing is wrong with the change itself.

One thing worth deciding separately, once the tests pass: this widens the
parallel job from `test_075*` + `test_10*` to the whole directory and adds np=4
alongside np=2. CI has already gone 19 -> 48 minutes for coverage once (#573 is
open about exactly that). The honest runtime for the full matrix is not known
yet, because this run never got to finish one.

Issue #616, opened 2026-08-19
Rotated free slip on a FLAT wall does not reproduce the equivalent component Dirichlet condition (2e-3, and not a convergence artefact)


lmoresi:
On a flat, axis-aligned wall `add_rotated_freeslip_bc(0.0, "Top")` and
`add_dirichlet_bc((None, 0.0), "Top")` are the same discrete constraint: the
measure-weighted node normal is exactly `(0,1)`, so striking the rotated normal
row is striking `u_y`. The two solves do not agree.

## Minimal reproduction

Unit box, `StructuredQuadBox(elementRes=(32,32))`, **uniform viscosity 1**,
`bodyforce = (0, cos(pi x) sin(pi z))`, free slip on all four walls (`Left`,
`Right`, `Bottom` by component Dirichlet), `petsc_use_pressure_nullspace = True`,
`tolerance = 1e-9`. Reference is a monolithic LU solve of the component-Dirichlet
problem. The two top corner nodes are EXCLUDED from every norm (they are the
known recovery corner, #608).

| run | ‖u − u_ref‖ / ‖u_ref‖ |
|---|---|
| component Dirichlet, default solver | 8.9e-09 |
| component Dirichlet, `pc_type=lu` | 0 (the reference) |
| **rotated free slip, default solver** | **2.0e-03** |
| **rotated free slip, `_rotated_use_lu = True`** | **2.1e-03** |

The last row is the point: with an exact linear solve on both sides the answers
still differ by 2e-3, so this is the discrete problem, not convergence.

## What it is not

- **Not the corner.** Dropping a ball of radius 1/32, 2/32, 4/32, 8/32 around each
  top corner takes the ratio of velocity errors from 1.49 to 1.19; a 20% excess
  survives dropping a quarter of the domain.
- **Not the tolerance.** `tolerance = 1e-9` and `1e-12` give bit-identical
  answers on the rotated path (3.313e-03 both, SolCx eta_B=10).
- **Not the pressure gauge.** The rotated path builds the constant-pressure null
  vector; its pressure comes back with a non-zero mean (−0.224 where the
  Dirichlet run gives 1e-12), but a constant pressure has no gradient and cannot
  move the velocity.
- **Not the viscosity contrast** — and this is the odd part. Same comparison
  against contrast: uniform 3.2e-03, `eta_B=10` 2.7e-03, `eta_B=1e6` **9.3e-08**.
  The two agree on the hard problem and disagree on the easy one.

## Why it matters

[...]

Issue #614, opened 2026-08-19
boundary_flux() returns ~1e12 on Stokes_Constrained


lmoresi:
`solver.boundary_flux(boundary, normal=...)` — the consistent-boundary-flux
back-calculation — reads correctly on an ordinary `Stokes` solve and returns
values of order 1e12 on `Stokes_Constrained`.

## Measured

SolCx, `StructuredQuadBox(elementRes=(32, 32))`, free slip on all four walls,
`normal = [[0, 1]]` on `Top`, compared against `SolCx.topography_top` (mean
removed, corners excluded):

| solver | `Top` held by | max abs(CBF) | vs exact |
|---|---|---|---|
| `Stokes` | `add_dirichlet_bc((None, 0.0), "Top")` | 0.381 | rel 0.082, corr +0.997 |
| `Stokes` | `add_rotated_freeslip_bc(0.0, "Top")` | 0.497 | rel 0.087, corr +0.996 |
| `Stokes_Constrained` | `add_constraint_bc(0.0, "Top")` | ~1e12 | uncorrelated (corr 0.03) |

Exact peak is 0.379. The constrained solve itself is fine — velocity error
8.8e-06 at `eta_B = 1e6`, and its recovered `n·σ·n` matches the exact to 0.075 —
so this is the flux primitive on that solver, not the solve.

Likely `_assemble_volume_reaction()` picking up the multiplier field's rows or
the augmented-Lagrangian boundary stiffness (`r = 1e4·μ`, which the SolCx step
makes 1e10 on the stiff half); the magnitude is consistent with the latter.

## Why it matters

It is the natural cross-check for #607: at convergence the momentum row's
boundary term satisfies

```
M_Γ (h + r(n·u − g)) = −(A·u − b)|_Γ
```

so `h + r(n·u − g)` IS the CBF traction de-smeared by the boundary mass. Being
able to read both off one solve is how that identity gets verified, and at the
moment it cannot be. Across two separate solves the corrected multiplier and the
rotated CBF reaction agree to 3–5%, which is inside each route's own error.

Reproduction: `examples/solcx.py` in UWTN 2026-016.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
**Correcting my own report: the value is ~0, not ~1e12.** The 1e12 I quoted was a
relative measure whose denominator was the near-zero CBF itself. Measured
[...]

Issue #608, opened 2026-08-19
Rotated free slip meets a component Dirichlet wall: a local pressure error at the shared corner


lmoresi:
On a box, a rotated free-slip constraint on `Top` and the ordinary component
condition `add_dirichlet_bc((None, 0.0), "Top")` pin the same velocity
components at the corner where `Top` meets `Left`/`Right` — the corner node has
`u_x = 0` from the side wall and `u_y = 0` from either treatment. The two runs
do not return the same answer there.

## Measured

SolCx, `StructuredQuadBox(elementRes=(32, 32))`, `eta_B/eta_A = 10`,
`petsc_use_pressure_nullspace = True`, free slip on `Left`, `Right`, `Bottom`
via component Dirichlet. Two runs differing only in how `Top` is held:
`add_dirichlet_bc((None, 0.0), "Top")` against `add_rotated_freeslip_bc(0.0, "Top")`.

Velocity errors against the analytic solution are 1.7e-03 (Dirichlet) and
3.3e-03 (rotated). The pressure difference between the two runs, both mean
removed:

| x | y | p_rotated − p_dirichlet |
|---|---|---|
| 1.0000 | 1.0000 | +2.32e-01 |
| 0.9687 | 1.0000 | −1.25e-01 |
| 0.9375 | 1.0000 | +2.49e-02 |
| 0.0000 | 1.0000 | −2.33e-02 |

rms below the top row: 4.2e-04, against a pressure rms of 1.1e-01. So the
disagreement is the corner and its immediate neighbours, not the wall.

Downstream: the recovered surface stress along the top wall reads 3.2e-01
relative for the rotated run against 4.8e-02 for the Dirichlet run. Trimming two
elements from each end brings them together — 5.6e-02 against 5.4e-02 — so the
whole of that factor of six is the two corner nodes. Same at every viscosity
contrast from 10 to 1e6.

Not necessarily a bug in the constraint: at a corner the rotated frame's normal
row is a mix, which `boundary_normal_traction` already handles by projecting the
Cartesian reaction. The question is whether the corner row of the rotated
velocity block should reduce to the component condition when the two coincide,
as it does here.

Reproduction: `examples/solcx.py` in UWTN 2026-016 (*Boundary conditions on
non-planar boundaries*), `python3 solcx.py contrast`.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
[...]

PR #617, opened 2026-08-19
Return the whole boundary traction from the constrained solver (#607)

### LINKED ISSUE #607: Stokes_Constrained: multiplier() is not the whole boundary traction — the augmented-Lagrangian share is missing
`add_constraint_bc` assembles the momentum row as

```
fn_f = (h + r*(n·u − g)) * n
```

so the traction holding the boundary is `h + r(n·u − g)`, not `h`. `multiplier()`
and `topography()` return `h` alone, and the docstring states that accuracy is
independent of the augmentation (`augmentation_base`, default `1e4`):

> Accuracy is independent of this value (the multiplier carries the exact constraint)

That holds in the exact limit. Discretely the constraint row is satisfied to the
solver's tolerance, and `r` multiplies that residual back into the traction.

## Measured

Annulus, `radiusInner=0.5`, `radiusOuter=1.0`, isoviscous, smooth density
`(r/r_o)^3 cos 2θ`, exact velocity Dirichlet on `Lower`, `add_constraint_bc(0.0,
"Upper")`. The exact surface radial stress is `0.1506696 cos 2θ`
(`uw.analytic.CylindricalStokes`, Kramer et al. 2021). Relative error in that
amplitude, read off the multiplier's boundary trace:

| cell | r | max abs(n·u) | h alone | h + r(n·u) |
|---|---|---|---|---|
| 0.100 | 1e4 | 5.2e-07 | 2.9e-02 | 8.5e-04 |
| 0.100 | 1e2 | 5.6e-07 | 3.7e-03 | 3.7e-03 |
| 0.075 | 1e4 | 4.8e-07 | 2.5e-02 | 1.7e-03 |
| 0.075 | 1e2 | 4.4e-07 | 2.6e-03 | 2.6e-03 |

At the default `r = 1e4·μ` the reported multiplier is a few per cent low, and
the error tracks `r` rather than the mesh — at `cell=0.05` the same run
converged to `abs(n·u) = 1.1e-08` and `h` alone read 1.5e-03, so the column does
not converge monotonically either. Adding the augmentation share removes the
dependence on `r`.

For comparison, `boundary_normal_traction` on the rotated constraint reads
2.1e-03 at `cell=0.075` on the same problem, and a projection of `n·σ·n` out of
the solved fields reads 6.3e-03.

## Suggested fix

Return `h + r(n·u − g)` from `multiplier()` (and hence `topography()`), or —
if the bare multiplier is wanted as well — keep `h` available and make the
traction the default. Either way the docstring's independence claim needs to
name the condition under which it holds.

Found while measuring surface stress for UWTN 2026-016 (*Boundary conditions on
non-planar boundaries*); the reproduction is `examples/stress.py` in that note.

lmoresi:
Closes #607.

`Stokes_Constrained` assembles the momentum row's boundary term as
`(h + r(n·u − g))·n`, so the traction holding the boundary is that sum.
`multiplier()` returns `h`, and `topography()` was built on `h` alone — short by
the augmented-Lagrangian share, which is `r` times the discrete constraint
residual.

With the viscosity-weighted default `r = 1e4·μ(x)` that share is a few per cent
of the surface traction on a uniform-viscosity annulus. Across SolCx's `1e6`
viscosity step, where `r` reaches `1e10` on the stiff half, it is most of it:

| route | max abs(h) | corr with exact | relative l2 |
|---|---|---|---|
| `multiplier()` as returned | 0.042 | **−0.53** | 1.04 |
| `traction()` (this PR) | 0.382 | +0.999 | **0.047** |
| exact `SolCx.topography_top` | 0.383 | — | — |
| projected `n·σ·n`, same solve | 0.383 | +0.997 | 0.075 |

The corrected read is the most accurate surface stress available on that problem
— better than the projection — and the solve itself was never at fault
(velocity error 8.8e-06, `max abs(n·u)` 1.7e-10).

## What changed

- **`traction(boundary)`** — new, returns `h + r(n·u − g)` symbolically.
- **`topography()`** — built on it, so the public dynamic-topography path is
  correct with no change at the call site.
- **`multiplier()`** — unchanged behaviour (it returns `h`) and now documents what
  it is not.
- Docstrings that claimed accuracy is independent of `r` now say the CONSTRAINT
  and the `traction()` read are independent of it, and `h` alone is not.

## Why it survived validation

The existing check scored `corr(λ, −n·σ·n) = 0.9999`. A correlation is scale-free
and cannot see a systematic amplitude deficit, which is precisely what a missing
share of the load is. `tests/test_1063_constrained_traction.py` scores a relative
`l2` against the exact SolCx topography and carries the bare multiplier as its
**negative control** (it must read > 0.5), so a regression that quietly reverts to
`h` cannot pass. `test_1061`'s API test asserted `topography == h/scale`, which
pinned the defect; it now asserts the outcome.

## The identity behind it

At convergence `M_Γ(h + r(n·u − g))` balances the volume residual restricted to
the boundary, which is the CBF nodal load (Zhong, Gurnis & Hulbert 1993). So the
multiplier route and `boundary_normal_traction` on a rotated constraint are the
same computation; measured across two solves they agree to 3–5%, inside each
route's own error. This is also why the free-surface work was right to reject the
multiplier as returned and keep the rotated lid + CBF.

## Documentation

- `CONSTRAINED_FREESLIP_MULTIPLIER.md`: dated correction, corrected CBF identity.
- `docs/advanced/curved-boundary-conditions.md`: the penalty free-slip recipe now
  uses `mesh.boundary_normal` rather than `mesh.Gamma`, with the measurement that
  a facet-normal penalty does not converge — velocity error 0.60 flat and stress
  error 0.21 → 0.26 under refinement, while the leak reads 1e-5 throughout. The
  projected-normal recipe is marked superseded.
- `CHANGELOG.md` entry.

## Tests

`tests/test_1061`, `1062`, `1063`, `1064`, `1024`: 26 passed, 2 xfailed.

Related, filed separately and NOT addressed here: #608 (the reaction reported at
a corner shared with an essential BC), #614 (`boundary_flux` returns ~1e12 on
`Stokes_Constrained`), #616 (rotated free slip does not reproduce the equivalent
component condition on a flat wall, 2e-3, with an exact linear solve on both
sides).


lmoresi:
Extended to cover the other two paths, so the fix lands as one consistent set
rather than leaving them to follow.

## The corner was never constrained (#616, and the mechanism behind #608)

`build_rotation` skipped any boundary node with a velocity DOF constrained out of
the global vector — every node where a rotated wall meets a wall held by an
essential condition. The node was dropped from the rotation, so **the wall-normal
component was never constrained there**. The wall leaked at its own end points
while every interior node was exact.

Unit box, rotated lid, component free slip on the other three walls, uniform
viscosity:

| | before | after |
|---|---|---|
| max abs(u_y) on the lid | 4.0e-03 (16% of abs(u)max, both corners) | **0.0** |
| vs a monolithic LU solve of the component-Dirichlet problem | 2.0e-03 | **8.9e-09** |
| same, with `_rotated_use_lu` on both sides | 2.1e-03 | 8.9e-09 |

The second row with an exact linear solve on both sides is the point: it was
never a convergence artefact, it was a different discrete problem.

The fix: a partially constrained node keeps the constraint on what is left. With
the pinned components at zero, `n·v = 0` reduces to `n_F·v_F = 0` on the free
subspace, so the frame is built there and its normal rows constrained. A node
whose normal lies entirely in the pinned subspace is already implied and is still
skipped; a node carrying a prescribed `v_n` datum keeps the old behaviour and says
so once (reducing the affine constraint against the pinned values is not
implemented).

Guard: `tests/test_1066_rotated_meets_essential.py`. Existing suites unaffected —
42 rotated / Nitsche / constrained tests pass.

## boundary_flux on a constrained boundary (#614) — and a correction

I reported that it returns ~1e12. **It returns ~0** (rms 3.9e-13 against a
traction of 0.37); the 1e12 was a ratio whose denominator was that near-zero. And
it is expected: a multiplier-constrained boundary has no reaction left to read,
because the constraint enters the same row it constrains. That is the #607
identity from the other side.

It now raises and points at `traction()` rather than handing back a quiet zero.

Issue #614 is corrected on the issue itself.

Issue #641, opened 2026-08-25
evaluate() returns large negative values for a non-negative expression (sqrt invariant) at in-domain points near a boundary singularity


lmoresi:
`uw.function.evaluate` returns **large negative values** for an expression whose range is
provably non-negative, at points **inside** the domain.

`Unknowns.Einv2` is `sympy.sqrt((E**2).trace()/2)` — the square root of a sum of squares
of a real symmetric tensor. It cannot be negative for any velocity field.

Lid-driven cavity, `UnstructuredSimplexBox` at `cellSize=1/8`, P2/P1, direct solve:

```
negative in-domain samples:
   coord [0.00536 0.86533]        value -0.49758
   coord [0.00783 0.83353]        value -0.09049
   coord [0.01006 0.87349]        value -0.42628
```

All three hug the **left wall just below the lid corner** — the cavity's velocity
singularity, where `Einv2` is large and varying sharply. Magnitudes of −0.1 to −0.5 are
not round-off.

Properties:

- **Reproducible.** Identical values whether the points are evaluated in a batch of 1620
  or one at a time, so it is not a batching or ordering effect.
- **In-domain.** The same run has 815 points deliberately placed *outside* the box, and
  all of those came back positive. The defect is not extrapolation.
- **Independent of the grad-div penalty.** Present at `penalty=0` (min −0.4976) and
  `penalty=10` (min −0.4838) alike.
- Not obviously the same as #551, which is about locator passes and scaling rather than
  returning a value outside the expression's range.

A negative square root means the value returned is not the expression evaluated at that
point. The plausible shape is that a representation of the *scalar* is being interpolated
rather than the expression being evaluated pointwise from the field, and interpolation of
a sharply peaked non-negative function undershoots. Not confirmed — it needs someone who
knows the evaluate path to say which branch is taken here.

### Why this surfaced

`tests/test_0112_swarm_add_particles.py::test_swarm_access_write_after_add_particles`
accumulates `0.5 * evaluate(Einv2, swarm.data)` and asserts the running total is
non-negative. That premise is unsound for as long as this bug exists. The test currently
passes only because the accumulation happens to stay positive; perturbing the velocity
field slightly (we hit it by changing `Stokes.DEFAULT_PENALTY`) flips it negative and the
test fails deterministically — 4/4 runs, while 4/4 pass unperturbed. It reads as a
penalty regression and is nothing of the sort.

[...]

PR #646, opened 2026-08-25
Add distributed harmonic projection of rotated reactions

gthyagi:
## Summary

This PR adds a distributed weak projection of rotated free-slip boundary reactions and exposes it through the generic spherical-shell geoid postprocessing adapter.

The change is intentionally opt-in:

- `projection="centroid"` remains the default for compatibility.
- `projection="reaction"` directly contracts the assembled normal reaction with the requested harmonic.
- The new `Stokes.boundary_normal_traction_integral(boundary, fn, remove_mean=True)` API is generic and is not tied to Zhong 2008.

## Problem

The existing centroid workflow first recovers pointwise boundary-normal traction, gathers boundary samples to rank zero, reconstructs a spherical triangulation with `ConvexHull`, and fits the harmonic over triangle centroids.

That workflow was introduced for a valid reason: as discussed in #414, #431, and #404, raw P2 vertex values on faceted curved boundaries contain a slowly converging chord-versus-arc geometry error. Midpoint, fitted, and integral quantities are preferable to treating those vertices as physical point samples.

However, when the required output is only a harmonic coefficient, pointwise recovery is unnecessary. It also introduces:

- a global boundary-coordinate/value gather;
- rank-zero surface triangulation and integration;
- pointwise P2 mass recovery before fitting;
- an opportunity to mix a faceted FE reaction with an analytical spherical norm.

## Method

For the converged rotated-free-slip solve, let `R` be the assembled nodal normal-reaction load and let `phi` be a scalar boundary test function. The new API computes

```text
I(phi) = phi^T R - (1^T R / area) * integral(phi)
```

using owned reaction DOFs followed by MPI scalar reductions.

For a spherical harmonic `phi_l = P_l(cos(theta))`, the topography coefficient is normalized with the matching discrete boundary inner product:

```text
h_l = sign * I(phi_l) / (buoyancy_scale * integral(phi_l^2))
```

Using `BdIntegral(phi_l^2)` is important because both numerator and denominator then use the same faceted FE geometry. The method never reconstructs or consumes recovered P2 vertex values, so it follows the midpoint/fitted/integral guidance from #414 rather than bypassing it.

## Zhong 2008 layered validation

Benchmark: [`bench_020_zhong2008_layered_viscosity_response.py`](https://github.com/gthyagi/uw3-mantle-convection-benchmarks/blob/main/benchmarks/bench_020_zhong2008_layered_viscosity_response.py)

Configuration: `l=5`, load depth `0.5d`, `cellsize=1/64`, fitted `10^4` viscosity lid, P2P1, rotated free slip, 192 MPI ranks, `stokes.tolerance=1e-5`.

Both runs used the identical mesh, Stokes formulation, solution fields, and Zhong propagator reference. Only the topography projection changed.

| Self-gravity response | Analytic | Centroid | Centroid error | Reaction | Reaction error |
| --- | ---: | ---: | ---: | ---: | ---: |
| Surface topography | 0.456720 | 0.444767 | -2.62% | 0.457084 | +0.08% |
| CMB topography | 0.510978 | 0.515630 | +0.91% | 0.510959 | -0.00% |
| Surface geoid | 0.026961 | 0.025881 | -4.01% | 0.026994 | +0.12% |
| CMB geoid | 0.014956 | 0.015134 | +1.19% | 0.014956 | +0.00% |

| No-self-gravity response | Analytic | Centroid | Centroid error | Reaction | Reaction error |
| --- | ---: | ---: | ---: | ---: | ---: |
| Surface topography | 0.408246 | 0.398236 | -2.45% | 0.408551 | +0.07% |
| CMB topography | 0.466978 | 0.471107 | +0.88% | 0.466957 | -0.00% |
| Surface geoid | 0.022494 | 0.021589 | -4.02% | 0.022521 | +0.12% |
| CMB geoid | 0.012534 | 0.012695 | +1.28% | 0.012534 | +0.00% |

The surface-geoid error falls from about 4.0% to 0.12%. All eight topography/geoid coefficients are within 0.13% of the semi-analytical propagator after the change.

Velocity is measured independently and is unchanged by this postprocessing method. In this run, CMB velocity differs from the propagator by 0.02%; the near-zero surface velocity differs by 8.46%.

## Parallel and performance properties

- Each rank reads only reaction DOFs it owns.
- Ghost/shared boundary nodes are not double counted.
- Only scalar sums are reduced across MPI.
- No global boundary topology or recovered pointwise traction is gathered.
- No rank-zero `ConvexHull` or surface mass solve is used.

A controlled 192-rank before/after run used the identical mesh, solver configuration, and benchmark script revision apart from the projection selection. The centroid baseline was rerun from benchmark commit `a3344ac` as Gadi job `177431139`; the reaction run was job `177429270`.

| Metric | Centroid baseline | Reaction | Change |
| --- | ---: | ---: | ---: |
| Mesh loading | 45.16 s | 47.04 s | +4.2% |
| Stokes solve | 784.55 s | 772.01 s | -1.6% |
| HDF5 output | 4.22 s | 4.73 s | +12.1% |
| Response integrals, topography, and geoid | 120.14 s | 36.92 s | **-69.3% (3.25x faster)** |
| PBS wall time | 16:58 | 14:59 | -11.7% |
| Peak memory | 324.27 GB | 337.75 GB | +4.2% |
| Exit status | 0 | 0 | unchanged |

`Benchmark.ResponseIntegrals` includes the common velocity/divergence integrals as well as topography recovery and geoid evaluation. Those common calculations are identical between runs, so the measured 83.22 s reduction isolates the effect of replacing centroid recovery's global gather, rank-zero triangulation, and pointwise traction fit with the distributed reaction contraction.

The small Stokes, mesh, HDF5, and whole-job memory differences are treated as run-to-run variation; this PR does not claim improvements in those stages. The measured postprocessing improvement is the `120.14 s` to `36.92 s` reduction, together with removal of the global boundary gather and rank-zero surface reconstruction from the reaction path.

## Validation

- Focused serial geoid suite: `13 passed`.
- Focused four-rank MPI geoid regression: passed on macOS and Gadi OpenMPI.
- Gadi production validation: 192 ranks, `cellsize=1/64`, exit status 0.
- The MPI rank-consistency assertion uses near-machine-precision tolerance because collective reductions can differ in their final floating-point bits across MPI implementations.

## Files changed

- Add the reaction-integral kernel and Stokes wrapper.
- Add opt-in reaction projection to generic geoid postprocessing.
- Document when to use pointwise recovery versus weak projection.
- Add focused serial and MPI regressions.


lmoresi:
Reading this while finishing #633, which lands in the neighbouring code (`_desmear` / `boundary_normal_traction`). Two notes, both supportive.

**The direction is right, and it is the same lesson from the other side.** #633 turned on the fact that the nodal reaction is a *dual* object — an integrated force. Asking it for pointwise values is the ill-posed direction, and that is where every failure mode lives: the zero-mean P2 vertex basis (#414), the singular vertex rows of the P2 triangle mass, the chord-versus-arc geometry error (#431). `boundary_normal_traction_integral` never asks that question, so none of them apply to it. For a coefficient it is strictly the better construction.

The two changes are complementary rather than overlapping: this adds a new function and leaves `boundary_normal_traction` and `_desmear` untouched, while #633 improves the pointwise recovery that `dynamic_topography_field` still needs for a nodal field. The only collisions are adjacent text in `rotated_bc.py`, the `.pyx` and the changelog.

**On the workflow this replaces — the `centroid` path is worth retiring rather than keeping as the default.** Filed separately as #647 with measurements. `_project_spherical_harmonic_samples` gathers the boundary to rank zero and re-derives a triangulation with `ConvexHull(..., "QJ")`, when the boundary facets already exist and `uw.maths.BdIntegral` already integrates over them via `DMPlexComputeBdIntegral`. That path cannot work on a non-convex boundary at all, joggles its input, and applies analytic spherical triangle areas to a faceted FE reaction.

Measured on the Zhong l=2 shell, relative error against the analytic coefficients:

| cellSize | 0.25 | 0.20 | 0.16 | 0.13 |
|---|---|---|---|---|
| surface, ConvexHull | 0.0028 | 0.0001 | 0.0014 | 0.0013 |
| surface, BdIntegral | 0.0003 | 0.0014 | 0.0005 | 0.0008 |
| CMB, ConvexHull | 0.0041 | 0.0081 | 0.0059 | 0.0027 |
| CMB, BdIntegral | 0.0091 | 0.0098 | 0.0084 | 0.0037 |
| CMB, pointwise fit | 0.0313 | 0.0236 | 0.0188 | 0.0097 |

To be clear about what that does and does not show: **on accuracy the two integral arms are a wash**, so this is not an accuracy argument. It is that the same answer is available on the mesh we already have, on any boundary shape, without a gather or a joggle. The last row is the separate point — both integral forms beat the pointwise fit by 3-10x at coarse resolution, which independently supports #414.

Given your `projection="reaction"` is better still, the question for this PR is whether `centroid` needs to stay the default, or whether it can be deprecated in the same change.

lmoresi:
CI `test` is failing on this branch. Here is what I could establish, including what I could **not** reproduce.

## The failure

```
tests/test_1120_SLVectorCartesian.py::test_SLVec_boxmesh[mesh1]
>   assert np.allclose(vx_to_eval, anax_to_eval, atol=0.01)
E   assert False
```

**Not a magnitude error.** The two arrays shown are ~3e-18 and ~1e-5, both far inside `atol=0.01`, so `allclose` returning False points at a **non-finite element** somewhere in the middle of the array rather than a numerical discrepancy. Worth checking with `np.isfinite(vec_prof_uw).all()` before chasing tolerances — and note `vec_prof_uw` comes straight from `uw.function.evaluate`, which is #604/#641 territory.

## What rules out the easy explanations

- **The branch is current.** 0 commits behind `development`, 23 ahead. Not a stale base.
- **`development` CI is green** on its recent runs, so the test is not broken upstream.
- **I cannot reproduce it locally.** `test_1120` passes 3/3 on macOS/arm64 both on this branch and on a branch without it. So it is Linux/CI-specific, and I could not bisect it.

## The one thing in the diff I would question

Beyond the harmonic projection, this branch also changes `SolverBaseClass`:

```python
-  _names, _iss, _subdms = self.dm.createFieldDecomposition()
-  sgvec = gvec.getSubVector(_iss[0])
-  _subdms[0].localToGlobal(self.Unknowns.u.vec, sgvec)
-  gvec.restoreSubVector(_iss[0], sgvec)
+  self.dm.localToGlobal(self.Unknowns.u.vec, gvec)
```

Three reasons this is worth a second look:

1. It is in the `else` branch — the **single-field** path — which is exactly what `SNES_Vector` takes, and `test_1120` is a semi-Lagrangian **vector** test. The changed code is on the failing test's path.
2. `self.Unknowns.u.vec` is the **variable's** local vector. The old code mapped it through field 0's subDM; the new code assumes the solver DM's local layout matches the variable's. Where they coincide it is equivalent, and where they do not it writes to the wrong slots — which would produce exactly the ~1e-18 (i.e. never-written) values in the failure.
3. It is not mentioned in the PR description, which describes the harmonic projection. It reads as an opportunistic simplification carried along from the benchmark branch.

**Cheapest decisive experiment:** revert that hunk alone and re-run CI. If it goes green, that is the answer; if not, it is eliminated and the NaN hypothesis above is next.

## A process note, offered rather than insisted on

23 commits reach `development` under a title describing one feature, and they include shared-solver changes. Extracting the unrelated `SolverBaseClass` change into its own PR would make both reviewable and would let the harmonic projection — which I think is the right construction, and which I have argued elsewhere should replace the ConvexHull path (#647) — land on its own merits rather than waiting on an unrelated regression.

Happy to push the revert-and-retest to this branch if you would like, but I did not want to touch someone else's branch uninvited.

lmoresi:
Pushed `fb268de3` to this branch (with @lmoresi's go-ahead) restoring the field-decomposition path, as the cheapest decisive experiment on the red CI.

Only that one hunk is touched. Everything else on the branch — including the harmonic projection this PR is actually for — is untouched.

**Why this hunk.** The failing test is `test_1120_SLVectorCartesian.py::test_SLVec_boxmesh[mesh1]`, a semi-Lagrangian **vector** test, which takes the single-field `else` branch that was simplified. `self.Unknowns.u.vec` is the *variable's* local vector: the original mapped it through field 0's subDM, the direct call assumes the solver DM's local layout matches it. Those coincide on a plain single-field solver, which is why the simplification looks equivalent — where they differ it writes to the wrong slots and the field comes back never-written. That matches the failure, where recovered values were ~1e-18 against an analytic ~1e-5.

**Stated plainly: this is not a confirmed fix.** The failure does not reproduce on macOS/arm64 — `test_1120` passes 3/3 there both with and without the change — so I could not bisect it locally. If CI is still red, this hunk is eliminated and the next candidate is a NaN coming out of `uw.function.evaluate` into `vec_prof_uw`: the two compared arrays both sit far inside `atol=0.01`, so `allclose` returning False implies a non-finite entry rather than a numerical discrepancy (#604, #641).

**Checked before pushing**, since it is your branch:

- `test_1120`: 3 passed
- `test_1070_postprocessing_geoid` (this PR's own suite): 13 passed — the revert does not undo what the branch is for
- `level_1 and tier_a`: **1093 passed, 0 failed**

Revert it freely if the simplification was deliberate and load-bearing for something downstream — in which case the right move is probably to split it into its own PR with a test that pins whatever it fixes, so it does not ride along with the projection work.

lmoresi:
Rebase needed — `development` moved a long way today and this now conflicts.
Scoping it here so the rebase is a known quantity rather than a surprise.

**Five files conflict, and only two are substantive:**

| file | blocks | what moved underneath |
| --- | ---: | --- |
| `src/underworld3/utilities/rotated_bc.py` | 3 | the busiest one |
| `src/underworld3/cython/petsc_generic_snes_solvers.pyx` | 1 | #612's rank-0 JIT compile gate, #668's decomposition destroy |
| `tests/parallel/test_0760_swarm_cache_migration.py` | 1 | #602 added a regression test to the same file |
| `docs/developer/CHANGELOG.md` | 1 | append collision under 2026 Q3 — keep both entries |
| `docs/developer/subsystems/boundary-stress-and-projection-postprocessing.md` | 1 | — |

The rest of the 30 files merge cleanly.

**What landed today that this branch will want to know about**, because two of
them touch the same ground as `boundary_normal_traction_integral`:

- **#617** — `Stokes_Constrained.traction()` now returns `h + r(n·u − g)`, the
  whole boundary traction. `multiplier()` returns `h` alone and is short by the
  augmented-Lagrangian share.
- **#653** — `boundary_flux` now refuses a multiplier-constrained boundary
  rather than returning a plausible constant, and points the caller at
  `traction()`.
- **#602** — merged, so the non-finite recovery in `global_evaluate_nd()` is
  already on `development`; the conflict in `test_0760` is just two tests
  appended to the same place.
- **#612**, **#674**, **#680**, **#648** — all merged today.

I have not resolved these. A 30-file branch that reworks the rotated-reaction
path is yours to rebase — the two `rotated_bc.py` and `.pyx` conflicts need
someone who knows which side is intended, and guessing would be worse than
waiting.

Everything else of yours is in: #648, #674, #602, plus #667 and #668 earlier.

gthyagi:
Superseded by #686: https://github.com/underworldcode/underworld3/pull/686

@lmoresi, the remaining spherical reaction-projection changes are now isolated on `feature/spherical-reaction-projection`, based on current development `9da04b72`. The replacement contains one commit (`fd0a3bcc`) and five files, with no SUPG/SLCN, checkpoint, evaluator, or shared-solver changes. It reuses the generic reaction-integral API already merged through #648.

Validation was run against an isolated build of the new branch: 13 geoid tests and all 3 vector-SL cases passed; the spherical reaction/geoid regression passed on all four MPI ranks. The replacement PR distinguishes this fresh validation from the historical Gadi benchmark measurements.

Closing this broad PR in favour of #686. Its original description and discussion are preserved, and the mantle-convection development branch remains intact.

Issue #647, opened 2026-08-26
geoid harmonic projection re-derives the surface mesh with ConvexHull instead of integrating over the boundary facets we already have


lmoresi:
`geoid._project_spherical_harmonic_samples` (`postprocessing/geoid.py:240`) computes the
harmonic coefficient by gathering every boundary coordinate and value to rank zero,
discarding the mesh connectivity, and **re-deriving a triangulation of the point cloud**
with `scipy.spatial.ConvexHull(unit_coords, qhull_options="QJ")`. It then integrates with
analytic *spherical* triangle areas and one-point centroid values.

The boundary facets already exist — it is the FE trace, and `_desmear` walks exactly
those facets a few modules away. `uw.maths.BdIntegral` already integrates over them
properly, through `DMPlexComputeBdIntegral`, with the real quadrature and no gather.

The coefficient is then just the projection onto the harmonic written in the FE inner
product the assembler already uses:

```python
numerator   = uw.maths.BdIntegral(mesh=mesh, fn=h.sym[0] * legendre, boundary=b).evaluate()
denominator = uw.maths.BdIntegral(mesh=mesh, fn=legendre**2,        boundary=b).evaluate()
coefficient = sign * numerator / denominator
```

### What is wrong with the current path

- **It cannot work on a non-convex boundary.** A convex hull of the boundary nodes is
  only the boundary for a convex surface. An annulus arc, an internal boundary, or a
  deformed free surface will silently get the wrong triangulation.
- **`QJ` joggles the input** — it perturbs coordinates to break degeneracies. That is a
  poor foundation for a reproducible diagnostic.
- **It mixes measures.** Analytic spherical triangle areas are applied to a faceted FE
  reaction; the two disagree at O(h²), which is the chord-versus-arc family of #414/#431.
- **It gathers the whole boundary to rank zero** and triangulates there — the same
  scaling objection as #408.
- Centroid value = mean of three nodal values is a one-point quadrature.

### Measured

Zhong l=2 shell, same solves, coefficient three ways, relative error against the analytic
0.41920 (surface) / 0.77060 (CMB):

| cellSize | 0.25 | 0.20 | 0.16 | 0.13 |
|---|---|---|---|---|
| surface, ConvexHull | 0.0028 | 0.0001 | 0.0014 | 0.0013 |
| surface, BdIntegral | 0.0003 | 0.0014 | 0.0005 | 0.0008 |
[...]

PR #648, opened 2026-08-26
Add cylindrical annulus geoid postprocessing

gthyagi:
## Summary

- add the Simons cylindrical sheet Green function for positive Fourier modes
- assemble two-boundary annulus potential, geoid, and self-gravity coefficients
- add an MPI-safe adapter for completed rotated-free-slip Stokes solves
- document sign, density, Fourier, unit, and n=0 gauge conventions

The coefficient API is independent of the physical-space Poisson meshes and convergence drivers.

## Kramer-Simons annulus validation

The fixed nondimensional case uses Fourier mode n=2, radii ri=1.22, rload=2.0, ro=2.22, G=1, unit boundary density contrasts, unit internal sheet coefficient, and unit reference gravities. Consequently geoid N and potential Phi have the same numerical coefficient. Errors below use abs(numerical-reference)/abs(reference).

### Spectral response with varying Stokes h

Here h is the target annulus Stokes element size. Each row recovers dynamic topography from that Stokes solve and applies the cylindrical spectral operator. The reference is the fully analytical Kramer traction plus Simons gravity response.

| h | Mantle cells | Outer N spectral | Outer N analytic | Outer rel. error | Inner N spectral | Inner N analytic | Inner rel. error |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 1/4 | 476 | +1.840897156 | -0.663601170 | 377.410173% | +0.729297687 | -0.502494144 | 245.135559% |
| 1/8 | 1,684 | -0.396340543 | -0.663601170 | 40.274285% | -0.373554174 | -0.502494144 | 25.659995% |
| 1/16 | 6,578 | -0.645005977 | -0.663601170 | 2.802164% | -0.494736091 | -0.502494144 | 1.543909% |
| 1/32 | 26,376 | -0.662116990 | -0.663601170 | 0.223656% | -0.502457968 | -0.502494144 | 0.007199% |

The coarse 1/4 and 1/8 errors are dominated by recovered Stokes traction/topography. Because boundary topography and the internal load substantially cancel, the under-resolved 1/4 topography also changes the total-geoid sign. The converged 1/32 result has 0.224% outer and 0.0072% inner error.

### Physical Poisson response with varying gravity h

This comparison isolates the Poisson discretization: every row uses the same numerical topography coefficients from the 1/32 Stokes solution, while only the P2 gravity mesh spacing h changes. The scalable gravity domain has one radial buffer layer on each side of the mantle: rmin = ri - h and rmax = ro + h. Exact regular/decaying single-mode Robin continuation is imposed at rmin and rmax. The spectral columns use exactly the same topography and source coefficients as the corresponding Poisson solve.

| h | Gravity cells | Potential DOFs | Radial extent | Outer Phi Poisson | Outer Phi spectral | Outer rel. error | Inner Phi Poisson | Inner Phi spectral | Inner rel. error |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1/4 | 662 | 1,414 | 0.97000-2.47000 | -0.660430832 | -0.662116990 | 0.254662% | -0.491171097 | -0.502457968 | 2.246331% |
| 1/8 | 2,040 | 4,254 | 1.09500-2.34500 | -0.661847328 | -0.662116990 | 0.040727% | -0.499466553 | -0.502457968 | 0.595356% |
| 1/16 | 7,282 | 14,912 | 1.15750-2.28250 | -0.662076818 | -0.662116990 | 0.006067% | -0.501710188 | -0.502457968 | 0.148824% |
| 1/32 | 27,772 | 56,238 | 1.18875-2.25125 | -0.662107296 | -0.662116990 | 0.001464% | -0.502269626 | -0.502457968 | 0.037484% |

The inner-interface error is approximately second order under each halving of h. At h=1/32 the one-layer domain uses 27,772 cells, only 18.73% of the 148,288-cell full-disc reference, while giving smaller errors at both interfaces.

### Poisson buffer-width and exterior-radius study

At fixed h=1/32, the gravity-domain padding was varied independently. The first three rows extend the physical annulus by h, 2h, or 3h on both radial sides; the last row is the legacy full disc from the origin to exterior radius 4.44. Every row uses the same P2 discretization, 1/32 Stokes topography, and spectral targets: outer Phi=-0.662116990 and inner Phi=-0.502457968.

| Domain extension | Inner truncation | Exterior radius | Gravity cells | Outer Phi Poisson | Outer rel. error | Inner Phi Poisson | Inner rel. error |
|---:|---:|---:|---:|---:|---:|---:|---:|
| h | 1.18875 | 2.25125 | 27,772 | -0.662107296 | 0.001464% | -0.502269626 | 0.037484% |
| 2h | 1.15750 | 2.28250 | 29,740 | -0.662107026 | 0.001505% | -0.502269395 | 0.037530% |
| 3h | 1.12625 | 2.31375 | 31,856 | -0.662106778 | 0.001542% | -0.502268883 | 0.037632% |
| Full disc | 0.00000 | 4.44000 | 148,288 | -0.662102946 | 0.002121% | -0.502262510 | 0.038901% |

One h of padding is sufficient: increasing to 2h or 3h changes the interface coefficients only slightly, while the h mesh is also more accurate than the full 0-4.44 disc. At h=1/32 it uses 18.73% of the full-disc cells.

## Validation

- 89 relevant serial Kramer, Zhong, spherical-geoid, and cylindrical-geoid tests passed
- 20 tests passed per rank in a two-rank MPI run
- all 12 outputs matched the frozen Kramer-Simons reference at the stored precision
- Kramer Stokes, Poisson geoid, and physical self-gravity passed their standard serial 1/4 and eight-rank 1/8 runs
- the 1/32 extension study and all three retained convergence checkers passed
- Black, mypy, deprecated-pattern, and diff checks passed
- Sphinx documentation build succeeded
- repository Level-1 run reached 1636 passed, 35 skipped, and 2 xfailed before an unrelated timeout-reliability test left its divergent subprocess running

## Scope and limitations

- Fourier modes n >= 1 are supported
- n = 0 is rejected explicitly because its logarithmic solution requires a potential gauge
- tests begin at tier C while the API receives review


lmoresi:
Reviewed. **The physics and the validation are solid; my concern is architectural, and it is cheap to address now and expensive later.**

## What is good

The Kramer–Simons convergence campaign is real evidence, not a smoke test: 0.224% outer and 0.0072% inner at h=1/32, with clean second-order behaviour on the Poisson side under each halving. The input validation in `cylindrical_cosine_boundary_coefficient` is careful — shape, finiteness, minimum sample count all checked with specific messages. The buffer-width study justifying the one-layer gravity domain (27,772 cells against a 148,288-cell full-disc reference, with *smaller* errors at both interfaces) is a genuinely useful result on its own.

## The concern: this adds a second bespoke projection path

`cylindrical_cosine_boundary_coefficient` gathers boundary samples to rank zero, sorts them by `arctan2`, and integrates with a hand-rolled trapezoid. That is the 2-D analogue of `_project_spherical_harmonic_samples`, which #647 argues should be retired: gather the boundary, re-derive an ordering the mesh already knows, integrate with bespoke quadrature instead of the FE boundary integral.

Three consequences worth weighing:

1. **It assumes the boundary is a circle.** Sorting by θ and trapezoid-integrating is exact only for samples on a circle. A deformed annulus — a free surface, the case this machinery ultimately serves — breaks it, in the same way convexity limits the spherical path.
2. **Chord versus arc.** Integrating in θ over a faceted boundary is the #414/#431 error family again, applied to an FE reaction.
3. **It gathers to rank zero** (`MPI.COMM_WORLD.gather`), which is the scaling objection in #408, and uses `COMM_WORLD` rather than the mesh communicator.

## The alternative is already in your own #646

`boundary_normal_traction_integral` contracts the assembled reaction with the test function over owned DOFs plus MPI reductions — no gather, no ordering, no pointwise recovery. For a coefficient that is strictly the better construction, and it is the one I argued in #647 should replace the spherical path. `uw.maths.BdIntegral` is the other UW-native option where a nodal field genuinely exists.

Routing this projection through one of those is a change to one function, not a rewrite. Doing it before merge avoids maintaining two bespoke paths and then migrating both.

## Practical collisions if both land

- #646 and this PR both modify `src/underworld3/postprocessing/geoid.py` — expect a conflict.
- Both add a `test_1071_*` file (`tests/parallel/…spherical_shell_geoid_parallel` vs `tests/…postprocessing_cylindrical_geoid`). Different directories, same number; worth renumbering one.

## Verdict

Not blocking on correctness — CI is green and the validation stands. But I would rather see the projection go through #646's integral (or `BdIntegral`) than merge a second gather-and-trapezoid path that #647 then has to unwind twice.

gthyagi:
Implemented the requested distributed finite-element boundary projection in commit `b30a02bc`.

### What changed

- Added `Stokes.boundary_normal_traction_integral(boundary, fn, remove_mean=True)`.
- The new method contracts the assembled rotated-free-slip reaction directly with the scalar test function at velocity interpolation nodes.
- Only owned reaction DOFs are counted, followed by an `allreduce` on the mesh communicator.
- Constant normal-traction gauge removal uses the finite-element boundary integrals of `1` and `fn`.
- Numeric test functions are converted to SymPy expressions before evaluation.
- The cylindrical-annulus adapter now projects against `cos(n theta)` using this reaction functional and divides by the matching `BdIntegral(cos(n theta)^2)` norm.
- Removed pointwise traction recovery, boundary-sample deduplication, angular sorting, `MPI.COMM_WORLD` gathering, and rank-zero trapezoidal projection from the finite-element adapter.
- Retained `cylindrical_cosine_boundary_coefficient()` for genuinely external sampled circular data.
- Renamed the cylindrical geoid test from 1071 to 1073 to avoid the existing development-branch test-number collision.
- Updated the subsystem documentation and changelog.

### Validation

- Cython extension compiled successfully.
- Focused serial suite: **21 passed**.
- Focused two-rank MPI suite: **21 passed on each rank**.
- Existing rotated-free-slip annulus regression: **1 passed**.
- The focused suite includes a real annulus Stokes solve exercising the complete assembled-reaction adapter, not only mocked projection data.
- In the real smoke calculation, serial and two-rank MPI fitted reaction coefficients agreed to approximately `5e-7` absolute.

This makes the annulus topography/geoid projection distributed and tied to the actual finite-element boundary measure, including faceted or deformed boundaries.

Issue #637, opened 2026-08-24
3-D boundary-flux recovery supports only P1/P2 triangular traces, so dynamic topography cannot be cross-validated


lmoresi:
In 3-D, `_desmear` — the primitive behind `boundary_flux`, `boundary_normal_traction`
and `dynamic_topography` — accepts only P1 and P2 **triangular** traces. Anything else
raises:

- `src/underworld3/utilities/boundary_flux.py:327` — non-triangular facets (so any hex
  or prism mesh, whose boundary facets are quadrilaterals)
- `src/underworld3/utilities/boundary_flux.py:331` — a facet carrying interior DOFs
  (so P3 and above)

The 2-D path has no such restriction: it handles arbitrary trace degree, building the
element mass from the actual node parameters along the edge chord.

Two consequences.

**Dynamic topography has exactly one supported discretisation in 3-D.** Velocity must
be P2 on tets. There is no second discretisation to check the first against, so a
recovery error that is a property of the P2 triangle — the vertex-integral checkerboard
the `mass="consistent"` docstring already warns about — cannot be distinguished from a
property of the physics by varying the element.

**It blocked a cross-element control in #633.** The vertex-versus-midpoint split under
the grad-div penalty was to be tested at P3/P2 and on hexes, to see whether it tracked
the node or the polynomial order. Both arms raise, so that discrimination was done on a
2-D annulus instead (`~/+Simulations/topography_penalty_633/`), where the trace degree
is free.

Worth having, in rough order of value:

1. Quadrilateral facets (hexes), P1/P2 — the tensor-product mass matrices are standard
   and it opens a genuinely different element family.
2. P3 and above on triangles — needs the facet-interior node coordinates, which the 2-D
   path already builds via `_trace_interior_coords`.

Neither is needed for correctness of what ships today; both are needed to validate it.

Issue #625, opened 2026-08-20
Stokes pressure sub-solve runs at its 200-iteration cap — the inner Schur solve is silently unconverged and dominates solve time


lmoresi:
With the outer fixed to fgmres (#624), the measured bottleneck of a genuine Stokes solve moves squarely to the pressure sub-solve: fieldsplit_pressure runs fgmres+gasm with ksp_rtol = solver tolerance and ksp_max_it = 200, and the iteration readout sits AT 134-200 per outer application — 200 is the cap, so the Schur solve routinely exits unconverged (the "repeating iteration count is a cap" pattern; converged reason is never checked on the sub-block). On the 85k-cell contrast benchmark (~/+Simulations/place_route_health/) the pressure block accounts for most of the ~285-308 s genuine solve, for both velocity preconditioners (custom-P FMG and GAMG), which is why the velocity-block choice moves the total by only ~8%.

Two directions, not mutually exclusive:
1. Make the inner Schur affordable: a better pressure preconditioner than gasm-on-a11 for discontinuous P0 pressure with variable viscosity (the eta-scaled pressure mass matrix is the standard candidate), so the cap is never reached.
2. Make the inexactness deliberate: if the design intent is an inexact inner Schur under a flexible outer (the CITCOM margin design), the tolerance/cap should say so explicitly and cheaply (far fewer than 200 iterations), instead of paying 200 iterations for an accuracy that is then not achieved anyway.

Either way the current default pays maximum cost for unconverged inner solves, silently.

Underworld development team with AI support from Claude Code

lmoresi:
Element-pairing dependence measured (same 85k-cell mesh, contrast, tolerance, fgmres outer, gamg velocity — only the pressure space changed):

- P2-P0 discontinuous (the contact/fault tests' pairing): genuine solve 307.7 s, pressure block 134 its (and AT the 200 cap during other applications)
- P2-P1 continuous (Taylor-Hood): genuine solve 147.5 s, pressure block 43 its

The mass-preconditioned Schur iteration count scales with the inverse square of the inf-sup constant, and P2-P0's is the weakest of the common stable pairs — the grind is substantially the PAIRING, not the gasm preconditioner alone. Cumulative ledger for the genuine solve since the start of this campaign: 543.9 s (gmres outer, P0disc) -> 147.5 s (fgmres outer, Taylor-Hood), 3.7x, with no change to the operator being solved.

[...]

PR #649, opened 2026-08-26
Rebuild P2 boundary-recovery vertices from the midpoints, and exonerate the grad-div penalty (#633)

### LINKED ISSUE #633: Grad-div penalty breaks POINTWISE traction recovery (28% on dynamic topography); facet-integrated is fine
The grad-div penalty corrupts the recovery of **pointwise** boundary traction from reaction loads, while leaving the facet-integrated value correct. This is the one thing blocking `Stokes.DEFAULT_PENALTY` from being turned on (PR #632, where everything else is in place).

Measured, `test_1018_rotated_freeslip::test_rotated_freeslip_spherical3d_reaction_topography`, spherical shell, rotated free-slip, `penalty = 10`:

| quantity | expected | measured | |
|---|---|---|---|
| `surface` (facet-integrated) | 0.41920 | 0.4192 | pass |
| `surface_vertices` (pointwise) | 0.41920 | **0.30211** | **-28%** |
| `cmb` (facet-integrated) | 0.77060 | 0.7706 | pass |

At `penalty = 0` all of them pass. That test's own docstring is "3D reaction loads must be divided by boundary mass to recover **pointwise** stress", which is exactly the step that stops working.

**Not a `p_mech` bookkeeping error.** The first hypothesis was that `sigma_nn` was being built from the Lagrange multiplier without the documented `p_mech = p - lambda*mu*(div u)` correction. It is not: `boundary_normal_traction` recovers `sigma_nn` from the Cartesian nodal reaction `r_c = A.u - b`, which already carries the penalty because the penalty is in the operator. Moving the term into the total stress (PR #632) left the measured value **byte-identical** at 0.30210742214244796.

**Working hypothesis.** `lambda*mu*(div u)` is non-zero cell-by-cell for P2-P0 — the discrete divergence is only weakly zero, against the pressure space — so it averages out over a facet integral but not at a vertex. The de-smearing from reaction loads to pointwise stress divides by a boundary mass that does not account for the augmented contribution.

**Why it matters.** Dynamic topography is the main product of this machinery, and a 28% error in it is not something to ship silently behind a default. It is also worth knowing independently of the penalty: if the pointwise recovery is sensitive to a term the integral is not, that is a property of the recovery, not of the penalty.

**Suggested first step.** Compare `surface_vertices` against `surface_midpoints` as `lambda` is swept 0, 1, 10, 100 — the test already computes both, and whether they diverge together or separately should distinguish a mass-weighting error from a sampling one.

Context: the penalty is measured to be a strong lever on the Schur count under FMG (Schur iterations per application 59 -> 18, total velocity work 546 -> 270, 21% faster) and the accuracy cost elsewhere is a consistent perturbation — same convergence rate, gap shrinking 1.102 -> 1.087 -> 1.066 over three refinements. So this is the last thing standing between that and a default.

lmoresi:
Closes #633.

## The penalty was innocent

With `stokes.penalty = 10`, spherical dynamic topography recovered from the rotated
free-slip reaction was 28% low **at vertices** while the facet-integrated value stayed
correct. Grad-div augmentation was the obvious suspect. It is not responsible.

The de-smearing mass for a 3-D **P2 triangular** trace has vertex rows summing to exactly
zero. Row sum *i* of a mass matrix is `∫φᵢ`, since `Σⱼ∫φᵢφⱼ = ∫φᵢ·Σⱼφⱼ = ∫φᵢ` for a
partition of unity — so `[0,0,0,60,60,60]` says the P2 triangle vertex basis function has
**zero surface mean**.

That was already known: **#414** documented it and gave the sharper mechanism, which this
PR adopts rather than re-deriving. Because the vertex basis has zero mean, the vertex
reaction carries essentially only the O(h) facet-normal/geometry error, and the consistent
solve *faithfully reconstructs that error*. It is not amplifying noise. What is new here is
the separation from the penalty, and the fix — which is #414's own action item (2),
"consider having `dynamic_topography` on curved boundaries return a midpoint-weighted or
fitted field", previously unactioned.

## How it was isolated

A 2-D annulus — curved, rotated free-slip, `cos(2θ)` internal load — reproduces the sign
pattern exactly but ~30x weaker. Matching its resolution and mass to the shell's closes
most of that gap (12.5% vs 22% at penalty=10, cellSize 0.25). The two then part company
**under refinement**, and that is the discriminating measurement:

| vertex deficit at penalty=10 | | | | |
|---|---|---|---|---|
| annulus, cellSize 0.25 → 0.0375 | 0.1245 | 0.0538 | 0.0091 | 0.0046 |
| shell, cellSize 0.30 → 0.16 | 0.2842 | 0.2550 | 0.2651 | 0.2793 |

The annulus converges ~O(h²); the shell is flat over a 3.2x node-count range. So it is
neither a consistency error nor curvature — the annulus is curved and recovers. The 2-D P2
**line** mass has vertex row sums of 5, which is why 2-D never showed it.

## The fix

`mass="auto"` on a 3-D P2 trace no longer asks for vertex values at all. It takes the
consistent solve, keeps its superconvergent midpoints, and **reconstructs the vertices
from them**: the three midpoints of a facet determine a unique linear function, so a
vertex reads its two adjacent midpoints and subtracts the opposite one, averaged over
incident facets. Worst-node error against the analytic Zhong coefficients:

| cellSize | 0.25 | 0.20 | 0.16 | 0.13 | 0.11 |
|---|---|---|---|---|---|
| surface, `p1` | 0.041 | 0.026 | 0.018 | 0.013 | 0.008 |
| surface, **reconstructed** | 0.016 | 0.012 | 0.012 | 0.003 | 0.004 |
| CMB, `p1` | 0.116 | 0.067 | 0.047 | 0.030 | 0.025 |
| CMB, **reconstructed** | 0.094 | 0.058 | 0.043 | 0.024 | 0.015 |

Better at every resolution on both boundaries, by 1.8x to 4.9x. `mass="p1"` remains
available and unchanged.

It is not only topography: on a **flat** boundary with a *varying* flux (Poisson heat
flux, manufactured harmonic solution) the reconstruction is ~3x better at vertices than
`consistent` — 0.169/0.124/0.084 against 0.459/0.358/0.265 over cellSize 0.45 → 0.22. Note
this sharpens #414's "exact on flat boundaries" claim, which holds for a *uniform* flux
(what `test_1019` checks) but not a varying one.

## The penalty default was flipped and reverted

#633 no longer blocks `Stokes.DEFAULT_PENALTY`, so it was set to 10 on the #625 evidence
and then **reverted in the same branch**. Three tier-A/B tests fail at 10 and pass at 0,
and the same three run **8.3x slower** — 9.27 s to 76.92 s, warm JIT cache both ways. The
#625 win needs an FMG hierarchy; without `refinement>=1` the velocity block falls back to
GAMG, which is where grad-div augmentation drives the solve into its iteration cap. The
default path is the one without a hierarchy, so the default serves it.

Two of those three failures are not penalty defects and are recorded as such:
- `test_1060` (Nitsche free-slip leak 1.234e-4 against 1e-4) — augmentation perturbing a
  *weakly* imposed constraint. A strong rotated constraint is untouched.
- `test_0112` (swarm) — actually **#641**: `evaluate` returns −0.4976 for
  `sqrt((E**2).trace()/2)` at in-domain points near the lid-corner singularity, at
  `penalty=0` as well. The penalty only moved an accumulated total across zero.

## Tests

- `test_1018` refined cellSize 0.25 → 0.13 and tolerances tightened **0.10/0.12 → 0.01/0.05**,
  set from measured error with ~2-4x headroom. Every node class is asserted, not the
  aggregate — the failure was confined to one class and an aggregate assertion passed
  straight through it.
- #414 asked that these goldens *not* be tightened, because they were absorbing the vertex
  bias. That caveat is **discharged, not ignored**: `auto` no longer consumes the biased
  vertex reactions, so the quantity under test changed. The test says so and cites #414.
- Negative control at cellSize 0.13: `consistent` gives 0.31525 at the surface vertices and
  **fails** the new tolerance; the reconstruction gives 0.41604 and passes.
- `test_p2_triangle_mass_has_zero_vertex_row_sums` — a no-solve guard on the identity, so
  the reason stays attached to the choice.
- **tier A+B: 1963 passed, 0 failed** (39:19), on this branch merged up to current
  `development`.

## Scope and what is left

This changes `mass="auto"`, which is what a direct caller of `boundary_flux` /
`boundary_normal_traction` gets. It does **not** touch the free-surface path:
`free_surface.py:146` already translates to `mass="p1"` in 3-D and consumes a P1 field.

Deliberately not addressed here:
- **#637** — 3-D recovery accepts only P1/P2 triangular traces, so there is exactly one
  supported discretisation in 3-D and none of this can be cross-validated against a second
  element. That is the real remaining gap.
- **#647 / #646** — for a *coefficient*, the right construction is @gthyagi's
  `boundary_normal_traction_integral`, which contracts the reaction with the test function
  and never forms a pointwise field. This PR improves the pointwise default; it does not
  compete with that.

Measurements and scripts: `~/+Simulations/topography_penalty_633/`.


lmoresi:
## Adversarial review of this PR

Posting the case against it, since nobody else has the context to and it should not merge on the strength of its own description.

**The headline mechanism was not discovered here, and the first version of this branch overclaimed it.** #414 had the zero-mean P2 vertex basis, the drift-away-under-refinement, and the recommendation that became the fix. The commit history shows the correction (`e8dc357c`) rather than hiding it. The genuine contribution is narrower than the title suggests: separating the effect from the grad-div penalty, and implementing an action item that had sat unactioned for a month.

**The reconstruction inherits its safety from the midpoints.** `v = m_a + m_b - m_c` amplifies noise up to 3x by construction. It is only sound because the midpoints are superconvergent. On an element whose midpoints are not, this is worse than useless — so #637 must not be closed by generalising this by analogy to quads or P3. That reasoning would be invalid.

**`test_1018` measures a coefficient by fitting pointwise values**, which is exactly what #414 and #646 both argue you should not do for a coefficient. The tolerances are now tight against a metric that is itself the wrong construction. It would be better computed with @gthyagi's `boundary_normal_traction_integral`, and I have left it alone only to avoid entangling this PR with #646.

**The tolerance tightening overrides an explicit request in #414 not to tighten these goldens.** The justification — that `auto` no longer consumes the biased vertex reactions, so the quantity under test changed — is in the test docstring and should be checked by a reviewer rather than taken on trust. If you disagree with it, the tolerances are the thing to revert, not the recovery.

**The CMB is the weak arm throughout.** Reconstructed worst-node error is 0.015 at cellSize 0.11 against 0.004 at the surface, and `rtol=0.05` carries only ~2x headroom at the test's resolution. That is the assertion most likely to become flaky on a different gmsh version.

**The penalty revert rests on a small sample.** The 8.3x figure is three tests — the three that failed, so selected for penalty sensitivity. It was measured back-to-back with a warm cache and I believe it, but it is not a suite-wide benchmark, and the earlier attempt at one was contaminated and discarded rather than repeated.

**What I am confident in:** the zero row sums (exact, arithmetic); the h-independence of the consistent vertex error over five resolutions on the shell against convergence on the annulus; the negative control, which fails the new tolerance with the old recovery; and tier A+B at 1963 passed / 0 failed.

PR #658, opened 2026-08-27
Fault networks: the network mesher, one specification with two realisations, and the junction glue

lmoresi:
## What this adds

A fault network is specified once (traces, hierarchy, width) and realised either as split-node cuts or as a volumetric weak plane on the **same mesh**, with the junction glue derived from the mesh itself.

**Meshing**
- `place_thin_volume(mesher="network")`: every strand of a network placed in one fused call with embedded spines, so strands may touch and every cut walks its own vertices at any width.
- A placed mesh owns its multigrid hierarchy (`adopt_hierarchy`); the split child inherits it, so FMG / FAC is automatic on every solver.
- The 2-D cavity fill grades from the band's rung size to the far field.
- Collinear abutting pieces are placed on **one spine** (two ribbons along one line interleave their vertices into slivers), cut at their own ends; `place_fault_ribbon_2d` takes `spines=` separately from the traces and a per-end tip margin.

**`FaultNetwork`**
- `build(width=, realisation="split" | "ti")`: one call, one mesh, either realisation; `apply()`, `ti_fields()`, `slips()` dispatch on it.
- `junction_cells()`: the places the split cannot join, read off the mesh — band cells off a spine's cut (a tip margin, a gap on a shared spine) where two pieces' ribbons meet, dilated one vertex ring. Free tips are excluded on purpose.
- `junction_patch(eta_0, ratio=0.01)`: the glue — weak isotropic material on those cells, a viscosity ratio rather than a yield stress. No stress scale, linear solve.
- `build()` extends the tip margin at a prepared junction end until the ribbon reaches the other piece's cut; `prepare()`'s near-miss pull-back opens a join to one ligament, not three.

**Measured** (S-fault rig, two resolutions, against the two end members "the fault is longer" and "the fault is cut"): the patch recovers 0.8–1.0 of a continuous fault's transmission across a stepover; the slip crosses on the cut itself; the rest of the network keeps the split's answer within 2.5 %; the weak patch matches a fully plastic patch to 1–2 % and is insensitive to the ratio over two decades; velocity cost is the split's. The bare junction cells without the ring recover a quarter at most — the weld's stiffness is the ring of intact material around the tips. The shipped path reproduces the rig's own placement to three figures at both resolutions and in both realisations.

**Docs and tests**: `docs/advanced/fault-networks.md` rewritten around the end-member framing; `test_0857` (network mesher), `test_0858` (realisations), `test_0859` (junction glue, shared spine, near-miss). Regression over the placer, toolkit, prep, 3-D and stress-glyph tests: 63 passed.

## Known limits
- `junction_cells` is a 2-D ribbon rule; 3-D networks keep `damage_yield` (the tube glue).
- On a graded base at fine width the carve can fail to leave one simple cavity at the default `carve_clearance` (`TODO(BUG)` at the check; 1.0 works where 0.3 does not, or build on the uniform base).
- The glue tests are serial; the ring dilation is rank-local.


lmoresi:
## Adversarial review

We read the diff against `development` (11 files) looking for what would break first.

**Findings**

1. **Parallel is unverified.** `junction_cells` and `_vertex_ring` work on rank-local cells and DMPlex closures; `test_0859` skips under MPI. A junction straddling a partition boundary gets its ring from whatever overlap the rank holds. The geometric part of the rule is deterministic per cell, so the two ranks agree on the un-dilated set; the ring is the exposure. Needs a `ptest` on a two-piece network cut across ranks before the glue is called complete.
2. **`_shared_spines` only chains head-to-tail.** A piece digitised in the opposite direction (its END facing the other piece's END) is not grouped, and the sliver storm this PR removes comes back for that input. Evidence: the grouping test uses `B[0] - A[-1]` only. The fix is to test both orientations and reverse the follower.
3. **Two reaches that should be one.** `_junction_margins` extends a spine until it reaches the other piece's cut, but the overlap count in `junction_cells` continues each cut piece by a fixed two rings. A ligament longer than two rungs plus half a width is reached by the spine and missed by the count (`prepare(ligament=2)` sits at that edge). The count should use the same reach as the margins.
4. **Memory of the brute-force distances.** `junction_cells` and the footprint match form `cells × samples × dim` arrays. At rig scale that is nothing; at 10^5 band cells against 10^4 densified samples it is gigabytes. A KD-tree or chunking is due before 3-D networks use this rule.
5. **Behaviour change in `prepare()`.** The near-miss pull-back was a whole ligament from each end; it is now the clearance deficit, shared. Existing users get smaller joins (this is what the docstring always promised). `test_0849`/`test_0850` are unchanged and pass, but anything tuned to the old join size will move.
6. **Graded base + fine width fails to carve** at the default `carve_clearance` (the `TODO(BUG)` at the check). The docstring gives the remedy; the placer should size the clearing by the local cell size.

**Checked and fine**: the shipped path reproduces the S-fault rig's own placement to three figures at coarse and fine, split and TI; the merge with `development` resolved `place_surface.py` to this branch's version, verified identical to the #638 squash's base; 63 tests pass post-merge.

**Verdict**: mergeable as the 2-D production path, with 1–3 as the follow-up issues to file before it is used on a distributed mesh or a network with mixed digitising direction.

PR #662, opened 2026-08-28
Fault networks in parallel: co-located multigrid tail, gather-safe pair diagnostics

lmoresi:
## What this fixes

A placed fault mesh is gathered onto its surgery rank while the multigrid tail stays load-balanced. The geometric transfer then pairs fine nodes with coarse cells on other ranks: on the S-fault rig at np=2, 488 of 5614 coarse DOFs on one level had no fine image, the build fell through to the RBF rescue, and repairing that many columns leaves a preconditioner that does not converge.

- `adopt_hierarchy` now **co-locates each coarse level with the finest** (coarse cell → the rank holding the fine cells over it; shell partitioner) — ptest_0004's co-resident construction. The fine mesh and its FAC patch never move; only the small coarse levels do. Rig at np=2: barycentric transfer at serial sparsity, the serial 14 iterations, wall time equal to serial.
- The fallback messaging claimed a dense RBF transfer; `rbf_prolongation` has been the sparse local kd-tree interpolator since #429. Corrected, and a parallel counterpart of the serial zero-column repair added for the cross-partition build.
- `fault_pair_jumps(gather=True)` all-gathers the rank-local pairs, for diagnostics that go on to make collective calls.
- `ptest_0859_fault_network_parallel`: the network built, glued and solved at np=2 gives the serial slips and records no preconditioner fallback.

Includes the merge of #656 (the empty-rank `evaluate` deadlock, #611): with it, the rig's full diagnostic pass runs at np=2 to the serial answer. **Merge #656 first**; this branch then adds only the files above.

Serial custom-MG suite: 263 passed. Parallel: ptest_0004, 0845, 0852, 0855, 0859 pass at np=2.

## Not in scope
No rebalancing of the fine mesh (the placement stays gather-first), no GAMG path. A fine mesh gathered on one rank gives no speed-up; that is the placement policy, not this PR.


lmoresi:
## Adversarial review

We read the diff for what would break first.

1. **Co-location moves the tail, and the tail is shared.** `adopt_hierarchy` now replaces the coarse `Mesh` objects with redistributed copies. Anything holding the ORIGINAL tail meshes (a solver on the base mesh, a checkpoint that reloads the hierarchy, a second placed mesh adopting the same base) sees a different partition from the placed child's. Each placed child gets its own copies, so nothing is mutated in place, but two children of one base now carry two tails; memory doubles per child in parallel. Acceptable at this scale; worth a note in `adopt_hierarchy`'s docstring.
2. **Ownership by nearest centroid is a heuristic.** A coarse cell straddling the partition seam is assigned to whichever rank's owned fine centroid is nearest; a fine node inside that coarse cell can still be owned by the other rank, and the co-partitioned build then misses it — the `"auto"` cross-partition rebuild covers that case, so correctness holds, but the fast path is not guaranteed. The rig and the toy showed no such miss at np=2; np=4 on a 3-D placement is the case to try.
3. **The parallel zero-column repair is now dead weight on the happy path.** With co-location the barycentric build has no orphans, so `_repair_zero_columns_parallel` only fires on a genuinely non-nested pair. We kept it because the serial path has the same repair and the guard after it still refuses loudly; if it fires with hundreds of columns the level pairing is wrong, not the columns (that is exactly what it looked like before co-location).
4. **`gather=True` is opt-in.** `FaultNetwork.slips` and the rig gauges now gather, but any other caller of `fault_pair_jumps` that goes on to a collective call still hangs on the empty rank. The docstring says so; a grep of callers found none outside the rig and the network.
5. **#656 is inside this branch.** Merge #656 first; this PR then reduces to `custom_mg.py`, `fault_contact.py` and the new ptest.

Checked: serial custom-MG suite 263 passed; ptest_0004/0845/0852/0855/0859 at np=2; the S-fault rig at np=2 runs gauges and checkpoint to the serial partition.

**Verdict**: mergeable after #656, with 2 as the follow-up to measure at np=4 in 3-D.

PR #663, opened 2026-09-01
3-D fault networks in parallel: one redistribution for the whole network

lmoresi:
Sequential `split_fault` calls refused multi-fault networks in parallel — each split redistributes, and a prior pairing cannot migrate. `split_faults(mesh, names)` performs the 2-D `add_fault` pre-pass for already-labelled faults: one redistribution keyed on the union of the network's facets, then every split runs with serial topology. The seam test now covers a facet's full closure (3-D corners), and `split_fault` redistributes only when the seam is actually touched.

Measured (test_0851's crossing patches): np=1/2/3 identical slip peaks on a genuinely distributed mesh. New `ptest_0851_fault_network_3d_parallel`; ptests 0845/0848/0852/0855/0859 pass at np=2; serial fault suite 125 passed.

Toward 3-D parity, still open (tracked in the workflows notes): the volumetric width/TI realisation in 3-D, a 3-D junction-cells rule (the tube glue stands meanwhile), parallel FAC patch smoothing, the np=4 co-location measurement.


lmoresi:
## Adversarial review

We read the diff for what would break first.

1. **The closure-based seam test is stricter in 2-D too.** A chain vertex on the seam was already caught via the cone; the closure adds nothing in 2-D, so behaviour is unchanged there — verified by the 2-D ptests (0845, 0855, 0859) and the 125-test serial suite. In 3-D it is the correctness fix: a patch corner on the seam previously escaped the test while `split_fault` compensated by always redistributing.
2. **`split_fault` no longer redistributes unconditionally in 3-D.** A single 3-D fault whose star straddles the seam still redistributes (the seam test fires); one that does not touch the seam now skips a move it never needed. The risk case is a fault whose star crosses the seam WITHOUT any labelled facet's closure being shared — we could not construct one (the star is the closure's cells), but if it exists the split's own rank-interior gates refuse loudly rather than corrupting.
3. **`split_faults` requires the names to be in `mesh.boundaries`.** True for the embedded patches and the placed sheets; a caller with a label not in the enum gets a KeyError, not a message. Acceptable for an internal-facing util; a docstring line says the contract.
4. **The union redistribution gathers every fault star in one move.** For a network whose combined star is a large fraction of the mesh, the imbalance is the placement policy's (gather-first), not this PR's — but the one-move form concentrates ALL stars on one rank where per-fault moves might have spread them. That matches the 2-D contract and the pairing constraint; a per-fault target assignment inside one redistribution is possible later if the imbalance matters.
5. **The pairing-migration refusal is retained** for the genuine case (a split that must redistribute while prior pairs exist), so nothing silently corrupts if a caller bypasses `split_faults`.

Checked: np=1/2/3 identical slips on the crossing network, distributed cells; the five parallel ptests and 125 serial fault tests pass.

**Verdict**: mergeable; 2 is the one theoretical hole and it fails loud, not wrong.

Issue #644, opened 2026-08-25
FMG geometric-MG coarse solve uses PCREDUNDANT: coarse LU replicated on every rank — OOM at scale when the base mesh is fine


bknight1:
## Problem

When using FMG with the default `redundant` coarse solver (`pc_type=redundant` + `pc_factor_mat_solver_type=lu`), the coarse-grid LU factorization is **replicated on every MPI rank**. For models with large base meshes (e.g., 10 km cells, 800x230 km domain), the coarsest level still has ~14k velocity DOFs, causing:

- **~300 MB/rank replicated memory** from the coarse LU factorization (fill ratio 5.16× on 14k DOFs)
- **Total ~9 GB equivalent** replicated across 30 ranks
- **OOM kills at 900 MB/rank** cap on Setonix/Pawsey

## Root Cause

The `multigrid_options.py` default `geometric_mg_bundle()` sets `mg_coarse_pc_type=redundant` (see `multigrid_options.py:262`). This is documented as "(small) coarse system" but fails when base mesh resolution is high.

Related: PR #213 (`_use_iterative_solver`) already avoids MUMPS in parallel for the same reason, using `redundant`+`svd` instead — but `svd` is infeasible at this scale (dense 14k×14k).

## Solution Found

Using a finer base mesh with more refinement levels keeps the **finest grid identical** while shrinking the coarsest level:

```bash
# Original (OOM)
-uw_csize_km 10 -uw_refinement 2   # 14k coarse DOFs → ~300 MB/rank replicated

# Fixed (fits in 900 MB/rank cap)
-uw_csize_km 20 -uw_refinement 3   # ~1.9k coarse DOFs → divided / minimal replication
```

| Config | Coarsest vDOFs | Linear solve time (serial) | Peak RSS/rank (np=30) |
|--------|----------------|----------------------------|----------------------|
| `csize10 ref2` | 14,195 | ~100 s | **683 MB** ❌ OOM at 900 MB cap |
| `csize20 ref3` | ~1,900 | **111 s (+11%)** | **416 MB** ✅ |

- Finest grid essentially unchanged (233k vs 227k velocity DOFs)
- Trade-off: **+11 % linear-solve time buys −39 % per-rank peak RSS**
- Causal check: shrinking *only* the coarsest level removed **261 MB/rank**, matching the redundant gather + LU-fill arithmetic (~45 MB gathered copy + 232 MB factor ≈ 280 MB)

## Proposed Fix Options

[...]

Issue #643, opened 2026-08-25
FMG + redundant coarse solver: replicated memory scales with base mesh size, causing OOM at HPC scale


bknight1:
## Problem

When using FMG with the default `redundant` coarse solver (`pc_type=redundant` + `pc_factor_mat_solver_type=lu`), the coarse-grid LU factorization is **replicated on every MPI rank**. For models with large base meshes (e.g., 10 km cells, 800x230 km domain), the coarsest level still has ~14k velocity DOFs, causing:

- **~300 MB/rank replicated memory** from the coarse LU factorization (fill ratio 5.16× on 14k DOFs)
- **Total ~9 GB equivalent** replicated across 30 ranks
- **OOM kills at 900 MB/rank** cap on Setonix/Pawsey

## Root Cause

The `multigrid_options.py` default `geometric_mg_bundle()` sets `mg_coarse_pc_type=redundant` (see `multigrid_options.py:262`). This is documented as "(small) coarse system" but fails when base mesh resolution is high.

Related: PR #213 (`_use_iterative_solver`) already avoids MUMPS in parallel for the same reason, using `redundant`+`svd` instead — but `svd` is infeasible at this scale (dense 14k×14k).

## Solution Found

Using a finer base mesh with more refinement levels keeps the **finest grid identical** while shrinking the coarsest level:

```bash
# Original (OOM)
-uw_csize_km 10 -uw_refinement 2   # 14k coarse DOFs → 300 MB/rank

# Fixed (fits in 900 MB cap)  
-uw_csize_km 20 -uw_refinement 3   # 1.9k coarse DOFs → 100 MB/rank
```

| Config | Coarsest vDOFs | Linear solve time (serial) | Peak RSS/rank (np=30) |
|--------|----------------|----------------------------|----------------------|
| `csize10 ref2` | 14,195 | ~100 s | **683 MB** ❌ |
| `csize20 ref3` | ~1,900 | ~111 s (+11%) | **416 MB** ✅ |

Finest grid DOFs nearly identical (233k vs 227k). Linear solve time increases ~11%.

## Proposed Fix Options

1. **Size-aware default in `multigrid_options.py`** (preferred):
   - Query coarsest-level DOFs from `mesh.dm_hierarchy`
   - Use `redundant+lu` only if coarsest DOFs < threshold (~5k)
   - Above threshold: `gamg` or `asm` (distributed, no replication)

2. **Document the workaround** (immediate):
[...]

Issue #622, opened 2026-08-20
Custom-P V-cycles across a cut level cost more than they save (7x slower than GAMG at 11x fewer iterations)


lmoresi:
With the cut mesh inheriting the adapt tail (add_conforming_sheet, PR #589), the velocity-block geometric MG installs and converges in 4 iterations against GAMG's 46 — and still loses the wall time 7x. Instrumented A/B on an 85k-cell cut child (same mesh, operator, tolerance; warm solves, setup excluded; scripts in ~/+Simulations/place_route_health/):

- inherited tail: warm 60.2 s, velocity block mg, 4 its (~13 s per preconditioner application)
- stripped:       warm  8.5 s, velocity block gamg, 46 its

The finest transfer crosses the CUT: the cut child sets _refine_dofs_coincide=False (a cut vertex is not an edge midpoint), so the exact nested prolongation cannot apply and the transfer comes from the geometric builder. The #576/#579 FMG wins (2.2x smooth, 13x banded) were measured on UNCUT adapt children with exact transfers. A cheap sparse transfer across a cut level is the missing piece before the inherited tail pays for itself; until then GAMG remains the right default on cut children, and the tail's value is that split-aware or exact-transfer improvements have somewhere to land.

Related: #424 (dense RBF transfer cliff). Also observed: cold solves at this size are setup-dominated (~700-750 s of JIT/assembly against 8.5-60 s of Krylov), so preconditioner comparisons must be made on warm solves.

Underworld development team with AI support from Claude Code

lmoresi:
Decomposed with a third timed solve (same run, operator unchanged): cold 784.7 s, warm 60.9 s, warm2 60.5 s. The ~52-55 s custom-P premium recurs on EVERY solve — it is per-solve SETUP (the geometric transfer build: re-triangulating the coarse DOF cloud and locating ~350k fine velocity DOFs, plus the Galerkin RAP chain), not cycle application. The 4 V-cycles themselves are cheap; the iteration win (4 vs 46) is real and currently thrown away by rebuilding transfers that depend only on the meshes.

Fix shape, in order of payoff:
1. Cache the built transfers on the mesh (keyed by the level chain and field layout) so repeated solves — a time loop above all — pay the build once. This alone should flip the comparison for any run with more than a handful of solves.
[...]

PR #666, opened 2026-09-02
Document why the multigrid coarse level must be small (#644)

lmoresi:
Closes the documentation gap agreed on #644 (and its superseded twin #643).

## The gap

bknight1 hit an OOM at Pawsey: FMG's coarse solve is `redundant`+LU, which replicates the coarse factorization on **every rank**, and his coarsest level still carried 14k velocity DOFs — ~300 MB/rank replicated, killed at the 900 MB/task cap. The conclusion on the issue was that this is not a bug, it needs documenting.

The premise the page never stated is the one that matters: multigrid is cheap because the bottom of the hierarchy is solved *exactly* at negligible cost, and that is only true if the bottom is genuinely tiny. Nothing in the method makes it so, and nothing warns you when it is not.

So a 14,000-DOF coarsest level is not a coarse-solver problem — it means the hierarchy was built from too fine a base with too few levels, and the "solve the coarse problem for free" premise never held. The fix is another level underneath, not a different bottom solve. Doubling the base cell size and adding one refinement level lands on the **same finest grid** with every coarser level ~4x smaller in 2-D (8x in 3-D). bknight1 measured exactly that: 683 → 416 MB/rank for +11% linear-solve time, finest grid and SNES iteration sequence unchanged.

## This is PETSc's choice, not ours

Worth stating plainly, because the issue reads as though Underworld picked `redundant` and thereby caused the OOM. `PCSetUp_MG` selects `PCREDUNDANT` when the communicator has more than one rank and `PCLU` when it does not — its own source comment reads *"coarse solve is (redundant) LU by default"* (`src/ksp/pc/impls/mg/mg.c`). Confirmed by inspection at `np` = 1, 2 and 4 on PETSc 3.25.

We set it explicitly only so a bundle applied after another cannot inherit a sibling's leftover options (#468). Deleting the keys and letting PETSc decide would change nothing — which is the useful part, because it puts attention back on the hierarchy.

## Two corrections to what was there

- The page claimed the default coarse solve was a bare `lu`, in **both** the bundle listing and the "Parallel coarse solve" section, and told users to set `redundant` themselves for large partitions. It has been the default all along, and a bare `lu` cannot factor a distributed matrix — it fails at `np > 1`.
- The SVD path is now documented as what it actually is. It keys on **verified rotation null modes**, not on a problem being "rotated": a rotated problem with Dirichlet walls or a split-fault contact box has none, and a blanket dense SVD coarse solve accounted for most of ~0.8 s per V-cycle application in #622. Advising `svd` for anything rotated would have been an expensive mistake.

## Also

A snippet to check the hierarchy you actually built (`mesh.dm_hierarchy[0]` is the coarsest), verified to run, and a rule of thumb — low thousands of DOFs at the bottom; if it reaches five figures, add a level.

`docs-build` succeeds with no warnings on this page.

## Not done here

The size-aware default from the issue's option 1 (warn or switch when the coarsest level is large). That is a code change, and the call on #644 was documentation. It would be a reasonable follow-up: the check is cheap, and the failure it prevents currently presents as an OOM kill with no hint of the cause.

### DIFF
--- a/docs/advanced/multigrid-preconditioning.md
+++ b/docs/advanced/multigrid-preconditioning.md
@@ -169,10 +169,19 @@ mesh = uw.meshing.UnstructuredSimplexBox(..., cellSize=0.10, refinement=2)
 mesh = uw.meshing.UnstructuredSimplexBox(..., cellSize=0.20, refinement=3)
 ```
 
-Each refinement level divides cell size by two, so in 2-D the level above has
-about 4x the cells (8x in 3-D). Doubling the base cell size and adding one level
-therefore lands on the same finest grid while making every coarser level — the
-bottom one included — about 4x (or 8x) smaller.
+Each refinement level divides cell size by two, so the level above has 4x the
+cells in 2-D and 8x in 3-D — measured exactly, 66 / 264 / 1056 and
+184 / 1472 / 11776 on a `refinement=2` box. Doubling the base cell size and
+adding one level therefore lands on the same finest grid while making every
+coarser level — the bottom one included — about 4x (or 8x) smaller.
+
+"The same finest grid" is approximate, because the base mesh is unstructured and
+gmsh does not produce exactly 4x the cells for half the target size. Measured on
+a unit box, `cellSize=0.10, refinement=2` gives 3,872 finest cells against
+`cellSize=0.20, refinement=3` at 4,224 — 9% more, while the coarsest level drops
+to 27% of its former size. The production model in #644 landed the other way:
+227k against 233k. Expect a few percent either side, and check rather than
+assume.
 
 Check what you actually built rather than assuming; `dm_hierarchy[0]` is the
 coarsest level and `[-1]` is the mesh you solve on:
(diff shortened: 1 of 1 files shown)


lmoresi:
## Adversarial review

Copilot approved this, but only on the grounds that it is documentation-only. Documentation-only is not the same as harmless: a doc that tells you to run something that cannot work costs more than silence, because you spend the afternoon believing the tool is broken. We reviewed it by **running** its claims rather than re-reading them. Five findings, two of them wrong advice. All fixed in `ffd8920f`.

### 1. The telescope recommendation cannot work at all — and we had promoted it

The page offered `mg_coarse_pc_type = "telescope"` for large partitions, and our rewrite gave it its own subsection, making it more prominent than before.

`PCTELESCOPE` has no DMPlex support, and every Underworld mesh is a DMPlex. Running the snippet as written, at np=2:

```
[0]PETSC ERROR: No support for this operation for this object type
[0]PETSC ERROR: Support for DMPLEX is currently not available
```

PETSc error 56, solve aborted. This was pre-existing text we inherited and endorsed. It is now a warning that the option is unavailable — which actually **strengthens** the page's thesis, since it means there is no bottom-solve escape hatch from an over-large coarse level. The hierarchy is the only lever.

### 2. The SVD snippet was inert

We showed:

```python
geometric_mg_bundle(coarse="svd")
```

That constructs an `MGSettings` namedtuple and discards it. It applies nothing. It read as a user action and did nothing at all. Removed — the solver already selects SVD from the verified rotation-mode count, which is the accurate story.

### 3. The "override it yourself" promise is false on exactly the path that needs it

The page promised you can keep `preconditioner = "auto"` and override the coarse solver through `petsc_options`. We verified that on the **standard** path: `fieldsplit_velocity_mg_coarse_pc_type = svd` is still set after the solve. Good.

It is **false on the rotated path** — the one path where an SVD coarse solve is the whole point. That route writes its bundle under the velocity sub-PC's own prefix, applies it, then deletes the keys again; anything set through `solver.petsc_options[...]` lives under the SNES prefix and never reaches it. Now scoped to the standard path, with the rotated exception called out.

### 4. "Lands on the same finest grid" overclaimed

Measured on a unit box: `cellSize=0.10, refinement=2` gives 3,872 finest cells; `cellSize=0.20, refinement=3` gives 4,224 — **9% more**, not equal, while the coarsest level drops to 27%. #644's production numbers go the other way (227k vs 233k). The base mesh is unstructured and gmsh does not halve cleanly. Now stated as a few percent either side, with both measurements, and a note to check rather than assume.

### 5. The rule of thumb omitted the two variables that set it

"Low thousands of DOFs" has no rank count and no memory budget in it, yet those are precisely what decide the threshold — the replicated factor is paid *per rank* against a per-rank cap. Rewritten around that arithmetic and calibrated on #644's own numbers (~14k DOFs → ~300 MB/rank at 5.16x fill).

### What survived

Stated here so the review is not only negative — each of these was run, not assumed:

- The `mesh.dm_hierarchy` diagnostic works on a **bare mesh**, with no solver or MeshVariable constructed first. The `.pyx` warns the hierarchy "may not be populated this early", so this was worth checking.
- Per-level growth is exactly 4x in 2-D and 8.00x in 3-D — 66 / 264 / 1056 and 184 / 1472 / 11776 on a `refinement=2` box.
- PETSc's own source selects `PCREDUNDANT` above one rank and `PCLU` at one, confirmed at np = 1, 2, 4.

Issue #671, opened 2026-09-02
The co-located multigrid tail fails at np=3 on the 3-D width network: a Galerkin coarse operator with a missing diagonal, and a 10% slip error


lmoresi:
Found while measuring the placement gather (#670) on the crossing-patches fixture (`tests/parallel/ptest_0863`, split realisation, linear viscous + contact). np=2 and np=4 reproduce the serial slips to five digits; np=3 does not, in two different ways depending on the partition.

**With the placement gather as on development** (the wide mark of #670), np=3 completes the solve but the smallest piece is 10% low:

```
AssertionError: Cross_2: parallel peak 0.0027 vs serial 0.00295
```

Main and Cross_1 pass. The velocity block is custom-FMG, as asserted.

**With the tightened gather mark** (branch feature/placement-local-gather, which moves about half as many cells and so changes the partition of the fine level), np=3 fails in the multigrid setup on rank 2, reproducibly:

```
[2]PETSC ERROR: Argument out of range
[2]PETSC ERROR: New nonzero at (1497,1497) caused a malloc.
#3 MatDiagonalSet_Default ... #8 MatGalerkin ... #9 PCSetUp_MG
```

A Galerkin coarse operator with no diagonal entry means a coarse degree of freedom that no fine row maps to, i.e. a zero column in the transfer that the parallel repair (`_repair_zero_columns_parallel`) did not catch, or a coarse row that is owned on a rank with no fine image on that rank. The serial path repairs such columns; the parallel guard is the piece the design note (`fault-patch-multigrid-2026-08.md`, parallel rule 5) lists as needing a parallel form.

Both symptoms are in the co-located tail (#662) at three ranks, not in the placement. np=3 was the rank count the design rule said to test, and neither `ptest_0863` nor the other network ptests run at np=3 in CI (they are not in `scripts/test.sh`'s globs at all).

Underworld development team with AI support from Claude Code

lmoresi:
Second occurrence, np=4, on a different layout: two faults in one network (split realisation) on a [-0.5, 1.5]^3 base, band gathered onto rank 2 with 12,803 of 28,656 cells. Same signature, on rank 0 this time:

```
[0]PETSC ERROR: New nonzero at (1599,1599) caused a malloc.
#3 MatDiagonalSet_Default ... #6 MatDiagonalSet_MPIAIJ
```

[...]

Issue #670, opened 2026-09-02
Placement gathers far more than the surgery needs, onto one rank for the whole network, and never sends it back


lmoresi:
Measured on the crossing-patches fixture of test_0863 (unit box, h_far = 0.24 refined once so base edges are about 0.126, h = 0.08, w = 0.04, two patches spanning 0.3 to 0.7).

**The fill is small; the gather is not.** The base has 5592 cells and the placed mesh 8405, of which 1663 are band. So the placement adds the band plus about 1150 fill cells, all within a cell of the patches. But at np=2 the gather moved 5087 of the 5592 base cells onto one rank before any surgery happened, and at np=4 the surgery rank ends with 8012 of 8405 cells. The imbalance is the gather, not the fill.

**Cause, in `place_thin_volume`.** The gather mark claims every vertex within `reach + 2 h_vertex` of the band's skin, and `_gather_region` then takes the star of the marked vertices and one further layer. Counted on the base:

| region | cells | share |
|---|---|---|
| carve reach (0.038) | 334 | 6% |
| reach + 1 h | 1313 | 23% |
| reach + 2 h (the mark) | 3486 | 62% |
| mark + star + layer | 5592 | 100% |

The seam rule the gather exists for (nothing the surgery deletes or creates may be shared) needs the carve victims, their vertex star, and one layer. The distance blanket of two cell widths looks like an earlier stand-in for the two topological layers, and both survive, so the margin is paid twice. On a large domain it is a shell two cells thicker than needed; on a small one it is the whole mesh.

**One target for the whole network.** The placer takes the network as one assembly and gathers the union of every patch's region to the single rank holding the most marked cells. Two faults each interior to a different rank are both moved to one of them. The split's `_redistribute_fault_interior` has the same one-target rule for the union of labels (its mark is topological, so it does not over-reach).

**No return.** The moved cells stay on the surgery rank. The moved region is exactly the neighbourhood the base's partition had balanced, so the extra load lands on the rank that was already heaviest there: anti-balanced, and paid for with a redistribution.

**Direction (discussion with Louis, 2026-09-02):** keep gather-first as the mechanism and make it hit the design: (1) a region whose star is already unshared on one rank is placed there with no move; (2) mark the victims only and let the star-plus-layer growth supply the seam freedom; (3) one region per connected cluster of patches, each to its own rank, surgeries concurrent; a move only for a region that actually straddles a seam. The accepted trade is extra load on the owning rank in exchange for no communication.

[...]

PR #672, opened 2026-09-02
Placement and split gather per region; the geometric tail solves on a distributed band (#670, #671)

lmoresi:
Stacked on #669 (the base branch here); three commits.

The gather-first surgery moves base cells onto one rank so that nothing it deletes or creates is shared. Three things were wrong with how much and where (#670): the mark reached two cell widths beyond the carve before the star and layer were grown from it; the whole network was one region with one target rank; and a region already interior to a rank was moved anyway.

**1. The mark covers only what the carve drops** (victims plus the crossed cells' vertices, one cell diameter out), the same rule at every placement path. `_gather_region` reports the moved count, and `ptest_0855` bounds it by the cells within three median diameters of the zone.

**2. The gather takes regions.** `_gather_regions` marks a chart of region ids, claims each region's star and layer, merges regions whose shells touch (collective union-find), and sends each to the rank already holding most of it — or leaves it where it is when it is interior to one rank. One shell partition moves them all. `ptest_0857` covers the three behaviours at np=2, 3, 4.

**3. The thin volume and the split use the regions.** One region per connected component of the assembly; the owning ranks carve and fill their own components concurrently on their compacted share of the assembly, and the collective rebuild sews them at once. The split's redistribution takes groups of faults and delegates to the same gather; the network passes it the regions the placement reported, so what the placement kept apart is not gathered together afterwards. Outcrop and ladder paths keep one region.

**Measured** (base 27,544 cells, two crossing patches; or two zones a domain apart):

| case | before | after |
|---|---|---|
| crossing patches, cells gathered | 9831 | 5046 |
| two zones apart, np=2, cells moved | 8451, one rank | 304, two owners |
| two faults in one network, split, np=2 | one rank | nothing moved: 14,339 / 14,332 cells |
| same, np=4 | one rank | each region to its own rank, 2507 moved |

**Tests**: `ptest_0855` (+2 cases), `ptest_0857` (new), `ptest_0863` (+1 case) at np=2 and 4 (0855 and 0857 also at 3); lines, sheet, remove-embedded, fault-split, 3-D split, network and place-split-contact parallel suites at np=2; the serial thin-volume and network suites.

**Not fixed here.** The contact solve on the balanced two-fault layout fails in the co-located multigrid tail's setup (#671) even at np=2, where nothing moved: the band is then genuinely distributed, and the tail was built for a band on one rank. The placement and the split complete. Also pre-existing: the gap fill's placed-vertex count varies with np by one or two (611 vs 612 on the same input under the previous build), a gmsh ordering effect.


lmoresi:
Adversarial review (we read the diff against what it does not measure).

1. **The bound is calibrated, not derived.** Three median diameters was chosen because the new mask sits at 5046 against 7269 and the old at 9831 on this box. A mesh with a wider spread of cell sizes than a refined gmsh box could move the shell past the median-based bound without any regression in the rule. The rule itself, victims plus star plus layer, is not what the test checks.
2. **The mark is still a distance, and the rule is topological.** "Crossed cells' vertices within one cell diameter" is an estimate of which cells the carve will drop; the carve decides by geometry afterwards. The surgery raises if the gathered region turns out to touch a shared point, so an under-reach is loud, and no path hit it at np=2, 3 or 4. A mark taken from the carve's own victim set would remove the estimate.
3. **The interior short-cut was not exercised by any test.** It fires when the star counts are non-zero on exactly one rank; on every fixture here the region straddles a seam at every rank count, so the branch has been measured only in that it did not fire. A test needs a partition it controls.
4. **`n_gathered` is reported by two of the six paths.** Lines, the 2-D thin volume and the ribbon carry the same mask change without the count.
5. **The one-target rule stands.** Two surfaces interior to different ranks are still gathered to one of them; the mask change halves what moves, it does not change where.

None of these change the measured numbers. Item 2 is the one that would make the rule exact rather than estimated.

lmoresi:
Review addendum for the per-region commits (we read the diff against what it does not cover).

1. **A merged region's owner is the majority rank of the union, chosen after the merge.** Two zones whose shells touch become one region and move together; that is correct for the seam rule, and it means "close" is measured in shells (three base cells each), not in zone widths. Two faults 0.8 apart on 0.126 cells merged in our measurement; 1.4 apart did not. The docs say so; a user cannot see the merge except through `n_regions`.
2. **The per-rank surgery compacts its share of the assembly, so the assembly node numbering the surgery sees is per rank.** Everything downstream that keyed on global assembly indices was checked (made-cell codes, placed rows, skin rows, labels); `embedded_nodes` are coordinates and unaffected. The outcrop path was deliberately left on one region because its cap and bowl machinery broadcasts from a single root.
3. **The split's groups come from the placement's report, not from the mesh.** `info["embedded_regions"]` identifies each mid-surface's region by matching its first node's coordinates against the assembly at 1e-12; a mid-surface that failed to match falls into region 1. A split called without the network (no groups) is one region, as before.
4. **The interior short-cut is now exercised** (ptest_0857 and the two-fault network at np=2), which the first review asked for.
5. **The solve is the open end.** Every measurement here stops at the sewn, split mesh; on the balanced layout the multigrid tail refuses (#671). The claim of this PR is placement and split parallelism, not the solve.

lmoresi:
Fourth commit: the tail on a distributed band (#671). Per-region placement produced the first layouts where the band spans ranks, and the co-located multigrid tail failed on them. Two defects, both measured:

1. The rotated prolongation zeroed its constrained rows by LOCAL block index; zeroRows takes global ones, so every rank but the first zeroed the wrong rows. That is the "New nonzero caused a malloc" in PCSetUp_MG: PETSc repairs a zero row of the Galerkin coarse operator with an identity diagonal, and the diagonal was never allocated.
2. The "auto" transfer mode replaced a co-partitioned transfer with 3 orphan columns by a cross-partition one with 16,791 (np=2) or 43,047 (np=4) of 96,009; the repair then injected all of them and the velocity KSP failed at its first iteration. The rebuild is now kept only where it does better, with a warning otherwise.

With both, the two-fault network solves on custom-FMG at np=2 and np=4 in one Newton step, slips within 0.5% of serial. The np=3 answer discrepancy on the crossing fixture (10% on the smallest piece) is unchanged and stays open in #671. The parallel custom-MG (1017) and rotated free-slip (1064, 1066, 1068) suites pass at np=2.

Review note on ourselves: the cross-partition builder losing 17 to 45 percent of the coarse columns on a co-partitioned pair is a defect in its own right, not fixed here; we only stop it from being chosen when it is worse. It is measured, warned about, and left for #671.

lmoresi:
Fifth commit: the design note (`docs/developer/design/fault-parallel-placement-2026-09.md`, linked from the developer index's toctree and authority map) and the layout throughput test beside it. The table from the test, np=3 on a 6 x 1 x 1 box with three faults:

| layout | cells moved | cells per rank (max/mean) | warm solve | velocity its | pressure its |
|---|---|---|---|---|---|
| serial | 0 | 20778 | 81.0 s | 3 | 200 |
| local (one fault per slab) | 122 | 1.06 | 45.8 s | 4 | 200 |
| one fault straddling a seam | 834 | 1.22 | 52.9 s | 4 | 200 |
| all gathered to one rank (pre-#672) | 5440 | 1.91 | 78.3 s | 4 | 200 |
| local, GAMG | 122 | 1.06 | 17.6 s | 26 | 200 |

The answer is layout-independent to four digits. Per-region placement turns np=3 from no gain (78 s against 81 s serial) into 1.8x; a straddling fault costs 15%. And on this fixture the pressure Schur solve is at its 200-iteration cap in every configuration, so wall time is the pressure block's, and GAMG beats the tail in wall time despite six times the velocity iterations. That is the #625 cap seen on a contact fixture, not a placement matter, and the note says so.

Also in this commit: the gather's growth beyond the star is a parameter, and the placement was measured at zero extra layers — its own gate refuses, so the third layer stays.

lmoresi:
Sixth commit: the layout table re-measured with the pressure solve converging. The first pass had every configuration's pressure Schur solve at its 200-iteration cap; the monitor showed a floor set by the inexact velocity solve inside the Schur application (mechanism on #625). With the pressure tolerance at 1e-4 (knob `solver._rotated_pres_rtol`, default unchanged):

| layout | cells per rank (max/mean) | warm solve | velocity its | pressure its |
|---|---|---|---|---|
| serial | 1.00 | 14.0 s | 3 | 26 |
| np=3 local | 1.06 | 6.8 s | 4 | 26 |
| np=3 one fault straddling a seam | 1.22 | 8.7 s | 4 | 26 |
| np=3 local, GAMG | 1.06 | 3.3 s | 26 | 26 |

The answer is unchanged to four digits. np=3 per region is 2.1x serial; one non-local fault costs 28%. GAMG is still twice as fast as the tail in wall time on this linear, uniform-viscosity fixture, which the note records as a red flag for a contrast-fixture measurement rather than a placement result.

lmoresi:
Seventh commit: the table at tolerance 1e-3, where the default margins converge the pressure solve in 26 iterations without any knob — the 1e-5 of the first pass was stricter than the resolution deserves. Warm: serial 10.1 s, np=3 local 5.4 s, one fault straddling a seam 5.3 s, GAMG 2.2 s. The non-local fault's cost is now stated as a bound (none to 28% across the three passes, single runs) rather than a number. The penalty is recorded as a traction-recovery matter per Louis's note on non-planar boundaries: the multiplier plus the augmentation is the traction, and the Coulomb law's reaction-fed normal stress needs that correction before a penalty is used with friction.

lmoresi:
29262483 adds the split through the partition seam in 2-D, the cut the conform placement was missing. The chain is assembled globally from the star-forest identities of its vertices, a vertex on the seam is duplicated on every rank holding it (replica owned with the original, so a coincident pair never straddles ranks), re-homed seam edges keep their star-forest entries, and the pair's normal at a seam vertex is recorded by the split so the contact solve needs no exchange. Nothing is redistributed.

Measured against the gathered answer: the long vertical fault gives 0.5094 with 69 owned pairs at np=1, 2 and 3 (cells 1212/1042 and 785/890/571); the three-piece network gives its serial peaks at np=2 and 3. `ptest_0865` pins both.

Two defects the measurement exposed are fixed in the same commit: the conform placement could put the two cells of a spine facet on different ranks (np=3), and the blind rule dropped an unclipped one-edge piece of a spine between two seam vertices. A fill ring that crosses itself is now refused before gmsh rather than spinning in it; the two conform fill configurations that produce one (a tip on the seam beside a junction; a junction within a band width or two of the seam) are recorded in the design note as open.

Issue #657, opened 2026-08-27
SUPG and Levelset


NengLu:
I've created a branch: https://github.com/underworldcode/underworld3/tree/levelset

which includes several tests in the test_tem directory. The SUPG code is saved in the solver_supg.py file.
And the levelset implementations are saved as level_set_SLCN and level_set_SUPG.

Please take a look and feel free to commit any changes. @lmoresi @gthyagi

lmoresi:
Thanks for this, NengLu. We measured the `levelset` branch against SLCN and then built the production version of the solver on top of your weak-form idea. PR #673 has the result; the numbers and the code findings are here so the discussion stays with the issue.

**What the SUPG prototype gets right.** The `F1 += tau u R` form works exactly as advertised on PetscDS with P2 elements, and the class runs unchanged on current `development`. On a rotating Gaussian at Courant 0.5 it is far more accurate than SLCN on the same mesh (0.6% against 21% round-trip error after one revolution) and 4 to 6 times cheaper per step. On your LeVeque level-set comparison (T = 2, 64 by 64) it gives a shape error of 0.028 against 0.051 for SLCN at half the wall time. The implicit scheme is stable at any cell Courant number: a band refined to h/9 across the orbit, at local Courant 13, changes the error in the third digit only. That is the property the transport design note asks for.

**What limits it.** Accuracy is set by u dt against the feature width, as dt^2 for Crank-Nicolson: 9.8% at Courant 2, 66% with undershoots to -0.35 at Courant 8, while SLCN stays at 8% until its trace-back arc exceeds about 10 degrees per step. So the two schemes cross over near Courant 2 on the feature's own scale.

**Findings in the branch as pushed.**
1. `level_set_SUPG.LevelSetSolver.solve` calls `self._adv_solver.solve(dt)` positionally after the solver's `solve` became keyword-only, so the LeVeque script cannot run without a one-line fix.
2. The discontinuity-capturing term's "residual" is `u . grad(phi_old)` alone, without the time-derivative part, so it is O(1) on any cleanly moving feature and adds first-order diffusion of size h|u| everywhere. With the default coefficient it erodes the Gaussian peak to 0.34 and loses 28% of the mass at Courant 2. Reinitialisation hides this on a tanh level set; on temperature it does not.
[...]

PR #673, opened 2026-09-03
Eulerian SUPG advection-diffusion with symbolic BDF and Adams-Moulton orders (#657 follow-on)

lmoresi:
## Credit

The SUPG weak form used here, writing the Petrov-Galerkin test-function perturbation as a flux so PETSc needs no modified test space, and its first working implementation on PetscDS with P2 elements are @NengLu's, from the `levelset` branch of issue #657, together with the LeVeque swirling-flow comparison against SLCN and the conservative level-set pipeline (reinitialisation and mass correction) that motivated it. This PR keeps that idea and its tau parameter, builds the time integration on the symbolic history machinery, and adds the measurements; it would not exist without that prototype. A follow-on PR brings the level-set solver from the same branch onto this class.

Follows issue #657 (@NengLu's SUPG and level-set branch). This PR builds the Eulerian transport solver on the symbolic time-history machinery so that the scheme's order is a construction argument, adds the analytic oracle and the tests that measure it, and fixes two defects found on the way, one of them general.

## What is here

**`uw.systems.AdvDiffusionSUPG(mesh, T, V_fn, order=2, integrator="bdf")`** (`systems/advection_diffusion_eulerian.py`). Fully implicit Eulerian advection-diffusion with SUPG stabilisation. Every past time level is a mesh variable held by the `Eulerian` history manager, so BDF 1-3 and Adams-Moulton 1-3 (theta at order 1, so `am, 1, 0.5` is Crank-Nicolson) are assembled from the same stored history, including the diffusive flux of past states. The SUPG term is the test-function perturbation written as a flux, as in the prototype. The timestep, multistep coefficients and the tau weights are runtime constants of the compiled kernels: a change of timestep costs nothing (the prototype recompiled on every change, 1.2 s against 0.03 s for a step). Diffusivity is set on the constitutive model like every scalar solver. The class reproduces @NengLu's Crank-Nicolson prototype to four digits.

**`uw.analytic.RotatingGaussian`**: a Gaussian carried round the origin by rigid rotation while diffusing, exact at every time, so a transport scheme's error is measured rather than eyeballed. Also fixes `AnalyticSolution.error(norm="integral")` for scalar variables, which had never been exercised.

**Tests** (named to match the CI batch globs): `test_1055` (API, no-recompile contract, adapt-child solve), `test_1100_advdiff_supg_rotating_gaussian` (temporal order 1 and 2, band invariance, round trip), `tests/parallel/test_1077` (np=2 equals serial to 1e-12), `test_1058` (the aux-layout fix below).

**Design note** `docs/developer/design/eulerian-supg-transport.md` with the measurements, and an example `docs/examples/convection/advanced/Ex_AdvectionDiffusionSUPG_RotationTest.py`.

## What was measured

Rotating Gaussian, P2, unstructured simplex box, one revolution, relative L2 error. The implicit scheme is stable at any cell Courant number, and cells the scalar does not need are free: a band refined to h/9 at local Courant 13 changes the error in the third digit only. Its accuracy is set by u dt against the feature width (dt^2 for the second-order schemes). SLCN's error is flat in dt but accumulates at small Courant (21% against 0.6% at Courant 0.5 on the same mesh) and costs 4 to 6 times more per step in serial; its limit is the trace-back arc, about 10 degrees per step.

The integrator study (res 32, Courant 0.25 to 8, pure advection and kappa 1e-3) decides the default: Adams-Moulton above order 1 blows up from Courant 1 (bounded stability region), BDF3 fails from Courant 4, Crank-Nicolson is three to four times more accurate than BDF2 at the same timestep but rings once the feature is under-resolved in time, and backward Euler carries 20 to 40% error at any practical timestep. Cost per step is identical across schemes. BDF2 is the robust default; the note says when to pick the others.

## Defects fixed

1. **Auxiliary fields packed and indexed by position** (`Mesh.update_lvec`, `_jitextension`). `mesh.vars` holds variables weakly; a variable released by `uw.reset_default_model()` (the test suite does it between tests) or by the statistics helpers' deliberate `del mesh.vars[...]` leaves its PETSc field in the DM. Every later variable was then packed into, and read from, the wrong slots: measured as a P0 cell-size field landing in a P2 slot as garbage, NaN residuals in one run and a subtly wrong answer in the next depending on when the collector ran. Now packed by field name (orphans zeroed) and offsets read from the DM's own field list. `test_1058` fails 2 of 3 checks without the fix. This predates the PR and affects any solver with auxiliary fields; it could be cherry-picked on its own.
2. **Mesh-owned multigrid pickup on a solver that owns its preconditioner** (`custom_mg.build_transfers`): installing the adapt child's PCMG hierarchy on an additive-Schwarz PC segfaulted inside PETSc. `_pc_option_prefix is None` now counts as the explicit choice it is.

## Not in this PR (recorded in the design note)

ALE mesh velocity (phase 1 keeps the default REMAP transfer with the material velocity, which is already correct to interpolation accuracy), a streamline element length from a mesh-owned metric tensor, a corrected discontinuity-capturing term, and the trajectory-based `estimate_dt` for SLCN (separate small PR). The level-set wrappers and scripts stay on the `levelset` branch for their author to rebase onto this solver.

### DIFF
--- a/docs/advanced/eulerian-advection-diffusion.md
+++ b/docs/advanced/eulerian-advection-diffusion.md
@@ -19,9 +19,19 @@ adv.constitutive_model.Parameters.diffusivity = 1.0e-3
 adv.add_dirichlet_bc(1.0, "Bottom")
 adv.add_dirichlet_bc(0.0, "Top")
 
-dt = 0.5 * adv.estimate_dt()
+dt = adv.estimate_dt()          # accuracy-based: 2% of the field's range per step
 adv.solve(timestep=dt)
 ```
+
+The one deliberate difference is the timestep estimate. The semi-Lagrangian
+`estimate_dt` reports the cell-crossing time, which for this solver is neither
+a stability limit nor an accuracy one. The Eulerian solver's `estimate_dt`
+instead returns the step at which the field changes by a given fraction of its
+range (0.02 by default), from the advective rate before the first solve and
+from the rate the last step actually produced after it. It does not depend on
+the mesh, so cells refined for the Stokes problem do not shrink it. A script
+that sizes its step in Courant numbers can still ask for
+`estimate_dt(basis="resolution")`.
 
 ## What carries over
 
@@ -32,7 +42,7 @@ adv.solve(timestep=dt)
 | `order=2, theta=1.0` | same | SL-BDF2 becomes BDF2 |
 | `order=2, theta=0.5` | refused | refused for the same reason: a BDF stencil does not pair with a centred flux |
 | `f`, `V_fn`, `constitutive_model`, `delta_t` | same | |
-| `estimate_dt(direction_aware, percentile)` | same | the cell-crossing time, a resolution guide for both |
+| `estimate_dt()` | accuracy-based by default | the field may change by `fraction` (0.02) of its range per step; `basis="resolution"` returns the cell-crossing time SLCN reports |
 | `solve(zero_init_guess, timestep, ...)` | same | |
 | `DuDt.set_initial_history(values, dt)` | same | plant an exact history to start at full order |
 | `restore_points_func`, `monotone_mode`, `old_frame_traceback`, `DFDt` | ignored, with a warning | they configure the trace-back |
--- a/docs/advanced/eulerian-advection-diffusion.md
+++ b/docs/advanced/eulerian-advection-diffusion.md
@@ -102,9 +102,14 @@ of the compiled kernels; nothing is recompiled.
 - The stabilisation parameter uses the local cell size (`mesh.cell_size()`) and
   three weights that are runtime constants (`solver.tau_weights`);
   `solver.supg_weight = 0` gives the plain Galerkin scheme for comparison.
-- The linear system is nonsymmetric, so the solver defaults to GMRES with an
-  additive-Schwarz ILU preconditioner instead of algebraic multigrid. Every
-  option can be overridden through `solver.petsc_options`.
+- The linear system is nonsymmetric, so the solver uses GMRES with an
+  additive-Schwarz ILU preconditioner, with the Krylov tolerance matched to the
+  SNES tolerance so that a step is one Newton iteration. Measured, this is the
+  cheaper solve at every Courant number up to eight ranks and its iteration
+  count does not grow with the rank count. `solver.preconditioner = "fmg"`
+  switches to geometric multigrid over the mesh's refinement hierarchy
+  (`refinement >= 1`) for very large rank counts. Every option can be
+  overridden through `solver.petsc_options`.
 
 ## Further reading
 
--- a/docs/advanced/eulerian-advection-diffusion.md
+++ b/docs/advanced/eulerian-advection-diffusion.md
@@ -37,9 +37,7 @@ adv.solve(timestep=dt)
 | `DuDt.set_initial_history(values, dt)` | same | plant an exact history to start at full order |
 | `restore_points_func`, `monotone_mode`, `old_frame_traceback`, `DFDt` | ignored, with a warning | they configure the trace-back |
 
-`order=3` (BDF3) is available; see below for when it is safe. `integrator="am"`
-above order 1 reaches the higher Adams-Moulton rules, which are for
-diffusion-dominated problems only.
+`order=3` (BDF3) is available; see below for when it is safe.
 
 ## When to use which
 
@@ -80,7 +78,7 @@ tables are in the design note.
 | BDF2 (`order=2`) | damped and stable at every Courant number; the choice for sharp or under-resolved fields |
 | BDF3 (`order=3`) | the most accurate scheme below Courant 1 when diffusion is present; on pure advection it grows slowly at any Courant number, so use it only with diffusion |
 | backward Euler (`order=1, theta=1.0`) | 20 to 40% error at any practical timestep; not for transport |
-| Adams-Moulton 2, 3 (`integrator="am"`) | third and fourth order below Courant 1; blow up on advection from about Courant 1 |
+| Adams-Moulton 2, 3 (not offered) | third and fourth order below Courant 1 but blow up on advection from about Courant 1, which is why there is no knob for them |
 
 All schemes cost the same per step: the history terms are extra kernel inputs,
 not extra solves. Changing the timestep between steps changes a runtime constant
--- a/docs/advanced/eulerian-advection-diffusion.md
+++ b/docs/advanced/eulerian-advection-diffusion.md
@@ -1,0 +1,105 @@
+# Eulerian advection-diffusion (SUPG): a drop-in for SLCN
+
+`uw.systems.AdvDiffusionSUPG` solves the same scalar transport equation as the
+semi-Lagrangian solver `uw.systems.AdvDiffusionSLCN`,
+
+$$
+\frac{\partial \phi}{\partial t} + \mathbf{u}\cdot\nabla\phi
+    - \nabla\cdot(\kappa\nabla\phi) = f ,
+$$
+
+but assembles every term on the mesh, implicit in time, with streamline-upwind
+(SUPG) stabilisation. There is no trace-back and no departure point. The two
+classes share their interface, so switching is one line:
+
+```python
+adv = uw.systems.AdvDiffusionSUPG(mesh, T, v.sym, order=1)   # was AdvDiffusionSLCN
+adv.constitutive_model = uw.constitutive_models.DiffusionModel
+adv.constitutive_model.Parameters.diffusivity = 1.0e-3
+adv.add_dirichlet_bc(1.0, "Bottom")
+adv.add_dirichlet_bc(0.0, "Top")
+
+dt = 0.5 * adv.estimate_dt()
+adv.solve(timestep=dt)
+```
+
+## What carries over
+
+| SLCN | SUPG | note |
+|---|---|---|
+| `order=1, theta=0.5` | same | Crank-Nicolson, the default for both |
+| `order=1, theta=1.0` | same | backward Euler |
+| `order=2, theta=1.0` | same | SL-BDF2 becomes BDF2 |
+| `order=2, theta=0.5` | refused | refused for the same reason: a BDF stencil does not pair with a centred flux |
+| `f`, `V_fn`, `constitutive_model`, `delta_t` | same | |
+| `estimate_dt(direction_aware, percentile)` | same | the cell-crossing time, a resolution guide for both |
+| `solve(zero_init_guess, timestep, ...)` | same | |
+| `DuDt.set_initial_history(values, dt)` | same | plant an exact history to start at full order |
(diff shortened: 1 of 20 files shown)


lmoresi:
Adversarial review of this PR, by the same session that wrote it. Findings first, evidence with each.

**1. The strong residual has no diffusion term, so SUPG is inconsistent for P2 fields with kappa > 0.** PETSc kernels see first derivatives only; the class documents this and the study ran at cell Peclet ~40 without visible harm, but nobody has measured the P2 error with kappa large enough for the missing term to matter. A lagged projected Laplacian is the known repair. Not fixed here.

**2. tau uses `mesh.cell_size()`, which is volume^(1/d), not an edge length, and is partition-dependent by ~1e-3 (documented TODO in `_assemble_cell_size`).** The weights (2, 2, 4) were kept from the prototype and shown to reproduce its numbers; they were not calibrated against the streamline length. The np=2 test passes to 1e-12 on a uniform mesh, which does not exercise the seam dependence.

**3. BDF3 silently falls back to variable-step BDF2 when consecutive timesteps differ by more than 5%** (`ddt._bdf_coefficients`). A user who varies dt gets second order without being told. Documented in the docstring, not warned at runtime.

**4. Adams-Moulton orders 2 and 3 are accepted by the constructor and blow up from Courant 1 on advection.** We kept them with a docstring warning rather than refusing them. A runtime check is impossible without knowing the flow; a warning at construction would be cheap. Judgement call, flagged.

**5. Orphaned DM fields are zeroed at packing but never reclaimed.** The fix makes the layout correct; the DM still grows by one field per dropped variable for the life of the mesh. A script that creates variables in a loop and drops them will see its aux vector and its kernels grow. That is the pre-existing memory behaviour, now merely correct.

**6. The default `order=2` changes nothing for existing users (new class), but `theta` is silently ignored at AM order 2 and 3.** We raise if theta != 1 there, so it is loud, not silent. Verified by `test_scheme_arguments_are_validated`.

**7. The rotating-Gaussian convergence slopes for BDF1 are 0.80 and 0.88, not 1.** The test tolerance is 0.35 below the expected order. The pre-asymptotic behaviour is real (backward Euler saturates fast on this problem); a tighter test would need smaller timesteps and a finer mesh than a level_2 test can afford.

**8. The study's res-64 rows were still running at PR time.** The res-32 table is complete and decides the default; res 64 is confirmation and will be appended to the design note when it lands.

What we checked and found sound: the two families agree to every digit where they coincide (BDF1 = backward Euler); Crank-Nicolson reproduces the #657 prototype to four digits at two Courant numbers; a timestep change keeps the JIT cache key and changes the answer by 2e-11 against a 2e-2 negative control; the aux-layout regression test fails 2 of 3 checks with the fix reverted; core batches test_00xx to test_02xx pass (398) with the layout fix in place.

lmoresi:
Res-64 rows of the integrator study are in (design note updated in the last commit). BDF2 and Crank-Nicolson track their res-32 errors at the same u dt. BDF3 is not safe for pure advection at any Courant number: its stability region misses the imaginary axis near the origin, so the finer mesh's low-frequency modes grow slowly (31x the exact field after 590 steps at Courant 1); it behaves with kappa = 1e-3. Review item 8 is closed; the BDF2 default stands.

lmoresi:
The auxiliary-field layout fix (commit e84dea98 and its test) is extracted as #676 so it can land on development ahead of this PR; this branch will rebase onto it once merged.

lmoresi:
Interface change after discussion: AdvDiffusionSUPG now takes the semi-Lagrangian solver's signature and semantics so it is a drop-in replacement (change the class name, nothing else). order=1 with theta=0.5 is Crank-Nicolson and the default, as for SLCN; order=2 takes theta=1 (BDF2) unless 0.5 is asked for explicitly, which is refused for the reason the SLCN docs give; the trace-back-only arguments are accepted and ignored with a warning; delta_t is settable and solve() reuses it. New user page docs/advanced/eulerian-advection-diffusion.md with the swap table and when to use which solver. Review items 4 and 6 stand as written; the default moved from BDF2 to Crank-Nicolson by the drop-in contract, and the design note records why that is also the better default on the measurements.

lmoresi:
Removed the `integrator` argument: `order` and `theta` already reach every scheme that is safe (Crank-Nicolson, backward Euler, BDF2, BDF3 with diffusion), and the only settings the argument added were the Adams-Moulton orders 2 and 3 that the study shows blowing up on advection from Courant 1. The family now follows the order; the higher Adams-Moulton assembly remains reachable only by switching it on the instance, for the record. Review item 4 is closed by removal.

lmoresi:
Timestep: `estimate_dt()` on the Eulerian solver is now accuracy-based. It returns the step at which the field changes by `fraction` (0.02) of its range, from the advective rate before the first solve and from the realised rate of the last step after it, so it does not depend on the mesh (the band test checks that the resolution estimate collapses 3x on the refined child while this one moves under 25%). The cell-crossing time SLCN reports is neither a stability nor an accuracy limit here; it stays available as `basis='resolution'`. At the default fraction the Crank-Nicolson round trip on the rotating Gaussian is under one per cent (test_1100). Credit to @NengLu is now in the PR description, the module docstring and the design note.

lmoresi:
**Linear solver settled (bd0b1a5c): matched tolerances, Schwarz by default, `preconditioner="fmg"` as a real switch.**

We set out to measure geometric multigrid against the solver's additive-Schwarz ILU on the level-set advection step (256² and 512² quad box with a refinement hierarchy, P2, Crank-Nicolson; script and logs in `~/+Simulations/supg_vs_slcn_657/parallel/`). The first result read as a 1.65x win for multigrid (0.94 s against 1.56 s per step in serial, one F-cycle against three GMRES iterations). `-log_view` said otherwise: the Schwarz run was doing two Newton iterations per step on a linear operator, because the Krylov default tolerance (1e-5) does not reach the SNES tolerance (1e-8), and that second Jacobian assembly cost more than every linear solve of the step. The linear solve is under a tenth of a step either way; assembly is the rest.

With the Krylov tolerance at 1e-9 (now the default) both do one Newton step, and the comparison at 256² is:

| Courant | ASM-ILU its (np 1 / 8) | s/step | FMG cycles (np 1 / 8) | s/step |
|---|---|---|---|---|
| 1/2 | 5 / 5 | 0.913 / 0.121 | 1 / 1 | 0.943 / 0.145 |
| 2 | 8.9 / 8.6 | 0.925 / 0.141 | 3.4 / 3.6 | 1.079 / 0.178 |
| 8 | 16.6 / 16.5 | 0.971 / 0.146 | 12.8 / 12.8 | 1.657 / 0.299 |
| 32 | 37 / 37.8 | 1.128 / 0.172 | 23.8 / 24.1 | 2.347 / 0.457 |

Schwarz is cheaper at every Courant number and its count is the same on one and eight ranks. The multigrid's cycle count grows with the Courant number nearly as fast (the Galerkin coarse operators inherit the fine-grid tau; four levels equal three at Courant 1/2, so they are not under-stabilised there), and a cycle costs about three Schwarz iterations. At 512² with four levels the same holds: 3.51 / 0.48 s (Schwarz) against 3.62 / 0.54 s (multigrid). GAMG as a control was slowest (5 iterations, 2.07 s serial). Smoother variants made no difference. So the default is GMRES + ASM-ILU on every mesh, and `solver.preconditioner = "fmg"` hands the block to the managed multigrid route for the rank count where a one-level method has no coarse space, which we have not measured beyond eight ranks.

Two defects found on the way:

- The solver's `solve()` pre-ran the three setup stages, which marks the solver set up, so the base `_build` returned before resolving the preconditioner choice and `preconditioner = "fmg"` was silently inert. It now builds through `_build`. The semi-Lagrangian solvers share the pattern and the defect: #683.
- The parallel timings in the design note were taken with the unmatched tolerance and carry the second Newton step; the note says so now.

Tests: `test_1055` gained the switch test (Schwarz and multigrid agree to the solve tolerance, and switching back rebuilds); the adapt-child test is unchanged in what it asserts. 18 serial and the np=2 parallel test pass.

lmoresi:
**Convection benchmarks, SUPG against SLCN (2026-09-04).** Same Stokes problem, mesh and timestep rule; only the transport scheme differs. Nusselt number as the boundary integral of the total heat transport q = vT − ∇T evaluated at the integration points and projected onto a nodal field, with the same field integrated across the mid-plane; P2 temperature, Courant 1. Study directory with driver, per-step logs, checkpoints every 25 steps and figures: `~/+Simulations/supg_vs_slcn_657/convection_benchmarks/`.

| case | scheme | Vrms | Nu wall (cold / hot) | Nu interior | transport s/step |
|---|---|---|---|---|---|
| Blankenbach 1a, Ra 1e4 (ref 42.865, 4.884) | SUPG | 42.79 | 4.910 / 4.909 | 4.872 | 0.058 |
| | SLCN | 42.82 | 4.893 / 4.887 | 4.875 | 0.640 |
| Blankenbach 1b, Ra 1e5 (ref 193.21, 10.534), from the 1a state | SUPG | 193.05 | 10.784 / 10.783 | 10.511 | 0.061 |
| | SLCN | 193.08 | 10.771 / 10.741 | 10.523 | 0.658 |
| annulus, Ra 1e4, no-slip, eight cells | SUPG | 38.39 | 2.514 / 2.507 | 2.500 | 0.179 |
| | SLCN | 38.28 | 2.473 / 2.672 | 2.483 | 1.553 |

- The answers are equivalent: Vrms within 0.3%, interior Nusselt numbers within 0.7%, temperature fields within 0.002 (box) to 0.007 (annulus) pointwise at steady state. Both reproduce 1a within 0.6% on this mesh and the 1b mid-plane transport within 0.2%, where the wall values run 2 to 3% high because the P2 mesh under-resolves the wall layer at Ra 1e5.
- SUPG's transport solve is 9 to 11 times cheaper per step, and its answer on four ranks is identical to serial; SLCN's moves in the fourth digit with the partition (#682). Doubling the Courant number moves SUPG's Nusselt number by 0.05% and SLCN's by 0.9%.
- SUPG's wall fluxes balance at steady state on both geometries. SLCN's inner-wall flux on the annulus reads 6 to 8% above the flow every interior shell carries: a first-cell artefact at the curved no-slip wall, not lost heat (#685).
- Two things found on the way and fixed here: `theta` had no setter (the shipped examples set it after construction), and the driver needed `solve()` to build through `_build` (already in bd0b1a5c).

Ruling from the maintainer: the schemes are equivalent on these benchmarks; SUPG is the better choice for convection (cheaper, closer to conservative), SLCN for large Courant numbers; SUPG is scalar-only where SLCN carries vectors and tensors through the same trace-back, which is now on the design note's deferred list.

lmoresi:
Superseded by #688, which contains every commit of this branch plus the DDt transport plugin, the Navier-Stokes solver and the final naming (`uw.systems.AdvDiffusion` is the composing solver; the `AdvDiffusionSUPG` name introduced here no longer exists). The review findings recorded here were addressed on that branch.

PR #676, opened 2026-09-03
Pack and index auxiliary fields by DM field, not by position in mesh.vars

lmoresi:
Extracted from #673 so it can land ahead of the solver work. Nothing here depends on that PR. The defect surfaced while testing the Eulerian SUPG solver that grew out of @NengLu's contribution in issue #657; the P0 cell-size field that solver reads is what made the mis-packing visible.

## The defect

`mesh.vars` holds variables weakly. The only strong reference to a MeshVariable is the default Model's registry, and the mesh outlives the model it was created under, so `uw.reset_default_model()` (the test suite does it between tests) releases every variable a script no longer names; the variable-statistics helpers also delete temporaries from the registry on purpose. A released variable is garbage-collected, but a DMPlex cannot shed a field, so its PETSc field stays in the DM.

Two places assumed the registry and the DM's field list line up by position:

- `Mesh.update_lvec` zipped `mesh.vars.values()` against `dm.createFieldDecomposition()`, so every variable after the orphan was packed into the wrong field and its own slot kept whatever it held;
- the JIT's `petsc_a[]` / `petsc_a_x[]` offsets were a running count over the live variables, skipping the orphan's components.

Measured: a DM with seven fields and a registry with three; a P0 cell-size field landing in a P2 slot as garbage (values of order 1.7); a solver giving `DIVERGED_FUNCTION_NANORINF` at iteration zero in one run and a subtly wrong answer (1e-4) in the next, depending on when the collector ran. It surfaced as a flaky test that only failed after a test which dropped variables.

## The fix

`update_lvec` packs by field name and zeroes an orphaned field so nothing stale reaches a kernel. `_jitextension._aux_component_offsets(mesh)` reads component offsets from the DM's own field list and `ccode_patch_fns` patches each variable from its `field_id`.

## Test

`tests/test_1058_dropped_meshvariable_aux_layout.py` (level_1, tier_a, in the `test_105*` CI batch): the premise (a dropped variable leaves an orphaned field), a Poisson solve with a field coefficient after two dropped variables matching a clean mesh to 1e-10, and the packed aux vector landing in the named fields. Two of the three checks fail with the fix reverted. Core batches `test_00xx` to `test_02xx` pass with the fix (398 tests).

Not changed: the DM still grows by one field per dropped variable for the life of the mesh. That is the pre-existing memory behaviour, now merely correct.

### DIFF
--- a/src/underworld3/discretisation/discretisation_mesh.py
+++ b/src/underworld3/discretisation/discretisation_mesh.py
@@ -3896,13 +3896,21 @@ class Mesh(Stateful, uw_object):
             # The field decomposition seems to fail if coarse DMs are present
             names, isets, dms = self.dm.createFieldDecomposition()
 
-            # traverse subdms, taking user generated data in the subdm
-            # local vec, pushing it into a global sub vec
-            for var, subiset, subdm in zip(self.vars.values(), isets, dms):
-                # var.vec lazily creates the PETSc local vector on first access
-                lvec = var.vec
+            # Traverse the DM's fields BY NAME. `self.vars` holds its
+            # variables weakly, so a dropped-and-collected variable leaves
+            # a field behind in the DM; a positional zip would then pack
+            # every later variable into the wrong field (measured: the
+            # cell-size field landing in a P2 slot as garbage, NaN
+            # residuals in a solver that reads it). An orphaned field is
+            # zeroed so nothing stale can reach a kernel.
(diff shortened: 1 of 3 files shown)


lmoresi:
Contained in #688, merged to development as 1d5e8c15: the by-name packing of auxiliary fields and the offsets from the DM field list, with `tests/test_1058_dropped_meshvariable_aux_layout.py`.

Issue #675, opened 2026-09-03
test_0855 and test_0873 fail at np=2 on Linux CI, and 0873 hangs the MPI job (first CI run they have ever had)


lmoresi:
`tests/parallel/test_0855_mesh_smoothing_parallel.py` and
`tests/parallel/test_0873_adapt_collective_stop_mpi.py` both fail at **np=2** on
the Linux CI runner, and the second one takes the whole MPI job down with it —
76 minutes of silence until GitHub cancelled the job, with `mpirun` and
`hydra_pmi_proxy` still alive at cleanup.

Neither has ever run in CI. They are two of the three files #615 identified as
matching **no glob in either test script** (`scripts/test.sh` runs `test_075*` +
`test_10*`; `scripts/test_levels.sh` runs `test_07*`). This is the first time
they have been executed by CI, and they do not pass.

The sharp end of it: `test_0873_adapt_collective_stop_mpi` was added in #596
specifically to guard against a rank-divergent collective stop. It has been dead
weight since the day it landed, and now that it runs, it fails and then hangs —
which is the exact failure mode it exists to catch.

## Evidence

From the `test` job of #615 (run 33690271100, np=2, batch of six files):

```
01:11:04.49  tests/parallel/test_0855_mesh_smoothing_parallel.py ......F
01:11:04.82  tests/parallel/test_0873_adapt_collective_stop_mpi.py F
02:27:58     ##[error]The operation was canceled.
             Terminate orphan process: pid (45533) (mpirun)
             Terminate orphan process: pid (45534) (hydra_pmi_proxy)
```

So: one of seven failures in `0855`, then `0873` fails, then nothing for 76
minutes. A live `mpirun` with no output is a rank blocked in a collective while
another rank has already left it. No traceback was captured — the job was
cancelled before pytest could summarise, so **the assertion messages are still
unknown**.

## What does NOT reproduce

Everything below is macOS (arm64, Open MPI 5.0.10, PETSc via the shared build),
`amr-dev`, at np=2:

| what was run | result |
| --- | --- |
| `test_0855` alone | 4 passed per rank |
| `test_0873` alone | 4 passed per rank |
| the failing batch of six, alone | 14 passed, 1 skipped, 6.6 s |
| batches 1-4 in sequence, same working directory | 42 / 19 / 23 / 14 passed |

The last row rules out contamination from the batches CI runs first — that was
the obvious suspect and it is not the cause. So this is Linux-specific, or
[...]

Issue #661, opened 2026-08-28
test_0053_hang_watchdog crashes its xdist worker intermittently (1 in ~15 CI runs)


lmoresi:
`tests/test_0053_hang_watchdog.py::test_watching_restores_the_previous_watchdog` failed in CI with a crashed worker:

```
__________________________ test_0053_hang_watchdog.py __________________________
worker 'gw0' crashed while running 'test_0053_hang_watchdog.py::test_watching_restores_the_previous_watchdog'
FAILED tests/test_0053_hang_watchdog.py::test_watching_restores_the_previous_watchdog
```

Run [33142970081](https://github.com/underworldcode/underworld3/actions/runs/33142970081), on `feature/pin-y20-relaxation-rate` at `6c46c783`.

## It is flaky, not deterministic

Re-running the **same commit** with no changes passed. The commit that failed touched only `tests/test_1072_free_surface_spherical.py`, which `scripts/test.sh` does not run at all (`test_106*py`/`test_107*py` are excluded pending the #504 triage), so the change could not have influenced it. Every other `test_uw3` run in the preceding ~15, across seven branches, was green.

## Why it is worth a ticket rather than a re-run

The failure mode is a **crashed worker**, not an assertion — the test process dies rather than reporting. That is the same class as #638's root cause (faulthandler repeat-dump against a hot interpreter loops or SEGVs) and #639's orphan leak, both in this watchdog machinery. An intermittent worker crash under xdist is the symptom you would expect if a dump can still land on an interpreter that is not in a safe state.

Two consequences for anyone reading CI:

- A red run here says nothing about the PR under test, so it invites re-run-until-green, which is how a real regression gets waved through.
- Because the crash takes the worker rather than the test, the tests xdist had scheduled behind it on `gw0` do not report either.

## Suggested next step

Not a re-run policy — a native stack. Per #638, theorising about this class without one has burned several wrong hypotheses. The reproduction there needed a cold `.pyc` cache and piped children; worth checking whether the CI runner meets those conditions more often than a warm local run does.

Underworld development team with AI support from [Claude Code](https://claude.com/claude-code)

lmoresi:
Second occurrence, next run, **different test in the same file**:

```
replacing crashed worker gw0
worker 'gw0' crashed while running 'test_0053_hang_watchdog.py::test_report_carries_the_main_thread_stack'
FAILED tests/test_0053_hang_watchdog.py::test_report_carries_the_main_thread_stack
```

[...]

PR #678, opened 2026-09-03
Supervise every parallel batch: bound a collective hang and name the rank that caused it

lmoresi:
## Summary

A rank blocked inside a collective produces no output and never returns. An
unsupervised batch therefore spends the whole job budget in silence and is
cancelled from outside with no diagnosis — measured at **76 minutes** on one run
(#675), which is also why `test_0855` and `test_0873` managed to fail there
without leaving a traceback behind.

`scripts/mpi_supervisor.py` bounds that and says what went wrong.

## Why not the mechanisms we already had

Every one of them asks the stuck process to cooperate, which is the one thing it
has stopped being able to do.

| mechanism | why it fails |
|---|---|
| `threading.Timer` + report | needs the GIL. Measured at np=4 against a 4 s block in `comm.allreduce`: fired **zero** times on the blocked ranks. |
| `faulthandler.dump_traceback_later` | a C thread walking other threads' live frames; can wedge or crash the process it is diagnosing (#661). |
| `MPI_Abort` from a watchdog thread | needs `THREAD_MULTIPLE`, not safe from inside a collective. |
| `mpirun --timeout` | Open MPI `--timeout` vs MPICH `-timeout`, different semantics, and it tells you nothing when it fires. |

The supervisor asks nothing of the job: a parent process holding a clock and a
signal. Identical behaviour under either MPI and on either platform, because it
is POSIX process management and knows nothing about MPI.

## It triggers on silence, not elapsed time

A batch that is still printing is alive. Judging on total runtime means
inventing a wall-clock budget for a matrix nobody has measured, and that number
is wrong the first time someone adds a slow test. The #675 failure was 76
minutes of *nothing*, not of slow progress.

## What it reports

```
=== MPI SUPERVISOR: no output for 300 s ===
    no rank moved in 20 s: this is stuck, not slow
    rank(s) 0 are somewhere the other 3 are not — that is where to look first
      rank 0 is in _rank_zero_misses_the_collective()
      the other 3 are in main()
    rank 0: 0% CPU, idle — blocked rather than polling
    rank 1: 100% CPU, spinning — a busy-wait, which is what an MPI progress engine does
```

Three independent pieces of evidence: whether anything moved between two samples
(stuck vs slow), which rank is somewhere the others are not, and CPU per rank.

Ranks are armed through a `usercustomize` module that registers a `SIGUSR1`
handler. That is a *signal handler*, not a concurrent walker thread — it runs on
the blocked thread itself, only when asked, so arming it costs a healthy run
nothing and it cannot reproduce #661.

## The control

`tests/test_0063_mpi_hang_supervisor.py` plants a hang whose answer is known:
rank 0 sits in a function no other rank can be inside while the others wait in a
barrier. The supervisor must end the job, call it stuck, and name rank 0.

It also asserts the **other direction** — a healthy job passes through untouched
with its own exit status. Without that, every other assertion could be satisfied
by a supervisor that kills everything it is given.

`3 passed in 109.62s`. A harness that catches hangs is worth nothing until it
has caught one.

## Two things worth flagging

**The spinning ranks are the innocent ones.** An MPI barrier busy-waits, so the
ranks burning 100% CPU are the ones *waiting* and the guilty rank is the one
asleep. An earlier version of this guessed the rank-to-CPU mapping by sorting
and reported it exactly backwards; the handler now records its own pid so the
attribution is read rather than inferred. A confident wrong label is worse than
no label.

**At np=2 there is no majority.** The report says so and declines to nominate a
rank on a one-all split, rather than manufacturing a culprit.

## Scope

`scripts/test.sh` routes its two existing parallel batches through it
(`PARALLEL_SILENCE` overrides the budget). This does not touch #615, which
rewrites those loops — that PR can rebase onto this and inherit the protection.

Does not fix #675 or #661; it makes #675 diagnosable and it stops CI arming the
mechanism behind #661.

### DIFF
--- a/docs/developer/guides/mpi-hang-supervision.md
+++ b/docs/developer/guides/mpi-hang-supervision.md
@@ -1,0 +1,117 @@
+# Bounding and diagnosing an MPI collective hang
+
+A rank blocked inside a collective is the worst kind of test failure: it produces
+no output, returns no status, and takes the whole job with it. In CI it costs the
+entire job budget and then gets cancelled from outside, so you learn nothing at
+all. That happened for 76 minutes on a single run ([#675]).
+
+`scripts/mpi_supervisor.py` bounds that, and says what went wrong.
+
+## Why not the mechanisms we already had
+
+Everything we tried first asks the stuck process to cooperate, which is precisely
+what it has stopped being able to do.
+
+| mechanism | why it fails |
+|---|---|
+| `threading.Timer` and a report | needs the GIL. Measured at np=4 against a 4 s block in `comm.allreduce`: a re-arming timer at 0.5 s fired **zero** times on the blocked ranks. |
+| `faulthandler.dump_traceback_later` | a C thread walking other threads' live frames. It can wedge or crash the process it is diagnosing ([#661]). |
+| `MPI_Abort` from a watchdog thread | needs `THREAD_MULTIPLE` and a schedulable thread; not safe from inside a collective. |
+| `mpirun --timeout` | Open MPI spells it `--timeout`, MPICH spells it `-timeout`, and the semantics differ. Kills the job without telling you anything. |
+
+The supervisor asks nothing of the job. It is a parent process holding a clock
+and a signal, so it behaves identically under Open MPI and MPICH, on macOS and
+Linux, because it is POSIX process management and knows nothing about MPI.
+
+## Usage
+
+```bash
+scripts/mpi_supervisor.py --silence 300 -- mpirun -n 4 python -m pytest --with-mpi tests/parallel/
+```
+
+The exit status is the job's own, or `124` if the supervisor had to kill it —
+the same number `timeout(1)` uses, so it reads the way people expect.
+
+`scripts/test.sh` already routes its parallel batches through it. Override the
+budget with `PARALLEL_SILENCE` if a batch legitimately goes quiet for longer.
+
+## It watches for silence, not for elapsed time
+
+A batch that is still printing is alive. Judging on total runtime would mean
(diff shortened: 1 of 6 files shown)


lmoresi:
First real use, and it found the bug it was built for: #675 is root-caused in
#681. Seventy-six minutes of silence became a named diagnosis in six.

```
tests/parallel/test_0855…::test_parallel_matches_serial_bit_identical FAILED
tests/parallel/test_0873…::test_premise_the_metric_splits_the_ranks   FAILED

=== MPI SUPERVISOR: no output for 121 s ===
    no rank moved in 20 s: this is stuck, not slow
    the ranks are in different places, and with no majority none can be
    called the odd one out:
      rank 0 is in _from_plexh5()
      rank 1 is in barrier()
```

Rank 0 inside a collective HDF5 read, rank 1 waiting at a barrier it never
reached — a rank-divergent collective, arriving immediately after a failure, and
correctly reported as a tie rather than a nominated culprit.

**One real limitation, worth a follow-up rather than a change now.** Killing on
silence costs pytest's end-of-run failure report: the supervisor says where the
ranks are stuck but not what failed just before, because pytest prints its
summary at the end of a run and there is no end. I got #675's assertion by
isolating each suspect test in its own `mpirun`, which worked, but that was a
manual step.

Two ways to close it, neither in this PR:

- ask pytest to report failures as they happen (`-x`, or an `--instafail`-style
  reporter) in the supervised batches, so the traceback precedes any divergence;
- have the supervisor send `SIGINT` before `SIGTERM`, which pytest handles by
  printing its summary — worth measuring, since a rank inside a collective may
  not act on it either.

Either belongs with #615, which rewrites these loops and decides how the batches
are split.

PR #689, opened 2026-09-05
Add CitcomS predictor-corrector DDt manager for composed transport

gthyagi:
## Summary

Add the Zhong/CitcomS predictor-corrector as a transport manager for the
composing `uw.systems.AdvDiffusion` solver introduced in #688.

```python
transport = uw.systems.ddt.EulerianSUPGPC(
    mesh, T, U.sym, method="citcoms", temperature_rate_field=Tdot,
)
thermal = uw.systems.AdvDiffusion(mesh, T, U.sym, DuDt=transport)
thermal.constitutive_model.Parameters.diffusivity = 1.0
thermal.solve(timestep=thermal.estimate_dt())
```

The solver assembles diffusion, source, and the manager's transport
contributions. The manager owns the persistent rate, steady directional
SUPG parameter, correction loop, stability estimate, and restart metadata.
The scalar solver supplies residual assembly through a small execution hook.
Implicit CN/BDF continue to use the existing `EulerianSUPG` manager.

## Numerical Behavior

| Manager selection | Purpose | Accuracy contract |
| --- | --- | --- |
| `method="citcoms"` | Fixed two-correction Zhong/CitcomS update | Matches the independently assembled recurrence; approximately first order for nonuniform diffusion |
| `method="pc_converged"` | Consistent-residual correction to tolerance | Second-order reference with `adv_gamma=0.5`; raises on non-convergence |

Both PC modes require continuous scalar P1 fields on triangles or tetrahedra.
Their timestep is `0.9 * min(dt_adv, dt_diff)`. They reuse PETSc vectors and
geometry arrays and allocate no unused implicit history.

## Restart and Integration

- Manager snapshots preserve rate initialization, numerical configuration,
  correction diagnostics, and timestep; mesh snapshots carry the rate field.
- Solver snapshots preserve the timestep and field-change estimator.
- Implicit cell-size fields are registered before fresh-process restoration.
- A PC manager must bind to its exact temperature field and mesh.
- Checkpoints must be restored into a matching object graph. Snapshots from
  the old standalone SUPG solver are not a migration format.

## Validation

Validated locally on 8 September 2026 against the merged development
architecture. Times are elapsed suite times, not CPU-hours. Independent
suites sometimes ran concurrently; rows are not an additive wall-time total.

| Check | Result | Time |
| --- | --- | --- |
| Focused serial API, manager mutations and snapshots | 33 passed | 32.58 s |
[...]

PR #691, opened 2026-09-05
Fix cell-family detection on empty MPI ranks

gthyagi:
## Summary

Determine a distributed mesh's cell family collectively when some MPI ranks own no cells.

`DMPlexIsSimplex()` is rank-local. On an empty rank it reports `False`, while populated ranks correctly report the actual family. UW3 then constructed different coordinate finite elements on different ranks. That desynchronised PETSc communication and caused a later labeled HDF5 mesh load to hang in `PetscSFSetUp_Basic`.

## Change

- Read the local simplex result on each rank.
- Contribute `None` from ranks with no cells.
- Collect the populated-rank results and use one cell family on every rank.
- Use that collective result when selecting triangle, tetrahedron, quadrilateral, or hexahedron element metadata.
- Document the distributed construction invariant.

The serial/populated-rank behavior is unchanged.

## Regression

`tests/parallel/test_0781_empty_rank_mesh_sequence.py` constructs one simplex or tensor cell in 2-D and 3-D on eight ranks, forcing seven empty partitions. It verifies:

- identical `mesh.isSimplex` values on all ranks;
- identical element-family metadata on all ranks;
- successful construction and integration on a second labeled HDF5-backed mesh, reproducing the sequence that previously hung.

## Validation

```text
mpirun -np 8 python -m pytest --with-mpi \
  tests/parallel/test_0781_empty_rank_mesh_sequence.py -v

4 passed per rank in about 23 s
```

This is split from #689 so the generic distributed-mesh correction can be reviewed and merged independently of Eulerian SUPG transport.

### DIFF
--- a/docs/developer/subsystems/meshing.md
+++ b/docs/developer/subsystems/meshing.md
@@ -33,6 +33,24 @@ The meshing subsystem handles computational mesh generation and manipulation for
 - QuadBox / HexBox            # Structured meshes
 ```
 
+## Empty MPI Partitions
+
+Use `mesh.isSimplex` for the mesh-wide cell family. PETSc's
+`mesh.dm.isSimplex()` is a rank-local query and returns `False` on a rank
+with no cells, even when the distributed mesh consists of triangles or
+tetrahedra. UW3 infers the family collectively from populated ranks before
+constructing coordinate finite elements or element metadata.
+
+This is also an MPI correctness requirement: constructing simplex and
+tensor-product coordinate elements on the same communicator consumes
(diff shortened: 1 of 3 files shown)


gthyagi:
This independent MPI mesh fix is now a prerequisite for completing the Eulerian SUPG integration stack:

1. merge #691 and #692 into `development`;
2. sync `feature/eulerian-supg-transport` with updated `development`;
3. refresh and validate #689 against that feature branch.

#691 itself remains one focused commit, has no outstanding review comments, and both current CI checks pass. Review is the remaining merge gate.

Issue #694, opened 2026-09-05
Cell size: one definition, one name — three are in play, the docstrings describe a fourth, and get_*_radius() is partition-dependent


lmoresi:
## The defect

`get_min_radius()`, `get_max_radius()` and `get_mean_radius()` present as the
canonical global "mesh length" API. Their docstrings say so explicitly:

- `get_min_radius`: "the smallest cell **anywhere in the mesh, not just on this
  rank**. Parallel-safe via MPI allreduce."
- `get_max_radius`: "the largest cell **anywhere in the mesh**."
- `get_mean_radius`: "**Use this anywhere you need a representative h0** ...
  rather than reaching for the rank-local `self._radii` array, **which gives
  different answers on different MPI ranks and leaks downstream**."

All three reduce over `self._radii` — the array that last docstring warns
against. The allreduce makes the answer identical **on every rank**. It does not
make it identical **for every rank count**, because the values being reduced are
themselves partition-dependent: `_radii` comes from a kd-tree over *this rank's*
centroids, so a cell near a partition boundary gets a different size depending
on how the mesh was split.

"Same on every rank" and "same for every rank count" are different properties.
These functions deliver the first and their names and docstrings promise the
second.

## Measured

`UnstructuredSimplexBox`, `cellSize=0.12`, 198 cells, on `development`:

| accessor | np=1 | np=2 | np=4 | np=8 |
| --- | --- | --- | --- | --- |
| `get_min_radius()` | 0.0502908499358 | same | same | same |
| `get_max_radius()` | 0.0670934714626 | same | same | **0.070373950134** |
| `get_mean_radius()` | 0.0592989430362 | 0.0594222426856 | 0.0594775301273 | 0.0597593860929 |

`get_max_radius()` moves by **4.9%** at np=8. `get_mean_radius()` moves at every
rank count. `get_min_radius()` is stable here — but incidentally, because the
minimising cell does not happen to sit where the kd-tree lookup differs, not
because anything guarantees it.

## Ruling (maintainer, 2026-09-05)

These ought to be rank-independent — that is part of the parallel contract they
advertise. If a quantity is local-only, its name has to say so.

## The fix is already half-built

#692 adds `_radii_own`: each cell's characteristic length from **its own**
vertices and centroid, which cannot depend on the partition. Measured on that
branch, per-cell values are bit-identical at np=1/2/4 (max abs diff 0.000e+00)
where the legacy field differs by 3.3e-03 and 4.1e-03.

[...]

PR #697, opened 2026-09-06
Take the cell radius from PETSc, and fix the radius accessors with it (#694)

lmoresi:
## Summary

Closes #694, #569 and #687.

Takes the per-cell characteristic length from PETSc's
`DMPlexComputeGeometryFVM` — `volume**(1/dim)` — instead of a kd-tree over this
rank's centroids. A cell's volume is a property of that cell, so this cannot
depend on how the mesh was split.

Builds on #692, which fixed `cell_size()` with a hand-rolled own-centroid RMS
and established the `_cell_radii` name.

## Why PETSc rather than the own-vertex RMS

| | #692 | this |
| --- | --- | --- |
| `cell_size()` partition independent | yes | yes |
| `get_min/max/mean_radius()` | left on the kd-tree field | fixed |
| docstrings | still describe a third definition | become true |
| geometry to maintain for tri/tet/quad/hex | hand-rolled | PETSc's |
| kd-tree loop, `_min_size`, `_search_lengths` | still present | deleted |

`get_max_radius()` moved **4.9% at np=8** and `get_mean_radius()` at every rank
count, while their docstrings advertise "the smallest cell anywhere in the mesh"
and "use this rather than the rank-local array". An allreduce makes an answer
agree on every rank; it does not make it agree at every rank *count* when the
values being reduced are themselves partition-dependent.

And both `cell_size()` and `get_mean_radius()` have always documented "the
`volume**(1/dim)` equivalent radius" — which is what this restores.

## The abandoned routine works

It carried a note that it "does not compute all cells". That does not reproduce:

| case | cells | non-finite/zero | max abs diff np1 vs np2 |
| --- | ---: | ---: | --- |
| 2-D simplex | 66 | 0 | **0.000e+00** |
| 2-D quad | 36 | 0 | **0.000e+00** |
| 3-D tetrahedra | 184 | 0 | **0.000e+00** |
| 3-D hexahedra | 64 | 0 | **0.000e+00** |
| 2-D simplex, deformed | 66 | 0 | **0.000e+00** |

The note also named `DMPlexGetMinRadius`, a different call that is not used here.

Accessors after the change, identical to 12 digits at np=1/2/4/8:

```
min=0.0575495804241  max=0.0868948034925  mean=0.0708408414261
```

## Two constants move with the length

The definition of `h` changed, so constants calibrated against it change too, or
the methods are silently different. Both were found by tests failing, not by
inspection.

**Nitsche `gamma`: 10.0 → 12.5** (maintainer's call). The penalty is
`gamma*mu/h` and `h` is ~19% larger in the mean:

```
development, gamma=10,   old h    8.954e-05
this branch, gamma=10,   new h    1.312e-04     <- past test_1060's 1e-4
this branch, gamma=12.5, new h    7.851e-05
```

`test_1060` passed `gamma=10.0` explicitly, so the default alone did not reach
it; it now omits gamma and follows the default, because what it asserts is that
Nitsche enforces `v.n = 0` at the *recommended* penalty.

**`max_surface_cfl`: 0.5 → 0.437.** The free-surface step is
`max_surface_cfl * get_min_radius() / speed`, and `get_min_radius()` grew by
1.1443, so the surface took ~14% larger steps and `test_1070`'s strong-constraint
volume leak reached 1.063e-03 against a 1e-03 threshold. `0.5 / 1.1443 = 0.437`
preserves the step.

## Tests

`tests/parallel/test_1078` guards the accessors — the half #692 does not reach —
by comparing against this module's own np=1 answer through `serial_reference`,
with the mesh fingerprint asserted so a host that triangulates differently is
reported as that rather than as partition dependence. #692's `test_1077` keeps
covering the field cell by cell; the two are complementary and both stay.

**Negative control**: restoring the kd-tree field makes it fail with
`max radius is partition dependent — np=1 0.0670934714626447 vs np=8 0.07037395013400187`.

`test_0010`'s oracle is rewritten, since it asserted the own-vertex RMS this
supersedes. It now checks `volume**(1/dim)` against a determinant volume from
the cell's vertices (simplices) and against the analytic cell volume scaled by
the affine map's determinant (structured boxes) — computed from coordinates, not
from PETSc, so it still checks the implementation rather than restating it.

`test_1069`'s `local_h=False` workaround is removed. That test measures partition
independence and had to disable the **default** code path to avoid measuring this
defect at the same time; it now runs the default.

| | |
| --- | --- |
| test_1060, 1065, 0010, 0008, 1070 | 26 passed, 1 skipped |
| parallel test_1077 + test_1078 | 3 passed at np=2 and np=4 |
| test_10*, test_0008*, test_005* (before the CFL fix) | 782 passed, 1 failed — that failure is test_1070, fixed here |

## Also

- `follow_metric`'s docstring recommended `2.0 * mesh._radii.mean()`, a
  **rank-local** mean, in the same breath as `get_mean_radius()` warns against
  exactly that. It now recommends `get_mean_radius()`.
- The `TODO(BUG)` on `_assemble_cell_size` describing this defect is removed.


lmoresi:
## Adversarial review of this PR

Copilot cannot be requested here (not a repo collaborator), so this is ours,
by execution.

### What was attacked

The change reads radii from PETSc's `cellgeom` vector while centroids still
come from `_get_coords_for_basis(0, False)`. Everything downstream indexes both
by cell, so the risks are ordering, starved ranks, and deformation.

| check | result |
| --- | --- |
| radii aligned with the cell ordering everything else uses | **verified** — `test_0010` compares each cell's value against a determinant volume from *that cell's* vertices, max error < 1e-12 |
| simplex `deform()` refreshes the radii | **verified** — doubling x scales lengths by sqrt(2), max error 5.6e-17 |
| starved rank, build + accessors | **verified** at np=8 with cells per rank `[2,2,2,2,0,2,2,2]` |
| accessors agree across ranks and rank counts | **verified** np=1/2/4/8 |
| both parallel guards | 3 passed at np=2, np=4 **and np=8** |
| `test_1069` on the restored default `local_h=True` | 7 passed at np=2 and np=8 |

### One thing found, and it is NOT this PR's

At np=8 on a deliberately tiny mesh (14 cells, so one rank owns none),
`mesh.cell_size()` diverges: ranks 0-3 and 5-7 sit inside `_assemble_cell_size`
while rank 4 has already moved on to the next statement.

It is not introduced here:

* `_assemble_cell_size` is documented as deliberately rank-local, and its
  empty-partition early return dates from #275 (`a7f0b11f`), well before this
  work or #692.
* The same 14-cell mesh at np=8 hangs on `development` **earlier still** —
  during mesh construction, before `cell_size()` is reached — so that
  configuration is pathological independently of anything here. That also means
  a clean before/after attribution is not available at that size.

At realistic sizes np=8 is green throughout, as the table above shows. Worth its
own issue in the empty-rank family alongside #691, rather than holding this.

### Corrections to my own earlier claims in this thread

* I proposed the `mean` needed an order-independent reduction. It does not —
  `get_mean_radius()` allreduces both the sum and the count, which is a true
  global mean, and partition-order last-bit differences are out of scope
  (maintainer, 2026-09-05).
* I said this "needs the Nitsche and SUPG benchmarks re-run". The guards are
  oracles, not baselines: `test_1065` and `test_1060` compare against an exact
  essential free-slip solve computed in-test, `test_1064` against published
  Zhong values. Nothing stored a cell-size-derived number.
* I first reported `estimate_dt()` as partition-dependent at ~1e-5. It is not —
  re-running with the velocity written in analytically rather than solved gives
  bit-identical answers at np=1/2/4/8. That spread was the iterative solve.

### Residual risk, stated

The two constants that moved — `gamma` 10.0 -> 12.5 and `max_surface_cfl`
0.5 -> 0.437 — restore the *measured* behaviour of the tests that caught them.
They are not derived from first principles, and no benchmark outside the test
suite has been run. SUPG's guards and the adaptivity/relaxation consumers of
`get_max_radius()`/`get_mean_radius()` have not been audited.

lmoresi:
Heads-up: #688 merged to development (1d5e8c15) and this branch now has one conflict against it, in `discretisation_mesh.py` (development carries #692's `cell_size` with the branch's by-name `update_lvec`). A rebase should be small.

lmoresi:
CI failure was a stale oracle in the test, not the code.

`test_0774_empty_rank_reductions_mpi` asserted `get_min_radius` against the centroid-to-corner half-diagonal this branch replaces:

```
SERIAL_RADIUS = math.sqrt(0.5 ** 2 + 0.25 ** 2)   # 0.559017
```

The fixture cells are 1.0 x 0.5, so the PETSc `DMPlexComputeGeometryFVM` definition gives `area**(1/2) = 0.7071068` — exactly the value CI reported. Updated the oracle to `(1.0 * 0.5) ** (1/2)`, which is still analytic rather than recorded.

Green locally at np=2 and np=4 (7 passed), and `test_0010_cell_size_geometry` plus `test_1078` still pass. Pushed as 291c330f.

Note this branch has not been rebased since #703, #705 and #707 merged.

lmoresi:
Merged development (a0083905). One conflict, in `estimate_dt`: development refactored the inline advective/diffusive block into `_advective_diffusive_dt(...)`, which this branch had rewritten in place. Took development's refactor.

The helper reads `mesh._radii`, which is one of the three names #694 is about; repointed it at the unified `mesh._cell_radii` this branch establishes. That was the only edit needed on top of development's version.

`test_0010_cell_size_geometry`, `test_1007_estimate_dt_scalar_contract` and `test_1057_ddt_transport_plugin` pass locally (12 passed), plus `test_0774` at np=2 and np=4.

lmoresi:
Measured this against the Zhong spherical-shell benchmark, which now runs in CI
(it matched no glob until #731, which is how #734 stayed hidden). **The gamma
change is right and it fixes #734.** Two things in the justification are not,
and one new test will fail until this PR updates it.

## The measurements

`test_1064` response case, `SphericalShell` r 0.55→1.0, cellSize 1/8, on
`development` at 347d7e60:

| gamma | `h` | surface err | CMB err |
|---:|---|---:|---:|
| **10** (current default) | vertex-RMS | **2.36%** | 0.94% |
| **12.5** (this PR) | vertex-RMS | **0.02%** | 0.47% |
| 14 | vertex-RMS | 0.10% | 0.34% |
| 17.3 | vertex-RMS | 0.13% | 0.19% |
| 18.4 | vertex-RMS | 0.13% | 0.16% |
| 10 | global `get_min_radius` | 0.20% | 0.22% |

CI measures 5.67% at gamma=10 where this machine measures 2.36% — gmsh
triangulates the shell differently across platforms — so the 5% tolerance is
currently platform-decided. At 12.5 that stops being true.

## gamma=10 is a cliff, not a mild miss

The surface error falls **100x** between gamma 10 and 12.5, then flattens. The
default was sitting on a steep part of the curve, which is why a modest change
in `h` produced a 5.67% benchmark failure rather than a proportionate drift.
Worth knowing independently of this PR: anything at or above ~12.5 is in the
flat region.

## The stated reason is the wrong way round in 3-D

The docstring says `mesh.cell_size()` is *"PETSc's `volume**(1/dim)`, about 19%
larger in the mean"*. Two corrections:

1. **That 19% is measured against the kd-tree `h` that #692 already replaced.**
   This PR predates #692 landing. On `development` today `cell_size()` is the
   vertex RMS about the own centroid, not the kd-tree statistic.
2. **On the 3-D shell the new `h` is smaller, not larger.** Measured on that
   exact mesh: vertex-RMS mean **0.097874**, `volume**(1/dim)` mean
   **0.070719** — 28% *smaller*, so the penalty gets **1.38x stronger** from the
   `h` change alone, before gamma moves at all. It does go the other way in 2-D
   on flat cells (`h/sqrt(2)` vs `2h/3`, 6% larger), so the direction is
   geometry- and dimension-dependent and the single "19%" figure cannot stand
   for both.

Combined, this PR lands at `1.25 x 1.384 = 1.73x` the current stiffness on this
shell — equivalent to gamma≈17.3 on today's `h`, which measures 0.13% / 0.19%.
Comfortably inside tolerance either way, so the conclusion holds; the reasoning
in the docstring just needs correcting so the next person recalibrating does not
start from a number that is wrong in sign.

## This will fail two tests that landed in #743

#743 pinned `cell_size` to its closed form precisely so a redefinition fails at
the quantity that changed instead of drifting into a benchmark. This PR is the
next redefinition, so it should fail — and should update the baselines as part
of declaring the change. On congruent right-isosceles cells of legs `h`:

| | current (vertex RMS) | this PR (`volume**(1/dim)`) |
|---|---|---|
| `cell_size` | `2h/3` | **`h/sqrt(2)`** |
| `cell_size / get_min_radius`, uniform simplices | `sqrt(2)` | **`3/2`** |
| `cell_size / get_min_radius`, tensor cells | 1 | 1 (unchanged) |

Both are in `tests/test_0010_cell_size_geometry.py`. The docstring correction in
`add_nitsche_bc` about the two measures differing on simplices needs the same
update — the factor becomes 3/2.

Issue #695, opened 2026-09-06
uw.maths.Integral, BdIntegral and CellWiseIntegral read every expression constant as zero: the kernels are compiled with PETSc constants that are never set


lmoresi:
## What happens

Any `uw.function.expression` (UWexpression) in the integrand of `uw.maths.Integral`, `uw.maths.BdIntegral` or `uw.maths.CellWiseIntegral` contributes nothing: the integral of `c * f` is zero for any expression `c`, whatever its value, and a fresh Integral built after the value changed returns the same zero.

```python
c = uw.function.expression(r"c", 2.0, "probe")
uw.maths.Integral(mesh, c * T.sym[0].diff(y)).evaluate()        # 0.0, expected 2.0
uw.maths.Integral(mesh, 2.0 * T.sym[0].diff(y)).evaluate()      # 2.0
uw.maths.BdIntegral(mesh, stokes.constitutive_model.flux[0, 1], "Top").evaluate()   # 0.0: eta is an expression
```

The constitutive viscosity is an expression, so every boundary traction formed from `solver.stress` or `constitutive_model.flux` integrates to its pressure part only. We found it on the DFG cylinder benchmark, where the drag came out 23 to 28% low at two mesh sizes and did not move with the SUPG weights: the whole deficit was the viscous drag, which the traction integral had silently dropped. The reaction (consistent-flux) form of the force, which uses a volume Integral of the same stress, was wrong in the same way.

## Cause

The JIT (`getext`) routes every expression constant to PETSc's constants array (`_extract_constants` / `PetscDSSetConstants`), so that a changed value does not recompile. The solvers call `PetscDSSetConstants` before every solve (`_update_constants`). The three integral classes compile through the same `getext` but never set the constants on the DS they integrate with, so the kernel reads zeros. `BdIntegral` integrates on a sandbox DM with its own DS (`DMCreateDS` in `UW_DMCreateBdIntegralSandbox`), so the values have to go to that DS.

`uw.function.evaluate` is unaffected (it goes through a different path).

## Fix

[...]

PR #700, opened 2026-09-07
Set the PETSc constants on the DS the integrals use, so expression values reach the kernels (#695)

lmoresi:
`Integral`, `CellWiseIntegral` and `BdIntegral` compile their integrand through the JIT with a constants manifest but never called `PetscDSSetConstants` on the DS they evaluate on, so every `uw.function.expression` in an integrand was read as zero by the kernel. A viscous traction integral over a cylinder therefore returned the pressure drag alone (the 'drag deficit' on the Navier-Stokes benchmarks, #695). The fix packs the manifest and sets it on the DS (the sandbox DS for boundary integrals) after the objective is set. `tests/test_0503_integral_expression_constants.py` covers a volume, boundary and cell-wise integral of an expression constant, the value changing after `c.sym = ...`, and a constitutive-model flux on a boundary.

Extracted from #688 as a bug fix in its own right. Closes #695.

### DIFF
--- a/src/underworld3/cython/petsc_maths.pyx
+++ b/src/underworld3/cython/petsc_maths.pyx
@@ -1,5 +1,6 @@
 from typing import Union
 import sympy
+import numpy as np
 
 import underworld3
 import underworld3.timing as timing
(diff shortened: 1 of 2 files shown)


lmoresi:
Contained in #688, merged to development as 1d5e8c15: the same `petsc_maths.pyx` change and `tests/test_0503` (with one more test there). #695 is closed against it.

PR #720, opened 2026-09-10
SLCN at the integration points: vector and tensor histories

lmoresi:
`IntegrationPointSemiLagrangian` refused anything but a scalar:

```python
if vtype != VarType.SCALAR:
    raise NotImplementedError(
        "IntegrationPointSemiLagrangian: scalar histories only for now")
```

so a Navier-Stokes momentum history or a viscoelastic stress history could not use the integration-point route at all. `Lagrangian_Swarm` has carried both since the VE stress history (`test_0070`); this brings the two level. The choice between them is *where the state lives*, not what shape it can take.

## What it took

The trace-back, the characteristic cache and the weighted sums were already shape-agnostic — only the fills were scalar-bound, at four sites. The one real subtlety is that a shaped field stores one dof per **independent** component rather than one per matrix entry:

| `vtype` (2-D) | symbolic shape | stored columns |
|---|---|---|
| `SCALAR` | 1×1 | 1 |
| `VECTOR` | 1×2 | 2 |
| `SYM_TENSOR` | 2×2 | **3** |

```python
DFDt = uw.systems.ddt.IntegrationPointSemiLagrangian(
    mesh, stress, v.sym, vtype=uw.VarType.SYM_TENSOR, degree=2, order=1)

DFDt.psi_star[0].sym          # 2x2 symbolic matrix
DFDt.psi_star[0].data.shape   # (npoints, 3)
DFDt.bdf()                    # 2x2, as psi_fn is
```

The column order was **measured, not assumed**: diagonal first, then the off-diagonals in row-major upper-triangular order — `(0,0), (1,1), (0,1)` in 2-D and `(0,0), (1,1), (2,2), (0,1), (0,2), (1,2)` in 3-D. Getting it wrong transposes a stress in silence, so `_storage_components` is pinned by a test against what the variable's own `.sym` reconstructs, in both dimensions.

Fills are now component-wise (`_write_components`), because a symmetric tensor's symbolic form repeats its off-diagonals and only the independent columns exist in storage.

## Accuracy

The scalar property, per component: with a uniform velocity and a field in the P2 space, every slot holds the snapshot evaluated at the **exact** departure point to `< 1e-12`, for one segment and for two. The scalar path is unchanged, and a test asserts that.

The tensor test uses three *distinct* independent components so a packing error or a transposed off-diagonal cannot pass.

## Docs

The documentation contradicted itself, which is likely why this never reached anyone: `docs/developer/subsystems/integration-point-variables.md` said *"Scalar components only for now; use one variable per component"* while `enhanced_variables.py` said vector and tensor were supported. Both now agree, and there is a new **Vector and tensor histories** section with the table, the storage convention, and a pointer to `Lagrangian_Swarm` for the same history on particles.

## Tests

`tests/test_0068_integration_point_slcn_tensor.py` (10). Full `level_1 and tier_a`: **1197 passed, 3 skipped, 1 xfailed** (development is 1187 + these 10).

https://claude.ai/code/session_01MSGAFeA7qYXgkuw9ud8F2G


lmoresi:
Review from the 2026-09 style/clutter audit. One finding, scoped to lines this PR adds.

**`ddt.py:4526` — flat `.data` where §7 mandates `.array`.**

```python
var.data[:, column] = np.asarray(...)
```

The value written has been through `np.asarray` conversion, so this is not the
one sanctioned flat-`.data` use (a raw variable-to-variable copy inside the
non-dimensionalisation boundary) and the units evaluation it bypasses matters.

For contrast, `ddt.py:4567` in this same PR —

```python
self.psi_star[k].data[...] = self.psi_star[0].data[...]
```

— *is* the sanctioned raw copy and needs no change. Listing both so the line is
visibly drawn rather than guessed at.

Filed as #722 with the instances in the merged PRs, since the pattern is wider
than this branch.

Also note #721: the `test_006*` band this work is tested in matches no glob in
`scripts/test.sh`, so the integration-point suite has not been running in CI.

Issue #721, opened 2026-09-10
scripts/test.sh: eleven test files match no glob, so the integration-point and swarm suites have never run in CI


lmoresi:
## What we found

`scripts/test.sh` selects serial tests with numeric batch globs. One of them,

```
$PYTEST tests/test_005[1-9]*py tests/test_006[0-1]*py || status=1
```

stopped covering its own band as soon as `test_0062` was added. Eleven test
files currently match no glob in the script and have never run in CI:

```
tests/test_0062_rotated_constraint_exclusivity.py
tests/test_0063_mpi_hang_supervisor.py
tests/test_0064_quadrature_point_fe.py          <- #703
tests/test_0065_integration_point_variable.py   <- #703
tests/test_0066_integration_point_slcn.py       <- #703
tests/test_0067_integration_point_proxy.py      <- #707
tests/test_0068_swarm_repopulation.py           <- #713
tests/test_0069_swarm_midtime_velocity.py       <- #713
tests/test_0070_ve_stress_history_on_particles.py <- #713
tests/test_0301_meshvariable_clone.py
tests/test_1030_analytic_zhong2008.py
```

That is the entire integration-point suite delivered by #703 and #707, and the
repopulation / mid-time-velocity / VE-stress-on-particles tests delivered by
#713. Each of those PRs shipped green.

This is separate from `test_106*`/`test_107*`, which are deferred on purpose
and documented in the script against #504.

## Why it is the shape of the problem, not one glob

A numeric range has to be widened by hand every time the band grows, and
nothing says when it wasn't. This is the #570 class. Widening
`test_006[0-1]` to `test_00[6-9]` fixes today's instance and leaves the next
one to chance.

## Fix

Widen the stale globs, and add `scripts/check_test_coverage.py`, which fails
the run if any `tests/test_*.py` matches no glob. Files deferred by decision
go in a `DEFERRED` list carrying the reason and issue, on the same terms as
the allowlist in `check_deprecated_patterns.py`: it may only shrink.

Underworld development team with AI support from Claude Code

PR #723, opened 2026-09-10
A guard for the test globs, and 22 session scripts out of the test tree (#721)

lmoresi:
Rebased onto current `development`. `586224f6` already closed the 006x-007x gap
and took the band **whole** rather than enumerating it — the better fix, and it
makes that range safe from recurrence. My widening of it is dropped. This is
what is left of #721 after that.

## The other ranges are still hand-maintained

`scripts/check_test_coverage.py` fails the run when any `tests/test_*.py`
matches no glob in `scripts/test.sh`. Run against `development` as it stands, it
finds two files the band fix could not reach, because they are nowhere near that
band:

```
tests/test_0301_meshvariable_clone.py     no test_03*  glob existed
tests/test_1030_analytic_zhong2008.py     no test_103* glob existed
```

Both globs are added here.

Files deferred by decision go in a `DEFERRED` list with their reason and issue,
on the same terms as the allowlist in `check_deprecated_patterns.py` — it may
only shrink. The deferred pattern is written `test_06[0-9][0-9]_*` rather than
`test_06*`, because the latter also swallows `test_0062..test_0069`, a different
suite which must run.

## The test tree

22 session scripts — drivers, plotting, benchmarks, profilers — were sitting in
`tests/` among the real test files. They are not tests, nothing imports them,
and they made the directory hard to read. They move to `scripts/sessions/` with
a README.

Three of them wrote their figures to an absolute path inside a worktree that no
longer exists, so they had been writing nowhere since that worktree was removed.
They now write beside themselves, and the three stale figures those runs left in
the repository root are removed.

`_mg_ladder.py` stays in `tests/` — it is a test helper and two tests import it.

## Verified

Guard passes: all 302 test files reachable from `scripts/test.sh`, 25 deferred
by decision.

Closes #721


lmoresi:
Review points addressed in `7198bd26`.

**The deferred pattern claimed a file that runs.** Correct — `test_107*` also
matches `test_1072_free_surface_spherical.py`, which `test.sh` pulls out of that
band and runs by name, so the guard reported an actively covered test as
deferred. Narrowed to the files actually deferred (1070, 1071, 1073).

**No stale-entry check.** Also correct, and the sharper point: `DEFERRED` was
only subtracted from the file list, so an entry that stopped being true sat
there indefinitely while the output still counted it as a deliberate deferral —
a shrink-only list quietly becoming a fiction. The guard now rejects an entry
matching nothing, and one matching only files a glob already runs. Verified
against a probe carrying one of each:

```
DEFERRED has stale entries — the list may only shrink:
    'test_00[0-4]*' matches only files that already run
    'test_9999_*' matches no test file
```

**A deferral with no issue behind it.** The `test_06NN` entry had only "pending
validation" because there is no issue — the band was disabled in `test.sh` as
"potentially problematic" and nothing was filed. Recorded as it stands rather
than dressed up, with a note that #731 re-enables the band and removes the entry
(it does: `DEFERRED` is empty there).

**Stale usage paths.** Eleven moved scripts still told the reader to run them
from `tests/`, which this PR's own move invalidated. Repointed.

The comment about the guard's location was already corrected in the rebase — it
now reads "reachable from a glob below", and the guard runs above them.

PR #728, opened 2026-09-11
Nitsche free-slip tests: a wiring check, a real assert, and the comparison at tier C

lmoresi:
Found by turning on `tests/test_1060_nitsche_freeslip.py`, which matched no glob
in `scripts/test.sh` and had never run in CI (#721). Two defects, both test-side.

## An absolute threshold that tracked the forcing

`test_nitsche_normal_velocity_zero` asserted `max|v_y| < 1e-4` and produced
`1.0081e-04`. The bound was **absolute**, so it scaled with the buoyancy forcing
rather than with anything about Nitsche: |v| peaks at 1.8e-2 in this fixture, so
it was ~5.7e-3 relative and sat a fraction of a percent from failing for reasons
unrelated to the method under test.

Nitsche is a weak constraint and does not drive `v·n` to machine precision —
documented behaviour, and the reason rotated strong free-slip exists
([`subsystems/rotated-freeslip.md`](docs/developer/subsystems/rotated-freeslip.md)).
The solver is right, so the test is what changed. It is now a **wiring check**
with a deliberately loose relative bound and a docstring saying not to tighten
it: a tight number would describe this fixture, not the code.

## A comparison test with no assertion

`test_nitsche_better_than_penalty_constraint` ended on

```python
        # Nitsche at gamma=10 should be comparable or better than penalty at 1e4
```

with no `assert`. The file was truncated mid-write at 150 lines, so it computed
both numbers, printed them, and passed unconditionally.

What replaces it is a **characterisation at tier C**, not a contract. The
ordering it records is what the free-slip rulings rest on: an essential BC holds
`v·n` to machine precision, while Nitsche and penalty leave a finite leak.

That assertion **can fail because the code got better** — a weak path coming out
exact would fail it — and tier C is the right home for exactly that: it demands
an explanation, not a revert. The failure message says so outright:

> a weak constraint reaching machine precision is GOOD NEWS and a change in
> behaviour. Explain it and re-characterise; do not revert to make this pass.

It deliberately does **not** rank the two weak methods against each other. Which
leaks less is problem-dependent and moves with gamma and the penalty
coefficient; asserting it would be cheering for a method rather than testing the
code (maintainer ruling, 2026-09-10). The measured numbers — essential 0.0,
penalty 1.5e-3, Nitsche 5.7e-3 at res=8 — are recorded in the docstring as a
dated characterisation, not as an assertion.

## Scope

No library code changed. Per Charter §8 the tests' own correctness was
established first, and the conclusion is that the tests were wrong. 5 passed
locally; the characterisation is independently selectable with `-m tier_c`.

### DIFF
--- a/tests/test_1060_nitsche_freeslip.py
+++ b/tests/test_1060_nitsche_freeslip.py
@@ -116,35 +116,78 @@ class TestNitscheFreeslip:
         print(f"Penalty vs essential: relative L2 diff = {diff:.4e}")
         assert diff < 0.01, f"Penalty differs from essential by {diff:.4e}"
 
-    def test_nitsche_normal_velocity_zero(self, solutions):
-        """Normal velocity on free-slip boundaries should be near zero."""
+    def test_nitsche_constrains_the_wall_normal_velocity(self, solutions):
+        """The Nitsche term is wired up and acting on the wall-normal component.
+
+        This is a WIRING check, not an accuracy claim. Nitsche is a weak
+        constraint: it does not drive v.n to machine precision and is not
+        supposed to (see docs/developer/subsystems/rotated-freeslip.md, which is
+        why rotated strong free-slip exists). How small the residual leak is
+        depends on gamma, the mesh, the forcing and the viscosity, so any tight
+        number here would be a property of this fixture rather than of the code.
+
+        The bound is therefore deliberately loose — it fails if the constraint
+        is absent or has no effect, and passes for any configuration in which it
+        is doing its job. Do NOT tighten it to track a measured value: the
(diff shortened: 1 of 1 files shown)


lmoresi:
On the review point that this test is not exercised by CI — correct, and it is
the reason the defects survived. `scripts/test.sh` leaves `test_106*` unbatched,
so `test_1060_nitsche_freeslip.py` has never run there.

That is fixed in **#731**, which batches `test_106*`/`test_107*` and the
`test_06NN` band. This PR is deliberately the first of the chain: #731 enabling
`test_106*` while the `<1e-4` bound is still wrong would simply turn CI red, so
the fix has to land first.

Merge order: **#728 → #729 → #723 → #731**.

Issue #730, opened 2026-09-11
CI runtime: shard the test batches across parallel jobs when the single job tightens again


lmoresi:
`build_uw3_and_test.yaml` referenced "the CI-runtime issue" as the home for the
durable fix to wall-clock. That issue was never actually filed. This is it, with
the measurements.

## Where it stands

The 120-minute cap was set on a 2026-08-15 measurement putting the serial phase
alone at ~55 minutes, before xdist distributed the batches. Measured 2026-09-12,
the five most recent successful runs spent this long in "Run tests":

```
34m19s   development
35m23s   development
42m32s   feature/particle-demos
44m20s   docs/tier-c-characterisation-tests
47m34s   docs/timestepping-pattern
```

That headroom is why the previously-deferred bands could simply be switched on
rather than traded against something else.

## The fix, for when it is needed

Split the batches across parallel GitHub jobs so wall-clock stops tracking total
test time. The serial phase already falls into natural groups that map onto a
job matrix without restructuring `scripts/test.sh` much:

- `test_00*`-`test_03*`
- `test_05*`-`test_08*`
- `test_10*`-`test_14*`
- the `--p N` mpirun phase, which runs its own batches

## Not yet, though

Doing it now would be machinery ahead of need. The single job fits inside its
budget with room, and every job added pays the ~2.5 min build again unless the
build artefact is shared between them — which is the part that needs designing,
not the matrix.

Worth revisiting when "Run tests" passes roughly 70 minutes, or when the build
step can be cached across jobs.

Underworld development team with AI support from Claude Code

PR #731, opened 2026-09-11
Stop excluding tests by band, let tier C report instead of gate, and quote $PYTEST

lmoresi:
Item 4 of the 2026-09 audit. Depends on nothing, but reads best after #729,
which defines what tier C means.

## The bands were never a cost decision

`test_06*` was disabled as "potentially problematic" and stayed that way.
`test_106*`/`test_107*` were held back pending "a triage/deselect decision" that
nobody made. Measured 2026-09-12, those three bands hold **243 tests, of which
only 26 are level_3 or tier_c**. 55 are level_1 **and** tier_a — hardened, fast,
and exactly what CI exists to run.

One slow test (`test_1064`) was quarantining its neighbours because they shared
a number. They run now; the only failure among them was a test-side defect,
fixed separately in #728.

## It is affordable today

The comment justifying the 120-minute cap measured the serial phase at ~55 min
on 2026-08-15, *before* xdist. The five successful runs before 2026-09-12 spent
**34, 35, 43, 44 and 48 minutes** in "Run tests". The deferred bands add ~22 min
single-process locally, considerably less under `-n 4`.

Sharding across parallel jobs stays the durable fix if this tightens again. It
was referenced in the workflow as "the CI-runtime issue" but never actually
filed — now #730, with these numbers.

## Tier C reports rather than gates

A tier C failure demands an explanation, not a revert: these are
characterisations that can fail because the code got **better**. They are
excluded from the batches that set `$status` and run in their own pass at the
end, which is read but cannot fail the build.

**One trap worth flagging in review.** The exclusion goes through
`PYTEST_ADDOPTS`, not appended to `$PYTEST`. `$PYTEST` is used *unquoted* at
every call site, so bash splits `-m 'not tier_c'` into the words `'not` and
`tier_c'` — and pytest then **silently collects nothing and exits 0**, reporting
a green run that tested nothing. Verified:

```
$ pytest -m "'not" "level_2'" tests/test_1070_free_surface_plume.py
no tests collected in 0.00s          # exit 0
```

pytest shlex-splits `PYTEST_ADDOPTS`, so the quoting survives there. The
reporting pass clears it with `env -u`, or it would select nothing itself.

## Also drops a line that now runs twice

`test_1072` was pulled forward out of `test_107*` by name *because* that band was
not batched. The band runs as a whole now, so the named line is redundant — it
is still covered, once.

## Verified

Dry run with a stub `pytest`: every gating batch carries the exclusion, all the
previously-deferred bands appear, `test_1072` appears exactly once, and the tier
C pass runs with `PYTEST_ADDOPTS` unset.

### DIFF
--- a/.github/workflows/build_uw3_and_test.yaml
+++ b/.github/workflows/build_uw3_and_test.yaml
@@ -90,3 +90,12 @@ jobs:
 
       - name: Run tests
         run: pixi run -e dev ./scripts/test.sh --p 2
+
+      # The tier C characterisations are a REPORT, not a gate, so they run on a
+      # push to development rather than on every pull request. They include the
+      # slow level_3 cases and add ~42 min, which a PR should not pay to learn
+      # something that cannot block it. Failures here are expected to be read
+      # and answered, not reverted.
+      - name: Tier C characterisations (report only)
+        if: github.event_name == 'push'
+        run: pixi run -e dev ./scripts/test.sh --tier-c-report
--- a/.github/workflows/build_uw3_and_test.yaml
+++ b/.github/workflows/build_uw3_and_test.yaml
@@ -45,13 +45,16 @@ env:
 jobs:
   test:
     runs-on: ubuntu-latest
-    # The suite has outgrown 60 minutes. Measured 2026-08-15: the SERIAL phase
-    # alone is ~55 min (three batches at 14-15 min each), so the parallel phase
-    # - which runs last - was being cut off mid-run and reported as a failure
-    # that looked like a test failure. Any addition at all pushed a run over.
-    # This is a stop-gap that unblocks landing work; the durable fix is to split
-    # the batches across parallel jobs so wall-clock stops tracking total test
-    # time. See the CI-runtime issue for that.
+    # Raised past 60 minutes on a 2026-08-15 measurement (serial phase ~55 min).
+    # xdist has since cut that: the five successful runs before 2026-09-12 spent
+    # 34, 35, 43, 44 and 48 minutes in "Run tests", well inside this budget. The
+    # headroom is why the previously-deferred bands could simply be switched on
+    # (#721 follow-up) rather than traded against something else.
+    #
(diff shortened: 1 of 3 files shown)


lmoresi:
Review points addressed.

**Dependencies, which this PR did not declare — correct, and now stated.** It
cannot land alone:

- **#728** fixes `test_1060_nitsche_freeslip`, whose bound is still `<1e-4`
  against an observed `1.0081e-4`. Enabling `test_106*` without it makes the
  batch fail.
- **#729** puts the `tier_c` mark on the `test_1070` comparison and updates the
  tier definitions in `tests/pytest.ini`, TESTING-RELIABILITY-SYSTEM.md and
  HOW-TO-WRITE-UW3-SCRIPTS.md. Without it this PR enables `test_107*` while that
  comparison still gates, and the new reporting pass contradicts a marker
  contract that still reads "development only, not for automation".

Merge order: **#728 → #729 → #723 → this**. Retargeted onto #723 to make the
last step explicit.

**PYTEST_ADDOPTS** — already fixed in an amended commit the review had not seen.
`$PYTEST` is now a bash array, which also removes the objection behind that
comment: an inherited `PYTEST_ADDOPTS` is no longer discarded, because the
script no longer sets it.

**One thing the rebase found.** Landing on #723 surfaced an integration defect
that #723's own guard caught: `globs_run_by` matched the literal `$PYTEST`, and
this PR rewrites every call site to `"${PYTEST[@]}"`. The guard saw no globs at
all and declared all 302 test files dark. It now matches both spellings — a
useful demonstration that the guard does its job.

`DEFERRED` is now **empty**. Every band that sat there is batched; nothing is
excluded from CI by its number any more. A test that should not gate says so
with `@pytest.mark.tier_c`, which is a property of the test rather than of where
it falls in the numbering.

Verified: the marker arrives as a single `ARG[not tier_c]`, the tier C pass runs
without it, and the guard reports 302 reachable, 0 deferred.

